PR #15, opened 2021-01-25
API change

abhi-agg:
Replace TranslationModelConfiguration with std::string

Tested the marian::Options generated in AbstractTranslationModel.cpp class
- User provided config options override the default config options


src/translator/AbstractTranslationModel.cpp
```diff
@@ -6,16 +6,19 @@
...
 std::shared_ptr<AbstractTranslationModel>
-AbstractTranslationModel::createInstance(const TranslationModelConfiguration& config) {
-	TranslationModelConfigToOptionsAdaptor adaptor;
-	auto options = adaptor.adapt(config);
+AbstractTranslationModel::createInstance(const std::string& config) {
+	auto options = std::make_shared<marian::Options>(YAML::Load(config));
+	marian::ConfigParser configParser(marian::cli::mode::translation);
+	const YAML::Node& defaultConfig = configParser.getConfig();
+	options->merge(defaultConfig);
 	return std::make_shared<TranslationModel>(options);
```
jerinphilip, line 23: Just pass the string down, why do you need `marian::` stuff here? This is supposed to construct other (non-marian) translators which implement AbstractTranslationModel as well, correct? Let the implementation handle how to do configs. Add a translator tag based on which the TranslationModel is built and use the tag to switch. If unknown, raise an unknown translator error.

PR #16, opened 2021-01-28
Formatting tweaks and Wasm target name refactoring

motin:
(And some formatting / cleanup)

Note that this is just a simple/dummy implementation to verify that the basic translation interface is working, like we discussed.


README.md
```diff
@@ -3,50 +3,60 @@
...
+1. Get the latest sdk: `git clone https://github.com/emscripten-core/emsdk.git`
+2. Enter the cloned directory: `cd emsdk`
+3. Install the lastest sdk tools: `./emsdk install latest`
+4. Activate the latest sdk tools: `./emsdk activate latest`
+5. Activate path variables: `source ./emsdk_env.sh`
+
 After the successful installation of Emscripten, perform these steps:
+
+```bash
+git clone https://github.com/browsermt/bergamot-translator -b wasm
```
abhi-agg, line 31: `-b wasm` should be removed from here.

motin: Currently, only the wasm branch contains the dummy api. Remove it when/before this code goes into the main branch again?

abhi-agg: With the current change, the build instructions become wrong. However, I am fine with the proposed suggestion.

abhi-agg:
@motin Lgtm. However, the title of the PR is misleading. Could you please change the title? I will merge after that.

PR #22, opened 2021-02-14
Including a more elaborated test page, a node webserver containing the proper cors headers and wasm mimetype

andrenatal:

PR #23, opened 2021-02-14
Use emrun to host wasm test page

motin:
Use emrun instead of missing node.js server script from #22


andrenatal:
Thanks for the heads-up, but let's use the node server which is more flexible and plays nicer with taskcluster and try. I've just pushed it.

Issue #50, opened 2021-03-09
Model loading from byte buffer


jerinphilip:
@XapaJIaMnu 

For model load I see modifications to [Service](https://github.com/browsermt/bergamot-translator/blob/f17f02a54431403956d1fa333f44d1144e66dd8f/src/translator/service_base.h) and [BatchTranslator](https://github.com/browsermt/bergamot-translator/blob/f17f02a54431403956d1fa333f44d1144e66dd8f/src/translator/batch_translator.h).  Service is currently subclassed as a [NonThreaded](https://github.com/browsermt/bergamot-translator/blob/f17f02a54431403956d1fa333f44d1144e66dd8f/src/translator/service_base.h#L45) implementation and a [multithreaded](https://github.com/browsermt/bergamot-translator/blob/f17f02a54431403956d1fa333f44d1144e66dd8f/src/translator/service.h#L18) one. Calls to initialize BatchTranslator happens in respective constructors.

The model loading happens actually in BatchTranslator.

https://github.com/browsermt/bergamot-translator/blob/f17f02a54431403956d1fa333f44d1144e66dd8f/src/translator/batch_translator.cpp#L26-L39

So I'll need createScorers (L32) from marian with a bytesbuffer which I pass all the way from whichever implementation of Service requires it. The concurrent-queuing implementation is a bit ahead and both messier and cleaner depending on places (imo), but I'm taking responsibility of bringing it to sync once you have integrated your changes in main.

XapaJIaMnu:
So... I have been working off main and apparently that branch is very far from state-of-the art. Sadly.
If you check https://github.com/browsermt/bergamot-translator/tree/load_as_byte_arr this exposes the byte array loading all the way to the top. Which branch do I need to merge this with so that I can write an example code?

jerinphilip:
> I hope we are doing something about the redundant abstraction. AbstractTranslationModel can be merged with TranslationModel. 

@abhi-agg wants multiple translators integration capability. My understanding is this means more than marian (maybe another CPP translator/whatever). Rename TranslationModel -> MarianIntegrationModel : AbstractTranslationModel and the existence could be justified, I guess.

> Which branch do I need to merge this with so that I can write an example code?

**Critical bit**: It is `main`. You're probably missing `-DUSE_WASM_COMPATIBLE_SOURCES=off` without which everything is turned off. bergamot-translator is running WASM default. I settled for the flag and the ability to switch modules to compile against full code in my development work.  Have a look at:
[...]

PR #51, opened 2021-03-10
Updated marian submodule

abhi-agg:
- Includes the change that fixes the binary model loading problem for wasm
 - submodule is still pointing to `wasm` branch of `browsermt/marian-dev`
     - It will be changed to point to `main` branch once the corresponding PR gets merged


jerinphilip:
There's a submodule change also in https://github.com/browsermt/bergamot-translator/issues/50, for some model loading. Is this synchronized?

abhi-agg:
> There's a submodule change also in #50, for some model loading. Is this synchronized?

The change that you mentioned is related to using models as byte array which is independent of this PR (that enables binary model loading from files) 👍

motin (approved):
I have no easy trajectory for testing that this works before approving, so I am approving this assuming that @abhi-agg have tested that this indeed makes binary models load and translate properly :)

abhi-agg:
> I have no easy trajectory for testing that this works before approving, so I am approving this assuming that @abhi-agg have tested that this indeed makes binary models load and translate properly :)

Yes @motin. I tested it. It works 👍

jerinphilip (commented):
Kindly link to the places this PR here helps. I'm assuming this is to bump it somewhere for the extension etc to fix and test model loads. 

I understand the information is asking is probably available scattered in the branches (marian-dev / extension) and etc, just not obvious to find and would want it here for completeness.

abhi-agg:
> Kindly link to the places this PR here helps. I'm assuming this is to bump it somewhere for the extension etc to fix and test model loads.
> 
> I understand the information is asking is probably available scattered in the branches (marian-dev / extension) and etc, just not obvious to find and would want it here for completeness.

We are not using the .bin model files in the extension as of now because it was not working for wasm. This PR is the first step towards achieving that 👍

jerinphilip (approved):
> This change however fixes loading the .bin files for wasm (@XapaJIaMnu provided this fix to us). It is necessary for us to merge it to main to be able to load .bin models which are smaller in size. Therefore, it would be best if we can merge it just now.

In that case, I'm conditionally approving the changes, @abhi-agg is expected to link PR in the extension which uses the changes in this bump (ideally pre-merge, there should be a WIP draft somewhere. I guess post-merge to main is also okay).

abhi-agg:
> Addendum: Ignore Ubuntu build failures, I'm on fixing it.

Thanks. I will merge it to main then 👍

abhi-agg:
This submodule update was pulled in [extension code ](https://github.com/mozilla-extensions/bergamot-browser-extension) (the submodule bergamot-translator contains this change) 👍

PR #55, opened 2021-03-18
Load mode as a byte array

### LINKED ISSUE #50: Model loading from byte buffer

XapaJIaMnu:
Fixes #50 requires further https://github.com/browsermt/marian-dev/pull/27


src/translator/service.h
```diff
@@ -46,6 +50,7 @@ class Service : public ServiceBase {
 
   size_t numWorkers_;      // ORDER DEPENDENCY
   PCQueue<Batch> pcqueue_; // ORDER DEPENDENCY
+  const void * model_memory_;
```
jerinphilip, line 53: Shouldn't this be in `ServiceBase`, as it is common to both `NonThreadedService` and `Service`? I see that this is not pleasant, which is my fault originally. Is there a way to keep multithread and single-thread separately better than the current setting?

XapaJIaMnu: Technically it shouldn't at all be in `ServiceBase` as it doesn't initialise the translator, but as I was writing it, I got confused by the multiple levels of abstraction. Because things get initialised in an extremely confusing way...

My suggestion would have the same class Service/ServiceBase and depending on the number of workers provided to either use the single or the multithreaded queue? Is there a scenario where there are multiple producers but a single consumer? If mozilla's thing can't compile with threads, use compile time guards that check for wasm.

 Also merge the translationModel/AbstractTranslationClass and what not.

Finally move the translator to the serviceBase? That way we can massively reduce the number of classes.

XapaJIaMnu: Anywho, completely removed `model_memory` from the `ServiceBase`. It was just an unused argument.

jerinphilip: > Technically it shouldn't at all be in ServiceBase as it doesn't initialise the translator, but as I was writing it, I got confused by the multiple levels of abstraction. Because things get initialised in an extremely confusing way...

Both NonThreadedService and Service initializes the translator (or translators). So the void* pointing to the memory should be available for both. For DRY, the void* should go to ServiceBase. The variation is in how the translators are initialized (std::thread launch vs no std::thread launch), a translation is queued (synchronous / asynchronous) and how the threads are shutdown which I'm trying to override as methods. 

Should we make the cleaning classes thing it's own issue?

XapaJIaMnu: Ugh. So, imagine you are a programmer. You go to the `ServiceBase` file and you see a member variable `model_bytes_`. You see that this variable is not used ANYWHERE in the class? Why is it there?

At the same time you go to another file and you see it makes use of a variable `model_bytes_`, which does not appear anywhere in the file. Do you see how this gets confusing really fast?

If you want to have serviceBase, make it so that the translator appears there (I still think it is a bad idea to have them that decoupled, but...)

No need to open an issue for the redesign, just halve the number of classes (;

kpu (approved):
I would have liked to see magic bytes but at least marian is checking model types so it should mostly be fine and we can't change the binary format without pain.

jerinphilip (approved):
Since my queries have been addressed, I'm approving these changes.

kpu:
PR has been around for a few days. Mozilla is pushing for this functionality and reported tests via other channels.

Issue #58, opened 2021-03-19
Binary shortlist loading from bytebuffer


jerinphilip:
@qianqianzhu You're supposed to follow @XapaJIaMnu's PR for model loading as a byte array (https://github.com/browsermt/bergamot-translator/pull/55), to provide the bytebuffer for shortlist (instead of model) as an argument at 5 places (or less). This would be `shortlist_memory` instead of `model_memory`. 

Whether you build against main or Nick's PR is upto you, if it were up to me I'd do it on top of Nick's PR and create a mess. Note that since both of you are editing at the same places to pass the bytearray, expect a (simple to fix) conflict which one of you will have to deal with depending on who gets their code into main first.

The shortlist loading happens here: https://github.com/browsermt/bergamot-translator/blob/f89c989b444841afd7458df97152ac1401f23750/src/translator/batch_translator.cpp#L21

All translation is dumped into the `BatchTranslator` abstraction (where the shorlist is loaded), with `Service` delegating the query to BatchTranslator to translator after preprocessing and getting the input ready as a batch. Due to bergamot's insistence of a bridge interface to potentially integrate multiple translators, you'll have to pass the `const void * ` (as argument) through:

`AbstractTranslationModel` -> `TranslationModel` -> `NonThreadedService` ->  `BatchTranslator` , [`ServiceBase`].  [] means optional, like Nick did.

I'd be grateful if you adjust it at `Service` as well, which is the multithreaded implementation. If too much, I'm competent enough to do it on my own once the remaining are ready. Advance apologies for this mess.

kpu:
Stop creating more of a mess.  Get Nick's PR in.

qianqianzhu:
@jerinphilip Can you first get https://github.com/browsermt/marian-dev/pull/26 merged? Then I can have the new `BinaryShortlistGenerator` to make changes.

kpu:
Merge into browsermt/marian-dev#26 done.  I've demanded @jerinphilip simplify Service (#61) to decrease the tax imposed by too many classes.

jerinphilip:
Hello @qianqianzhu. Your scope ends at demonstrating functionality at the boundary of `Service`, which is now collapsed into a single class with ifdefs. 

Probably keep a watch on how https://github.com/browsermt/bergamot-translator/issues/63 moves, as you'll need similar datatypes for shortlist as well?

Service handles interaction of translation components.

[...]

Issue #63, opened 2021-03-22
Pass size of model data


kpu:
We can't just pass a void * for the model data.  There needs to be some validation that the size allocated by the client is consistent with the size encoded in the model.

jerinphilip:
Does it make sense to abstract the datatype here `(const void *data, size_t size)` and reuse it for shortlist as well? I found the following in https://github.com/browsermt/bergamot-translator/issues/28#issuecomment-780177578.

```cpp
class MemoryGift {
  const char *data() const;
  size_t length() const;
  void early_free();
};
```

Does it make sense to bundle these for model, shortlist (seeing as to someone who'd use bytearray uses it for efficiency purposes) and unpack it at BatchTranslator (where these are loaded)? Vocabulary I'm not sure, but it's required at service as well.

abhi-agg:
@XapaJIaMnu Can we close this issue? I think https://github.com/browsermt/bergamot-translator/pull/94 was the last missing PR to close this one. Right?

PR #69, opened 2021-03-24
Enable binary shortlist loading from bytebuffer

qianqianzhu:
This PR addresses #58 to plumb binary shortlist into `bergamot-translator`. 

Lists of changes:
- update `service`
- update `batch_translator`
- update `service-cli-bytearray` to test shortlist

Manually tested `service-cli-bytearray`.


jerinphilip:
To save some time here, ignore the WASM build failures, it's unrelated to the changes in this PR. There are last 2 which are against full marian (not a reduced WASM one) which should pass though. The WASM build issues will be sorted once https://github.com/browsermt/marian-dev/pull/24 is resolved.

src/translator/byteArrayExample.cpp
```diff
@@ -42,4 +42,25 @@ void * getBinaryModelFromConfig(marian::Ptr<marian::Options> options) {
     }
 }
 
+std::vector<char> getBinaryShortlistFromFile(const std::string& filename){
```
jerinphilip, line 45: I want these functions in `marian::bergamot` for testing service. I see this as a marian::bergamot util function. Move the other functions (corresponding to model) as well in the process. Mozilla will directly supply const void* and perhaps size and they have no use for this function in a `bergamot::` namespace. For test-apps like `bergamot-translator-app`, they can pull it from marian anyway.

qianqianzhu: I see your point. Where should I put these function then?

src/translator/service.cpp
```diff
@@ -28,10 +28,12 @@ loadVocabularies(marian::Ptr<marian::Options> options) {
 namespace marian {
 namespace bergamot {
 
-Service::Service(Ptr<Options> options, const void *model_memory)
+Service::Service(Ptr<Options> options, const void *model_memory, const size_t model_memory_size,
+                 const void *shortlist_memory, const size_t shortlist_memory_size)
     : requestId_(0), vocabs_(std::move(loadVocabularies(options))),
       text_processor_(vocabs_, options), batcher_(options),
-      numWorkers_(options->get<int>("cpu-threads")), model_memory_(model_memory)
+      numWorkers_(options->get<int>("cpu-threads")), model_memory_(model_memory), model_memory_size_(model_memory_size),
```
jerinphilip, line 35: I'd prefer the following different constructors instead of a bloated constructor: 

1, `Service(Options)` which purely loads from an options object, no bytearray business. This way (service-cli, service-cli-test-bytearray) can help figure if it's bytearray that's causing problems.
2, `Service(Options, ...)` incorporating all memory address + size passing instantiations. Do not default, if someone's calling this constructor, the compiler should complain that values are not passed. I'd like to bundle ... together so it's an all or nothing setting. The bytebuffers are passed for efficiency, doing just one doesn't make sense.
3. `Service(std::string config, ....)`: Do what saves you time with this one. Probably parseOptions which is already done, and pass through the memory address + size.

Use 3 in your test-app to test functionality covering both 2 and 3, and as this constructor is used in TranslationModel at the moment.

The default for data-members should be what makes bytearray loads no-op.

void* and size_t should be bound in a struct together and the struct given to Service not provided separately.

src/translator/byteArrayExample.cpp
jerinphilip: Just adjust the namespace to `marian::bergamot`, propogate to where used and name the files in snake case (whatever marian adhere's to). As a general note, anything in `marian::bergamot` has to be named following marian-dev conventions.

src/translator/batch_translator.h
```diff
@@ -49,6 +50,9 @@ class BatchTranslator {
   std::vector<Ptr<Scorer>> scorers_;
   Ptr<data::ShortlistGenerator const> slgen_;
   const void * model_memory_;
+  const size_t model_memory_size_;
+  const void * shortlist_memory_;
+  const size_t shortlist_memory_size_;
```
jerinphilip, line 55: naming with camelcase and not snakecase (marian does this?). Also use the struct or a container of struct to bundle all.

src/translator/byteArrayExample.cpp
kpu: This isn't even a shortlist function.  It could be named load file to memory.  And it could be a method to some blob class.  

Also, use of std::vector<char> is discouraged because we want to have proper alignment.

src/translator/service.cpp
kpu: Agree we should have a const void * and size_t blob struct.  It should also own the memory.

src/translator/batch_translator.cpp
```diff
@@ -11,17 +11,30 @@ namespace bergamot {
...
-    : device_(device), options_(options), vocabs_(&vocabs), model_memory_(model_memory) {}
+                                 const void * model_memory,
+                                 const size_t model_memory_size,
+                                 const void * shortlist_memory,
+                                 const size_t shortlist_memory_size)
+    : device_(device), options_(options), vocabs_(&vocabs),
+    model_memory_(model_memory), model_memory_size_(model_memory_size),
+    shortlist_memory_(shortlist_memory), shortlist_memory_size_(shortlist_memory_size) {}
 
 void BatchTranslator::initialize() {
```
jerinphilip, line 22: I understand a separate initialize is what @kpu perhaps meant by 

> Jerin comes from the default-contructors-are-scary school 

This exists because at some point the objective I was directed at was to match marian-encoder and I had to offload model loading from files (which as you might already know is slow) into threads to change 48*x -> x when there were 48 threads. If this can use professional help, feel free. (This is my fault, I should have left a note).

src/translator/service.cpp
jerinphilip: I think @kpu has described what is desired from above more precisely already: https://github.com/browsermt/bergamot-translator/issues/63#issuecomment-804809630.

src/translator/batch_translator.cpp
```diff
@@ -11,17 +11,30 @@ namespace bergamot {
...
   // Initializes the graph.
   if (options_->hasAndNotEmpty("shortlist")) {
     int srcIdx = 0, trgIdx = 1;
     bool shared_vcb = vocabs_->front() == vocabs_->back();
-    slgen_ = New<data::LexicalShortlistGenerator>(options_, vocabs_->front(),
-                                                  vocabs_->back(), srcIdx,
-                                                  trgIdx, shared_vcb);
+    if (shortlist_memory_ != nullptr && shortlist_memory_size_ > 0) {
+        bool check = true;  // Whether to verify the shortlist content (not sure how to deal with this flag)
```
jerinphilip, line 28: Is this a todo? Are you asking? 

tl;dr: Always check, check logic within marian makes sense.

Service shouldn't be bothered to check the validity of shortlist, there should be a runtime abort from within the shortlist. So if you're getting a struct carrying *data and size, you *should* validate it during shortlist construction.

marian has ABORT for this. If model, shortlist shares similar traits then the validation routines can be moved into marian-dev with some common stuff between them.

If the doubt is regarding whether a `--check-shortlist` argument provided to user - no, always check.

qianqianzhu: Thanks. `check` flag is initially designed when the shortlist is loaded from file. The user (like Mozilla) can decide whether they want to check or not (e.g., for the first time use).  I was struggling with what to do with the binary buffer. I preassume that Mozilla should be very sure that they are passing the right memory to load the shortlist. 

What you said makes sense. I will keep `check` true.

kpu: check increases shortlist loading time because it has to go over all indices and check they are in range.  I specifically asked for it to be optional.  And honestly it should be.

jerinphilip: Single `--check-bytearray` (change to name you think is suitable) argument bound to `Service` through `Options`, loaded at configparser - for all, model, shortlist etc. Default to always check. Mozilla has to explicitly disable-for speed, baked into their WASM plumbing.

Below: 
https://github.com/browsermt/bergamot-translator/blob/34228d37bfcc44e36bf01d8ea70fbc2fce9d9ae9/src/translator/parser.h#L22-L25

src/translator/service.cpp
qianqianzhu: I will put `MemoryGift` class into `definitions.h`.

qianqianzhu: @jerinphilip it seems not possible to keep unified API boundaries (TranslationModel, AbstractTranslationModel) untouched. For the `service`'s constructors are changed (following what you said), I also need to adjust TranslationModel and AbstractTranslationModel.

jerinphilip: Hack the `string`-based constructor. It's only used by TranslationModel. Adding `MemoryGift` to WASM bound classes might have consequences (change wasm exports etc).

kpu: @abhi-agg @motin @andrenatal Need some help here on getting your side to give us those binary blobs.

abhi-agg:
@qianqianzhu Just a comment. You are seeing wasm CI failures because of the change in marian-dev submodule. I will work on https://github.com/browsermt/bergamot-translator/issues/70 that will ease this PR of yours 👍

src/translator/service.cpp
abhi-agg: @qianqianzhu You don't need to make any change to `AbstractTranslationModel` class as that class is gone (rebasing your PR on master will bring latest changes).

I can change the [constructor](https://github.com/browsermt/bergamot-translator/blob/main/src/TranslationModel.h#L36-L37) of `TranslationModel` class, push it to main branch and then you can rebase. However, the fastest way to unblock your PR would be if you just change this constructor to:

        TranslationModel(const std::string &config,
                           const void *model_memory = nullptr, size_t model_memory_size = 0,
                           const void *shortlist_memory = nullptr, size_t shortlist_memory_size = 0
                           )

You will not need to change anything elsewhere. Let me know if you are comfortable doing this change (and please feel free to ping me if it is not working for you 👍 )

Refactoring `const void *model_memory, size_t model_memory_size` with a `MemoryGift` struct or something can be done later once I generate JS bindings.

The bigger blocker of getting your PR approved is to fix the wasm CI failures and that has higher priority for me (compared to this small refactoring change) as I mentioned [here](https://github.com/browsermt/bergamot-translator/pull/69#issuecomment-806534147).

qianqianzhu: Thanks! I hacked a way to solve the conflict of refactoring MemoryGift and TranslationModel. I will keep TranslationModel part untouched to not get in your way.

qianqianzhu:
CI failures is mainly caused by my incomplete work on service(Option) constructor.

src/translator/definitions.h
```diff
@@ -21,6 +21,24 @@ template <class T, typename... Args> UPtr<T> UNew(Args &&... args) {
...
+  :memory_(memory),memorySize_(memorySize){}
+  const void* data() const{
+    return memory_;
+  }
+  const size_t length() const{
+    return memorySize_;
+  }
+  void early_free(){
+    free((char *)memory_);
+  }
```
jerinphilip, line 39: > And it could be a method to some blob class.

Weren't the loads supposed to be consolidated here per previous review?

kpu: Not sure what you're asking.  It's Mozilla's problem to make the (aligned) gift.

src/translator/service.cpp
```diff
@@ -28,10 +28,10 @@ loadVocabularies(marian::Ptr<marian::Options> options) {
 namespace marian {
 namespace bergamot {
 
-Service::Service(Ptr<Options> options, const void *model_memory)
+Service::Service(Ptr<Options> options, MemoryGift modelMemory, MemoryGift shortlistMemory)
     : requestId_(0), vocabs_(std::move(loadVocabularies(options))),
       text_processor_(vocabs_, options), batcher_(options),
-      numWorkers_(options->get<int>("cpu-threads")), model_memory_(model_memory)
+      numWorkers_(options->get<int>("cpu-threads")), modelMemory_(&modelMemory),shortlistMemory_(&shortlistMemory)
```
jerinphilip, line 34: `MemoryGift` is taken by value, does it even live past the constructor? Isn't this a potential issue?

src/translator/definitions.h
jerinphilip: replacement server (if deployed as a service and has a swap models operation) can also use the binary files correct, if load speeds are improved. It's also my problem to make `MemoryGift` in that case?

I'm overall confused about who owns MemoryGift. Does mozilla swap in swap out models in which case they do. If we are, there are a lot of leaks in the current code, is there not - due to ownership issues.

src/translator/byte_array_util.cpp
```diff
@@ -17,29 +18,41 @@ void * getBinaryFile(std::string path) {
         std::exit(1);
     }
     void *result;
-    int fail = posix_memalign(&result, 256, length);
-    if (fail) {
-        std::cerr << "Failed to allocate aligned memory." << std::endl;
-        std::exit(1);
+    if (isModelFile) {
```
jerinphilip, line 21: Seeing as to alignment to 256 is checked during model construction, can we shift this responsibility to the model constructor (from blob) and make this function cleaner? Maybe move as an std::string (fpath) based constructor to `MemoryGift`?

src/translator/byte_array_util.cpp
```diff
@@ -17,29 +18,41 @@ void * getBinaryFile(std::string path) {
...
+        int fail = posix_memalign(&result, 256, length);
+        if (fail) {
+            std::cerr << "Failed to allocate aligned memory." << std::endl;
+            std::exit(1);
+        }
```
jerinphilip, line 26: Seeing as to alignment to 256 is checked during model construction, can we shift this responsibility to the model constructor (from blob) and make this function cleaner? Maybe move as an std::string (fpath) based constructor to `MemoryGift`?

src/translator/service.cpp
qianqianzhu: Shall I pass in MemoryGift pointers instead?

jerinphilip: I'm in favour of pointers for the scope of this PR. The contents referred to by this pointer should not be modified by `Service` or `BatchTranslator` (have compiler complain if attempted to by some const pointer mechanism). This makes it clear from types that we do not own this memory.

qianqianzhu: OK. Will switch to `const MemoryGift *`.

src/translator/byte_array_util.cpp
qianqianzhu: As this alignment will be checked again during the translation, it does not hurt to remove this part. Move std::string (fpath) based constructor to MemoryGift is a bit dangerous. I was trying to keep IO apart from data.

src/translator/byte_array_util.cpp
```diff
@@ -45,14 +45,13 @@ MemoryGift getBinaryModelFromConfig(marian::Ptr<marian::Options> options) {
...
-    if (vals.empty()){
-        std::cerr << "No path to shortlist file given" << std::endl;
-        std::exit(1);
-    }
-    std::string filename = vals[0];
-    return loadFileToMemory(filename,false);
+  std::vector<std::string> shortlist = options->get<std::vector<std::string>>("shortlist");
+  if (shortlist.empty()) {
+    std::cerr << "No path to shortlist file is given." << std::endl;
+    std::exit(1);
```
jerinphilip, line 51: Do we need to if-else this? `options->get<std::vector<std::string>>` will abort during runtime when a missing key is attempted to access with informative error messages, so this if else looks a bit redundant.

src/translator/byte_array_util.cpp
```diff
@@ -7,26 +7,26 @@ namespace marian {
...
-        std::exit(1);
+  std::ifstream is (path, std::ifstream::binary);
+  uint64_t length = 0; // Determine the length of file in bytes
+  if (is) {
+    is.seekg(0, is.end);
+    length = is.tellg();
+    is.seekg(0, is.beg);
+  } else {
+    std::cerr << "Failed opening file stream: " << path << std::endl;
+    std::exit(1);
```
jerinphilip, line 18: No that we're in `marian::bergamot`, `ABORT` macro? You can probably use `ABORT_IF` and a single line well.

src/translator/service.cpp
```diff
@@ -40,21 +63,43 @@ Service::Service(Ptr<Options> options, MemoryGift modelMemory, MemoryGift shortl
...
+#ifndef WASM
+      // 0 elements in PCQueue is illegal and can lead to failures. Adding a
+      // guard to have at least one entry allocated. In the single-threaded
+      // case, while initialized pcqueue_ remains unused.
+      ,
+      pcqueue_(std::max<size_t>(1, numWorkers_))
+#endif
+{
+  Ptr<Options> options = parseOptions(config);
+  initialize(options);
```
jerinphilip, line 81: if initialize -> build_translators (which passes the modelMemory_), the code below this becomes dead code right?

src/translator/service.cpp
```diff
@@ -40,21 +63,43 @@ Service::Service(Ptr<Options> options, MemoryGift modelMemory, MemoryGift shortl
...
+    modelMemory_ = modelGift;
+  }
+
+  if(shortlistMemory!= nullptr) {
+    // Hacks to obtain shortlist memory size as this will be checked during construction
+    size_t shortlistMemorySize = sizeof(uint64_t) * (6 + *((uint64_t*)shortlistMemory+4))
+            + sizeof(uint32_t) * *((uint64_t*)shortlistMemory+5);
+    // MemoryGift* should be freed once it is no longer used; here is a hack to make TranslationModel works
+    const MemoryGift* shortlistGift = new MemoryGift(shortlistMemory, shortlistMemorySize);
+    shortlistMemory_ = shortlistGift;
```
jerinphilip, line 94: What you're doing is `f: void* -> MemoryGift*`.

call the two `MemoryGift* hackShortList(void*)`, `MemoryGift* hackMemory(void*)` for now and move to (constructor's) member initialization list. Easy to adjust later, and we know the blocks are a hack.

This way you will have it ready before `initialize()` is called as well.

Question, not review: Do we have code to validate model-size yet?

jerinphilip: The following equivalency makes our life easier, does it not? 

```cpp
Service(std::string config, const void* modelMemory, 
const void *shortlistMemory)
    : Service(parseOptions(config), hackMemory(modelMemory),
    hackShortList(shortlistMemory)) {} 
```

abhi-agg:
> @qianqianzhu Just a comment. You are seeing wasm CI failures because you changed the marian-dev submodule. I will work on #70 that will ease this PR of yours. After [this issue](https://github.com/browsermt/bergamot-translator/issues/70) gets resolved, you will not need to deal with these failures 👍

@qianqianzhu This is done. Latest changes are in main branch 👍

src/translator/service.cpp
jerinphilip: While we're at it, it may make sense to distill all construction routine into the most enriched one:

```cpp
Service(Ptr<Options> options):
    Service(options, nullptr, nullptr)
```

No default arguments though, because it'll end up creating many codepaths outside the API which will need change if we change stuff internally.

qianqianzhu: > Question, not review: Do we have code to validate model-size yet?

No.

qianqianzhu: For the comments for Service constructors, I will address in later commit (I knew current code refactoring is just a comprise but the best design).

src/translator/byte_array_util.cpp
qianqianzhu: I thought ABORT was not there. Will change soon.

src/translator/byte_array_util.cpp
jerinphilip: I'd prefer to have the validation responsibility shifted to the model constructor and make this `loadFileToMemory(fname)`. I agree on IO vs data.

src/translator/service.cpp
kpuatfb: You are not allowed to just pass const void *.  Sizing information is required to not segfault.  We may not have code to validate model size at this time, but it should still be mandatory to pass so it can be checked later.

kpu: Passing size is mandatory to prevent segfault.  Even if we can't validate it at this time.

src/translator/byte_array_util.cpp
qianqianzhu: It could be the case that the user specifies the key, but not the value. (Users always break things.)

qianqianzhu:
Trying to resolve conflicts with submodule marian-dev. But it seems @abhi-agg you are working on it now. I will fix this once the shortlist plumbing is done.

abhi-agg:
> Trying to resolve conflicts with submodule marian-dev. But it seems @abhi-agg you are working on it now. I will fix this once the shortlist plumbing is done.

@qianqianzhu I didn't understand you here. Which changes are you talking about?

qianqianzhu:
> > Trying to resolve conflicts with submodule marian-dev. But it seems @abhi-agg you are working on it now. I will fix this once the shortlist plumbing is done.
> 
> @qianqianzhu I didn't understand you here. Which changes are you talking about?

That was my misunderstanding.  Everything related to shortlist plumbing is done. Submodule conflicts are fixed and CI build succeeded.

jerinphilip:
> Everything related to shortlist plumbing is done

@qianqianzhu Can you please trigger the following from your side now so we can close this PR at the earliest.

- [ ] Merge your updated changes for plumbing to browsermt/marian-dev@master
- [ ] Update submodule here to point to the master (which now has the shortlist required changes)
- [ ] Mark this explictly as complete?

src/translator/definitions.h
jerinphilip: This is resolved in another thread. @qianqianzhu wanted to keep IO and the struct separate. Reopen if not the case.

src/translator/service.h
```diff
@@ -40,16 +62,18 @@ class Service {
...
+/// Construct Service from a string configuration.
   /// @param [in] config string parsable as YAML expected to adhere with marian
   /// config
   /// @param [in] model_memory byte array (aligned to 64!!!) that contains the
   /// bytes of a model.bin. Optional, defaults to nullptr when not used
   explicit Service(const std::string &config,
-                   const void *model_memory = nullptr)
-      : Service(parseOptions(config), model_memory) {}
+                   const void* modelMemory = nullptr, const void* shortlistMemory = nullptr)
+      : Service(parseOptions(config), hackModel(modelMemory), hackShortLis(shortlistMemory)) {}
```
jerinphilip, line 76: I want to be picky about `hackShotLis`, but won't insist since this function is meant to go away anyway.

src/translator/service.h
```diff
@@ -86,7 +110,9 @@ class Service {
 
   /// Number of workers to launch.
   size_t numWorkers_;        // ORDER DEPENDENCY (pcqueue_)
-  const void *model_memory_; /// Model memory to load model passed as bytes.
+  const MemoryGift* modelMemory_ = nullptr; /// Model memory to load model passed as bytes.
```
jerinphilip, line 113: `modelMemory_{nullptr}` nit? same for shortlist?

src/translator/byte_array_util.cpp
```diff
@@ -0,0 +1,41 @@
...
+namespace marian {
+namespace bergamot {
+
+MemoryGift loadFileToMemory(const std::string& path, bool isModelFile){
+  std::ifstream is (path, std::ifstream::binary);
+  ABORT_IF(is.bad(), "Failed opening file stream: {}", path);
+  uint64_t length = 0; // Determine the length of file in bytes
+  is.seekg(0, is.end);
+  length = is.tellg();
+  is.seekg(0, is.beg);
```
kpu, line 15: What if file isn't seekable?  This will fail silently.

src/translator/byte_array_util.cpp
```diff
@@ -0,0 +1,41 @@
...
+  }
+  is.read(static_cast<char *>(result), length);
+  return MemoryGift(result, length);
+}
+
+MemoryGift getBinaryModelFromConfig(marian::Ptr<marian::Options> options) {
+    std::vector<std::string> models = options->get<std::vector<std::string>>("models");
+    ABORT_IF(models.size() != 1, "Loading multiple binary models is not supported for now as it is not necessary.");
+    marian::filesystem::Path modelPath(models[0]);
+    ABORT_IF(modelPath.extension() != marian::filesystem::Path(".bin"), "Non binary models cannot be loaded as a byte array.");
```
kpu, line 30: Maybe say it should end with .bin

app/bergamot-translator-app-bytearray.cpp
```diff
@@ -62,7 +62,7 @@ int main(int argc, char **argv) {
   }
 
   // Clear the memory used for the byte array
-  free(model_bytes); // Ideally, this should be done after the translation model has been gracefully shut down.
+  free((char *)model_bytes); // Ideally, this should be done after the translation model has been gracefully shut down.
```
kpu, line 65: Why was this cast necessary?

src/translator/definitions.h
```diff
@@ -21,6 +21,24 @@ template <class T, typename... Args> UPtr<T> UNew(Args &&... args) {
 
 template <class T> UPtr<T> UNew(UPtr<T> p) { return UPtr<T>(p); }
 
+class MemoryGift{
```
kpu, line 24: I'm sorry to change my mind here, but the memory management is looking really questionable here.  It seems MemoryGift should own the memory.  

Moreover, to support multiple platforms with alignment there's actually different free.  

Might I suggest https://github.com/kpu/intgemm/blob/master/intgemm/aligned.h but one can get rid of the template argument if it's always storing void.  In theory you have this in the submodule.  Could copy here though.

src/translator/definitions.h
```diff
@@ -21,6 +21,24 @@ template <class T, typename... Args> UPtr<T> UNew(Args &&... args) {
...
+public:
+  MemoryGift(const void *memory,const size_t memorySize)
+  :memory_(memory),memorySize_(memorySize){}
+  const void* data() const{
+    return memory_;
+  }
+  const size_t length() const{
+    return memorySize_;
+  }
+  void early_free(){
```
kpu, line 37: clear()?

src/translator/byte_array_util.cpp
```diff
@@ -0,0 +1,41 @@
...
+  std::ifstream is (path, std::ifstream::binary);
+  ABORT_IF(is.bad(), "Failed opening file stream: {}", path);
+  uint64_t length = 0; // Determine the length of file in bytes
+  is.seekg(0, is.end);
+  length = is.tellg();
+  is.seekg(0, is.beg);
+
+  void *result = new char[length];  // To avoid the memory is destroyed after the function call
+  if (isModelFile) {
+    int fail = posix_memalign(&result, 256, length);
```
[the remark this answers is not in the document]
kpu: https://github.com/kpu/intgemm/blob/master/intgemm/aligned.h will also resolve this.

kpu:
@abhi-agg @motin I'd appreciate if you took a look at https://github.com/kpu/intgemm/blob/master/intgemm/aligned.h and declared if you're happy with passing memory to us using a class similar to that.  If I don't hear back in a few days we'll go forward with it.

abhi-agg:
> @abhi-agg @motin I'd appreciate if you took a look at https://github.com/kpu/intgemm/blob/master/intgemm/aligned.h and declared if you're happy with passing memory to us using a class similar to that. If I don't hear back in a few days we'll go forward with it.

I think a class similar to this should be fine. A method/constructor that populates the internal buffer with contents would be convenient for us on JS side.
(Something similar to: `AlignedVector::AlignedVector(std::string contents)`

kpu:
> I think a class similar to this should be fine. A method/constructor that populates the internal buffer with contents would be convenient for us on JS side.
> (Something similar to: `AlignedVector::AlignedVector(std::string contents)`

Why the copy?  Can you just allocate `AlignedVector` and write to `begin()`?

abhi-agg:
> > I think a class similar to this should be fine. A method/constructor that populates the internal buffer with contents would be convenient for us on JS side.
> > (Something similar to: `AlignedVector::AlignedVector(std::string contents)`
> 
> Why the copy? Can you just allocate `AlignedVector` and write to `begin()`?

(As per my understanding) If the memory allocation is happening explicitly in C++ side then we can't avoid this copy (or at least it will require few hacks to avoid it).

On the other hand, if the allocation is left for JS then we can avoid this copy.

@motin might have some context on it.

kpu:
> > > I think a class similar to this should be fine. A method/constructor that populates the internal buffer with contents would be convenient for us on JS side.
> > > (Something similar to: `AlignedVector::AlignedVector(std::string contents)`
> > 
> > 
> > Why the copy? Can you just allocate `AlignedVector` and write to `begin()`?
> 
> (As per my understanding) If the memory allocation is happening explicitly in C++ side then we can't avoid this copy (or at least it will require few hacks to avoid it).
> 
> On the other hand, if the allocation is left for JS then we can avoid this copy.
> 
> @motin might have some context on it.

Then we'll use free() for now and you can tell us what the appropriate call to free things is later.  Keep in mind we should be streaming this from js anyway so stuff will change.  Also my understanding is one can export a pointer from web assembly and javascript can write to it https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/Memory

src/translator/byte_array_util.cpp
```diff
@@ -0,0 +1,38 @@
...
+std::string getShortlistFileFromConfig(marian::Ptr<marian::Options> options){
+  auto shortlist = options->get<std::vector<std::string>>("shortlist");
+  ABORT_IF(shortlist.empty(), "No path to shortlist file is given.");
+  return shortlist[0];
+}
+
+size_t getMemorySizeFromFile(const std::string& filename, bool isModelFile){
+  uint64_t fileSize = filesystem::fileSize(filename);
+  if (isModelFile){  // As model memory need to be 256-byte aligned (AlignedMemory only promises 64-byte aligned memory)
+    fileSize = fileSize%256==0? fileSize : (fileSize/256+1)*256;
```
kpu, line 32: I added a constructor argument for alignment.

src/translator/service.h
```diff
@@ -38,18 +63,21 @@ class Service {
 
 public:
   /// @param options Marian options object
-  /// @param model_memory byte array (aligned to 64!!!) that contains the bytes
+  /// @param modelMemory byte array (aligned to 64!!!) that contains the bytes
```
kpu, line 66: Aligned to 256?

src/translator/service.h
```diff
@@ -38,18 +63,21 @@ class Service {
...
   /// of a model.bin. Optional, defaults to nullptr when not used
-  explicit Service(Ptr<Options> options, const void *model_memory = nullptr);
+  /// @param shortlistMemory byte array of shortlist
+  explicit Service(Ptr<Options> options, const AlignedMemory* modelMemory, const AlignedMemory* shortlistMemory);
```
kpu, line 69: It's really looking like we will own this memory, so I'd move them in here as members that might happen to be empty / null pointers.

src/translator/service.h
```diff
@@ -86,7 +114,9 @@ class Service {
 
   /// Number of workers to launch.
   size_t numWorkers_;        // ORDER DEPENDENCY (pcqueue_)
-  const void *model_memory_; /// Model memory to load model passed as bytes.
+  const AlignedMemory* modelMemory_{nullptr}; /// Model memory to load model passed as bytes.
```
kpu, line 117: Own it.  
AlignedMemory modelMemory_;

kpu (changes requested):
I've changed AlignedVector to make life easier for you https://github.com/kpu/intgemm/blob/master/intgemm/aligned.h

To pull in from intgemm needs a submodule update.  It's also weird to pull it in all the way from intgemm to define as part of the interface, since we promised the interface would not depend on internals.  Best to copy, which will also save you blocking on a submodule update.

src/translator/service.cpp
```diff
@@ -54,7 +55,7 @@ void Service::build_translators(Ptr<Options> options, size_t numTranslators) {
   translators_.reserve(numTranslators);
   for (size_t cpuId = 0; cpuId < numTranslators; cpuId++) {
     marian::DeviceId deviceId(cpuId, DeviceType::cpu);
-    translators_.emplace_back(deviceId, vocabs_, options, model_memory_);
+    translators_.emplace_back(deviceId, vocabs_, options, std::move(modelMemory_), std::move(shortlistMemory_));
```
jerinphilip, line 58: Service owns the memory. BatchTranslator gets reference collected as pointer.

Only the first BatchTranslator inits here. Others encounter illegal memories. 

I think. Run and check? More than 0 threads variations of service cli.

qianqianzhu: Agree. It was my misunderstanding here.

src/translator/service.h
```diff
@@ -86,7 +116,9 @@ class Service {
 
   /// Number of workers to launch.
   size_t numWorkers_;        // ORDER DEPENDENCY (pcqueue_)
-  const void *model_memory_; /// Model memory to load model passed as bytes.
+  AlignedMemory modelMemory_; /// Model memory to load model passed as bytes.
```
jerinphilip, line 119: I prefer ORDER DEPENDENCY (also on which other member) being explicit so cautious move around happens. https://isocpp.org/wiki/faq/ctors#order-dependency-in-members. The present ordering is correct. If tomorrow I decollapse Service into a single threaded version (not that this is happening, just one example where swapping unknowingly will create issues), I would like to immediately know that one member has to be initialized before the other when copy-pasting. Both memories should be init before BatchTranslator (this happens already, just indicate ORDER DEPENDENCY explicitly).

We should probably enable `-Werror=reorder` like marian has (but that's for another PR/task).

src/translator/byte_array_util.cpp
```diff
@@ -0,0 +1,34 @@
...
+#include <iostream>
+
+namespace marian {
+namespace bergamot {
+
+AlignedMemory loadFileToMemory(const std::string& path, bool isModelFile){
+  uint64_t fileSize = filesystem::fileSize(path);
+  io::InputFileStream in(path);
+  ABORT_IF(in.bad(), "Failed opening file stream: {}", path);
+  size_t alignment = isModelFile? 256 : 64;
```
jerinphilip, line 12: Can't this be accepted as an argument (alignment) instead of isModelFile and generalize this function a bit more? This doesn't do anything specifically to model does it, it only switches alignment, which can be 

```cpp
AlignedMemory loadFileToMemory(const std::string& path, size_t alignment)
```

src/translator/service.h
```diff
@@ -38,18 +65,21 @@ class Service {
 
 public:
   /// @param options Marian options object
-  /// @param model_memory byte array (aligned to 64!!!) that contains the bytes
+  /// @param modelMemory byte array (aligned to 256!!!) that contains the bytes
   /// of a model.bin. Optional, defaults to nullptr when not used
-  explicit Service(Ptr<Options> options, const void *model_memory = nullptr);
+  /// @param shortlistMemory byte array of shortlist
```
jerinphilip, line 70: Somewhere it mentioned alignment to 64, maybe add to make documentation consistent?

src/translator/service.h
```diff
@@ -38,18 +65,21 @@ class Service {
...
+  explicit Service(Ptr<Options> options, AlignedMemory modelMemory, AlignedMemory shortlistMemory);
+
+  explicit Service(Ptr<Options> options) : Service(options, AlignedMemory(), AlignedMemory()){}
```
jerinphilip, line 73: A slight worry I have here is that it's not straighforward what `AlignmentMemory()` is from reading (I'm guessing this allocates memory/set `nullptr` and set's size to 0?). Since we have our own `alignment.h` now, would it be preferable to have a [named constructor](https://isocpp.org/wiki/faq/ctors#named-ctor-idiom) - `AlignedMemory::EmptyMemory()` or something?

src/translator/batch_translator.cpp
```diff
@@ -30,12 +42,10 @@ void BatchTranslator::initialize() {
   graph_->setDevice(device_);
   graph_->getBackend()->configureDevice(options_);
   graph_->reserveWorkspaceMB(options_->get<size_t>("workspace"));
-  if (model_memory_) { // If we have provided a byte array that contains the model memory, we can initialise the model from there, as opposed to from reading in the config file
-    if ((uintptr_t)model_memory_ % 256 != 0) {
-      std::cerr << "The provided memory is not aligned to 256 bytes and will crash when vector instructions are used on it." << std::endl;
-      exit(1);
-    }
-    const std::vector<const void *> container = {model_memory_}; // Marian supports multiple models initialised in this manner hence std::vector. However we will only ever use 1 during decoding.
+  if (modelMemory_.size() > 0 && modelMemory_.begin() != nullptr) { // If we have provided a byte array that contains the model memory, we can initialise the model from there, as opposed to from reading in the config file
```
jerinphilip: Have `modelMemory_.empty()` consistent with `Alignment::EmptyMemory()`. The responsibility to maintain this consistency is now shifted to `AlignedMemory`.

src/translator/service.h
qianqianzhu: I do not understand here. modelMemory_ and shortlistMemory_ are two independent members. The order of initialisation does not break the constructor. They can be used together or separately, right?

kpu: Jerin has a peculiar coding style that attempts to annotate whether a subsequent member variable depends on a previous one.  I can see the value in this annotation in ensuring correct destruction order.  In this case he appears to be referring to the workers depending on the continued existence of `modelMemory_`.

src/translator/aligned.h
```diff
@@ -0,0 +1,70 @@
+#pragma once
+#include <cstdlib>
+#include <new>
+#ifdef _MSC_VER
+#include <malloc.h>
+#endif
+
+// Aligned simple vector.
+
+namespace intgemm {
```
kpu, line 10: Different namespace please, otherwise we get the same object defined twice in the binary.

src/translator/service.h
jerinphilip: Both memories are dependent on BatchTranslators (workers are std::threads) just to avoid confusion.

src/translator/service.h
kpu: No.  Default constructors are fine.

src/translator/batch_translator.cpp
jerinphilip: (@kpu comment resolves this).

src/translator/byte_array_util.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+AlignedMemory loadFileToMemory(const std::string& path, size_t alignment){
+  uint64_t fileSize = filesystem::fileSize(path);
+  io::InputFileStream in(path);
+  ABORT_IF(in.bad(), "Failed opening file stream: {}", path);
+  AlignedMemory alignedMemory(fileSize, alignment);
+  in.read(reinterpret_cast<char *>(alignedMemory.begin()), fileSize);
+  ABORT_IF(alignedMemory.size() != fileSize, "Error reading file {}", path);
+  return alignedMemory;
+}
```
jerinphilip, line 16: This is now passed around as return (is return copy or by reference). I'm always a bit scared and I write 
```cpp
void loadFileToMemory(const std::string& path, size_t alignment, AlignedMemory &out)
```

Is ownership proper here, since `AlignedMemory` I'm guessing behaves similar to an `std::string`?

kpu: https://stackoverflow.com/questions/4986673/c11-rvalues-and-move-semantics-confusion-return-statement

kpu (dismissed):

qianqianzhu:
Sorry have to update the submodule marian-dev here to master (otherwise the regression tests cannot be built).

jerinphilip:
@XapaJIaMnu's bytearray modifications seem intact after the `Service` constructor changes. The shortlist tests (also service-cli with bytearray) are segfaulting ([log](http://ix.io/2UKg)). `bergamot-translator-app` is incompatible after the stdin reading change in a recent PR and nobody cleaning up tests. The byte-array counterpart existing is convenient (and sufficient).

<details> <summary> Basic Regression tests: [2/4] </summary> 
<p>

```
jphilip@var:~/code/bergamot-translator-test-scripts$ MARIAN=/mnt/Storage/jphilip/bergamot-build/ TIMEOUT=1m BRT_THREADS=48 BRT_EXPECTED_MAXTIME=300 ./run_brt.sh
[04/01/2021 16:01:47] Running on var as process 66407
[04/01/2021 16:01:47] Using Marian binary: /mnt/Storage/jphilip/bergamot-build/marian
[04/01/2021 16:01:47] Version: v1.9.56 9337105c 2021-03-24 13:34:02 +0100
[04/01/2021 16:01:47] Build type: Release
[04/01/2021 16:01:47] Using compiler: /usr/bin/c++
[04/01/2021 16:01:47] Using MKL: COMPILE_CPU=on
[04/01/2021 16:01:47] Using CUDNN:
[04/01/2021 16:01:47] Using SentencePiece: USE_SENTENCEPIECE=on
[04/01/2021 16:01:47] Using FBGEMM:
[04/01/2021 16:01:47] Unit tests: COMPILE_TESTS=on
[04/01/2021 16:01:47] Using CUDA visible devices:
[04/01/2021 16:01:47] Using number of GPU devices: 1
[04/01/2021 16:01:47] Using time out: 1m
[04/01/2021 16:01:47] Checking directory: tests
[04/01/2021 16:01:47] Checking directory: tests/basic
[04/01/2021 16:01:47] Running tests/basic/test_bergamot_translator_app_intgemm_8bit.bytearrayload.cpu-threads.0.sh ... OK
[04/01/2021 16:01:49] Test took 00:00:1.605s
[04/01/2021 16:01:49] Running tests/basic/test_bergamot_translator_app_intgemm_8bit.cpu-threads.0.sh ... timed out
[04/01/2021 16:02:49] Test took 00:01:0.012s
[04/01/2021 16:02:49] Running tests/basic/test_service-cli_intgemm_8bit.bytearrayload.cpu-threads.4.sh ... failed
[04/01/2021 16:02:52] Test took 00:00:2.515s
[04/01/2021 16:02:52] Running tests/basic/test_service-cli_intgemm_8bit.cpu-threads.4.sh ... OK
[04/01/2021 16:02:55] Test took 00:00:2.840s
---------------------
Failed:
  - tests/basic/test_service-cli_intgemm_8bit.bytearrayload.cpu-threads.4.sh
Timed out:
  - tests/basic/test_bergamot_translator_app_intgemm_8bit.cpu-threads.0.sh
Logs:
  - /home/jphilip/code/bergamot-translator-test-scripts/tests/basic/test_service-cli_intgemm_8bit.bytearrayload.cpu-threads.4.sh.log
```
</p>
</details>

Speed tests are okay.
<details> <summary> Speed Tests [1/1] </summary> 
<p>

```
jphilip@var:~/code/bergamot-translator-test-scripts$ MARIAN=/mnt/Storage/jphilip/bergamot-build/ TIMEOUT=45m BRT_THREADS=48 BRT_EXPECTED_MAXTIME=300 ./run_brt.sh speed-tests/test_wngt20_perf.sh
[04/01/2021 16:06:47] Running on var as process 66694
[04/01/2021 16:06:47] Using Marian binary: /mnt/Storage/jphilip/bergamot-build/marian
[04/01/2021 16:06:48] Version: v1.9.56 9337105c 2021-03-24 13:34:02 +0100
[04/01/2021 16:06:48] Build type: Release
[04/01/2021 16:06:48] Using compiler: /usr/bin/c++
[04/01/2021 16:06:48] Using MKL: COMPILE_CPU=on
[04/01/2021 16:06:48] Using CUDNN:
[04/01/2021 16:06:48] Using SentencePiece: USE_SENTENCEPIECE=on
[04/01/2021 16:06:48] Using FBGEMM:
[04/01/2021 16:06:48] Unit tests: COMPILE_TESTS=on
[04/01/2021 16:06:48] Using CUDA visible devices:
[04/01/2021 16:06:48] Using number of GPU devices: 1
[04/01/2021 16:06:48] Using time out: 45m
[04/01/2021 16:06:48] Checking directory: speed-tests
[04/01/2021 16:06:48] Running speed-tests/test_wngt20_perf.sh ... OK
[04/01/2021 16:15:49] Test took 00:09:1.696s
---------------------
Ran 1 tests in 00:09:1.749s, 1 passed, 0 skipped, 0 failed

```
</p>
</details>

jerinphilip (approved):
Well, service-cli and  speed-tests are intact, why am I even waiting to approve this on imposing weight of tests people who should care about broke?

kpu (approved):
This hack thing must go.  But I can't edit the code so I'm gonna put this in and then fix it.

PR #71, opened 2021-03-24
Remove AbstractTranslationModel class and its references

abhi-agg:
- The commit removes AbstractTranslationModel class and all its references from the code base
- Tested by running all the apps to perform inference

The PR doesn't change/add any functionality. It is just a clean up.


kpu (approved):

jerinphilip (approved):
I understand editing further might break API for downstream (extension) so we will have to bear with this intermediate for a while, hence approving.

Issue #70, opened 2021-03-24
Update marian-dev submodule to use its master branch


abhi-agg:
Currently, the marian-dev submodule of this repository links to [wasm branch of marian-dev](https://github.com/browsermt/marian-dev/tree/wasm)

Change it back to `master` branch.

jerinphilip:
Are you on this @abhi-agg? I can couple this a workflow file rearrangement now that are no master / wasm differences.

abhi-agg:
I can do this but @XapaJIaMnu 's [PR](https://github.com/browsermt/marian-dev/pull/27) should be merged to marian-dev master before we make this switch. Otherwise, things will not work.

@XapaJIaMnu Could you please merge your change to browsermt/marian-dev's master? We can start this task as soon as you merge it.

PR #73, opened 2021-03-25
Update marian-dev submodule to point to its master

abhi-agg:
The PR fixes https://github.com/browsermt/bergamot-translator/issues/70

 - Earlier it was using 'wasm' branch
 - CMakefile changes
 - Github workflow change


jerinphilip (changes requested):
Can we have a rename of the workflow files w.r.t what they exactly do? Also requesting clarifications (which should be obvious from filenames once this is merged). 

- ubuntu.yml and macos.yml are full marian-capable tests (these were almost replicas of what exists upstream). They enable tests, enable a lot of extras which we are not expected to break in `marian::bergamot` for a future of the request response for a blob of text API being taken for a server upstream. We can probably suffix this with an `-upstream-marian-compatible`. 
- What was native? I understand we will have an ubuntu-wasm and macos-wasm from here-on, yes, which only builds for WASM target in addition to those with full marian-capability? What is the 'custom' bit in files now?

abhi-agg:
Here is the proposal then:

`native-ubuntu-full-marian`  or  `native-ubuntu`
`native-mac-full-marian`      or    `native-mac`

`wasm-ubuntu-custom-marian`
`wasm-mac-custom-marian`

We need to add `custom` for wasm workflow file names because wasm compilation doesn't use all sources of marian but a subset of these sources. Consequently, if we are adding `custom` to wasm workflow file names then it makes sense to add `full` to native ones just to be specific. I am fine with omitting `full-marian` from native ones altogether.

Let's agree just now and I will make the change 👍

jerinphilip:
Let's go for descriptive filenames?

* `<target-arch>-<nature-of-marian>-<runner-os>` 
* `(wasm|native)-(full_marian|adapted_marian)-(ubuntu|macos)`.

abhi-agg:
> Let's go for descriptive filenames?
> 
>     * `<target-arch>-<nature-of-marian>-<runner-os>`
> 
>     * `(wasm|native)-(full_marian|adapted_marian)-(ubuntu|macos)`.
> 
> 
> (or custom in place of adapted, if that's what you prefer).

@jerinphilip Wait. I don't see any difference b/w your proposal and mine. You just changed the positions of `runner-os` and replaced a hyphen with an underscore in `nature-of-marian`.

Anyway, I am not picky on this one. I will make the change 👍

jerinphilip (approved):

motin (approved):

PR #74, opened 2021-03-26
Marian submodule update

abhi-agg:
- Updates marian submodule (brings loading models from byte array change too which is required here)
- Adds workflows for building custom marian natively on mac and ubuntu

### DIFF
--- a/.github/workflows/native-custom_marian-mac.yml
+++ b/.github/workflows/native-custom_marian-mac.yml
@@ -1,0 +1,33 @@
+name: Native (Custom Marian) MacOS
+
+on:
+  push:
+    branches: [ main, ci-sandbox ]
+  pull_request:
+    branches: [ main, ci-sandbox ]
+
+jobs:
+  build-macos:
+    name: MacOS
+    runs-on: macos-10.15
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+      with:
+        submodules: recursive
+
+    - name: Configure CMake
+      run: |
+        mkdir -p build
+        cd build
+        cmake ..
+
+    - name: Compile
+      working-directory: build
+      run: make -j2
+
+    - name: Print versions
+      working-directory: build
+      run: |
+        ./app/bergamot-translator-app --version
--- a/.github/workflows/native-custom_marian-ubuntu.yml
+++ b/.github/workflows/native-custom_marian-ubuntu.yml
@@ -1,0 +1,33 @@
+name: Native (Custom Marian) Ubuntu
+
+on:
+  push:
+    branches: [ main, ci-sandbox ]
+  pull_request:
+    branches: [ main, ci-sandbox ]
+
+jobs:
+  build-macos:
+    name: Ubuntu
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+      with:
+        submodules: recursive
+
+    - name: Configure CMake
+      run: |
+        mkdir -p build
+        cd build
+        cmake ..
+
+    - name: Compile
+      working-directory: build
+      run: make -j2
+
+    - name: Print versions
+      working-directory: build
+      run: |
+        ./app/bergamot-translator-app --version
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,7 @@ include(CMakeDependentOption)
 
 # Project specific cmake options
 option(COMPILE_WASM "Compile for WASM" OFF)
-option(USE_WASM_COMPATIBLE_SOURCES "Use wasm compatible sources" ON)
+option(USE_WASM_COMPATIBLE_SOURCE "Use wasm compatible sources" ON)
 option(COMPILE_TESTS "Compile bergamot-tests" OFF)
 
 SET(PACKAGE_DIR "" CACHE STRING "Directory including all the files to be packaged (pre-loaded) in wasm builds")
@@ -22,11 +22,9 @@ SET(PACKAGE_DIR "" CACHE STRING "Directory including all the files to be package
 SET(COMPILE_CUDA OFF CACHE BOOL "Compile GPU version")
 SET(USE_SENTENCEPIECE ON CACHE BOOL "Download and compile SentencePiece")
 SET(USE_STATIC_LIBS ON CACHE BOOL "Link statically against non-system libs")
-if (USE_WASM_COMPATIBLE_SOURCES)
-  # Setting the marian submodule specific cmake options for wasm
+if (USE_WASM_COMPATIBLE_SOURCE)
   SET(COMPILE_LIBRARY_ONLY ON CACHE BOOL "Build only the Marian library and exclude all executables.")
   SET(USE_MKL OFF CACHE BOOL "Compile with MKL support")
-  SET(USE_WASM_COMPATIBLE_MARIAN ON CACHE BOOL "Enable those marian sources that compile to wasm. Useful for debugging wasm failures by building same sources natively")
   # # Setting the ssplit-cpp submodule specific cmake options for wasm
   SET(USE_INTERNAL_PCRE2 ON CACHE BOOL "Use internal PCRE2 instead of system PCRE2")
 endif()
@@ -54,6 +52,7 @@ if(NOT COMPILE_WASM)
 endif()
 
 if(COMPILE_WASM)
+  set(WORMHOLE ON CACHE BOOL "Use WASM wormhole in intgemm https://bugzilla.mozilla.org/show_bug.cgi?id=1672160")
   list(APPEND WASM_COMPILE_FLAGS -pthread -O3 -g2 -fPIC -mssse3 -msimd128)
   list(APPEND WASM_COMPILE_FLAGS "SHELL:-s WASM=1" "SHELL:-s ASSERTIONS=0" "SHELL:-s DISABLE_EXCEPTION_CATCHING=1" "SHELL:-s LLD_REPORT_UNDEFINED" "SHELL:-s FORCE_FILESYSTEM=1" "SHELL:-s ALLOW_MEMORY_GROWTH=1")
   list(APPEND WASM_COMPILE_FLAGS -Wno-error=pthreads-mem-growth)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,6 +26,7 @@ if (USE_WASM_COMPATIBLE_SOURCES)
   # Setting the marian submodule specific cmake options for wasm
   SET(COMPILE_LIBRARY_ONLY ON CACHE BOOL "Build only the Marian library and exclude all executables.")
   SET(USE_MKL OFF CACHE BOOL "Compile with MKL support")
+  SET(USE_WASM_COMPATIBLE_MARIAN ON CACHE BOOL "Enable those marian sources that compile to wasm. Useful for debugging wasm failures by building same sources natively")
   # # Setting the ssplit-cpp submodule specific cmake options for wasm
   SET(USE_INTERNAL_PCRE2 ON CACHE BOOL "Use internal PCRE2 instead of system PCRE2")
 endif()
--- a/src/translator/CMakeLists.txt
+++ b/src/translator/CMakeLists.txt
@@ -13,10 +13,15 @@ add_library(bergamot-translator STATIC
     sentence_ranges.cpp
     service.cpp
 )
+if (USE_WASM_COMPATIBLE_MARIAN)
+  # Using wasm compatible marian should include this compile definition;
+  # Has to be done here because of the marian's cmake practices
+  target_compile_definitions(bergamot-translator PUBLIC WASM)
+endif()
 
 if(COMPILE_WASM)
   # A dirty hack because of marian's bad cmake practices
-  target_compile_definitions(bergamot-translator PUBLIC USE_SSE2 WASM)
+  target_compile_definitions(bergamot-translator PUBLIC USE_SSE2)
   # Enable code that is required for generating JS bindings
   target_compile_definitions(bergamot-translator PRIVATE WASM_BINDINGS)
   target_compile_options(bergamot-translator PRIVATE ${WASM_COMPILE_FLAGS})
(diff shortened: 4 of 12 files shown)


src/translator/CMakeLists.txt
```diff
@@ -13,10 +13,15 @@ add_library(bergamot-translator STATIC
...
     service.cpp
 )
+if (USE_WASM_COMPATIBLE_MARIAN)
+  # Using wasm compatible marian should include this compile definition;
+  # Has to be done here because of the marian's cmake practices
+  target_compile_definitions(bergamot-translator PUBLIC WASM)
+endif()
 
 if(COMPILE_WASM)
   # A dirty hack because of marian's bad cmake practices
```
jerinphilip, line 23: At this point, I'm inclined to say all the paths generated by CMake courtesy of this project gives more headhurt than marian's CMakeLists. Drop the native marian act here (why do you need stuff outside WASM with WASM_COMPATIBLE sources here anyway?), now that you've added workflow files in browsermt/marian-dev. There's only WASM or full marian.

jerinphilip (changes requested):
Please come back with only the submodule update. Only WASM/non-WASM(full) switches here, builds are being tested in browsermt/marian-dev anyway with capabilities.

src/translator/CMakeLists.txt
kpu: I suppose the useful part of this is being able to natively test a "reduced" Marian.  Though I'm not convinced it's really a reflective test given the changes on top to disable threads.  And I still don't understand why Mozilla keeps disabling threads when the plan is to use them.

jerinphilip: * What I want is some sanity in the build system configuration to be able to edit without much struggle. 
Therefore I want **only** two paths here - one which compiles to target wasm and one which compiles to full marian. Why are these being set in translator directory's CMakeLists, are these supposed to be set project-wide (in the root CMakeLists). I'm talking about the compile flags being added for WASM. I particularly have a problem with this after the exercise of `WITHOUT_PTHREADS` -> `__EMSCRIPTEN__` -> `WASM`. The build system ninja techniques are creating more work for everyone (at review, at edits).
* The two additional (github) workflows are unnecessary - as WASM builds are necessary and sufficient enough. It can't be magically such that WASM works and native breaks with WASM_COMPATIBLE sources. The bergamot-translator sources only have the two choices of WASM and not WASM, no WASM_COMPATIBLE_SOURCES mode specially.

abhi-agg: > Drop the native marian act here (why do you need stuff outside WASM with WASM_COMPATIBLE sources here anyway?)

I am sure that you know that compiling to wasm platform for inference requires a "reduced" marian. Having capability to compile "reduced" marian natively is helpful in debugging the wasm runtime failures by first reproducing the same failures on native platform with the "reduced" Marian binaries. In the past, this has proven to be useful where we could quickly diagnose a few wasm specific issues (e.g. `size_t` inconsistency).

As far as workflows are concerned, they are additional checks and are not hurting any one. Having these checks in Marian is not good enough because bergamot-translator is the final wasm artifact that is used in extension. We plan to keep these workflows. Your `full_marian` workflows are untouched. So it should not bother you anyway.

abhi-agg:
> Please come back with only the submodule update. Only WASM/non-WASM(full) switches here, builds are being tested in browsermt/marian-dev anyway with capabilities.

As I said in my comments, these workflows are not hurting anyone. We intend to keep them for the reasons I explained in the comments. If 2 extra workflows are your only concern then please feel free to approve. We can merge it then 👍

jerinphilip (approved):
Approving in the interest of speeding up other work. 

The extra github workflows far as I'm concerned are abandoned (if an additional flag I add breaks these workflows, it's on the owners to fix - since I'm not supposed to be bothered).

abhi-agg:
> (if an additional flag I add breaks these workflows, it's on the owners to fix - since I'm not supposed to be bothered).

@jerinphilip Any workflow added to any repository doesn't work this way. The submitter of a PR is responsible to keep all CI checks successful.

These workflows are added for a reason (I already explained). If the wasm compatible sources don't compile natively, they will not compile on wasm anyway => we can't integrate it to extension. I hope this helps 👍

src/translator/CMakeLists.txt
```diff
@@ -13,10 +13,15 @@ add_library(bergamot-translator STATIC
...
-  target_compile_definitions(bergamot-translator PUBLIC USE_SSE2 WASM)
```
jerinphilip, line 19: Reset to this version. Remove the additional if block.

src/translator/CMakeLists.txt
```diff
@@ -13,10 +13,15 @@ add_library(bergamot-translator STATIC
     sentence_ranges.cpp
     service.cpp
 )
+if (USE_WASM_COMPATIBLE_MARIAN)
+  # Using wasm compatible marian should include this compile definition;
+  # Has to be done here because of the marian's cmake practices
+  target_compile_definitions(bergamot-translator PUBLIC WASM)
+endif()
```
jerinphilip, line 20: Remove this block.

CMakeLists.txt
```diff
@@ -24,6 +24,7 @@ if (USE_WASM_COMPATIBLE_SOURCES)
   # Setting the marian submodule specific cmake options for wasm
   SET(COMPILE_LIBRARY_ONLY ON CACHE BOOL "Build only the Marian library and exclude all executables.")
   SET(USE_MKL OFF CACHE BOOL "Compile with MKL support")
+  SET(USE_WASM_COMPATIBLE_MARIAN ON CACHE BOOL "Enable those marian sources that compile to wasm. Useful for debugging wasm failures by building same sources natively")
```
jerinphilip, line 27: Only two paths - (1) marian-dev full and (2) adapted marian targeting wasm. 

Successes at builds in (2) and additional tests for (2) are sufficient to cover the use-case here.

.github/workflows/native-custom_marian-ubuntu.yml
```diff
@@ -0,0 +1,33 @@
+name: Native (Custom Marian) Ubuntu
```
jerinphilip, line 1: Remove file.

.github/workflows/native-custom_marian-mac.yml
```diff
@@ -0,0 +1,33 @@
+name: Native (Custom Marian) MacOS
```
jerinphilip, line 1: Remove file. This github check now asks all contributors to both with editing CMakeLists which has unnecessary paths. Keep a simpler mental model. Read Section 6.2 [here](https://www.cs.cmu.edu/~ckaestne/pdf/ecoop15.pdf).

jerinphilip (changes requested):
Please resolve inline comments added, making the build system simpler and targeting only two paths. Since I'm told I'll have the responsibility to maintain these files if they're in, I'm updating review to requesting changes.

I foresee that allowing such a configuration allows ignorance of existing test-apps like https://github.com/browsermt/bergamot-translator/pull/76, and is quite unnecessary to support replicating already existing tests at service layer additionally in unified API.

CMakeLists.txt
jerinphilip: > Having capability to compile "reduced" marian natively is helpful in debugging the wasm runtime failures by first reproducing the same failures on native platform with the "reduced" Marian binaries. In the past, this has proven to be useful where we could quickly diagnose a few wasm specific issues (e.g. size_t inconsistency).

Bind these to WASM tests at WASM end (if tests in (2) fail and works at (1), we inspect the boundary).

andrenatal (approved):

abhi-agg:
> Since I'm told I'll have the responsibility to maintain these files if they're in, I'm updating review to requesting changes.

@jerinphilip Probably you misunderstood what I meant. Nobody told you that you are responsible to maintain the workflow files. The owner of each PR is responsible for keeping these CI checks passing to be able to merge stuff to main branch of this repository.

@kpu @andrenatal Jerin and I are having disagreement on 1 single topic in this PR and it is causing a lot of back and forth communication which is wasting time for me at least (I believe Jerin as well?). A short summary of disagreement:

I am proposing to keep 3 types of workflows in the CI:

1. compile bergamot-translator with **full-marian** on **native** platform
2. compile bergamot-translator with **reduced-marian** on **wasm** platform
3. compile bergamot-translator with **reduced-marian** on **native** platform

and @jerinphilip wants to keep only first 2 workflows.

Having capability to compile "reduced-marian" on native platform is helpful in debugging the wasm runtime failures by first reproducing the same failures on native platform with "reduced-marian" native binaries. In the past, this has proven to be useful where we could quickly diagnose a few wasm specific issues (e.g. inconsistent size of `size_t` data type on wasm and native platform). Therefore, I am proposing strongly to keep 3rd workflow.

Please make a decision and I will make the change.

jerinphilip:
> Probably you misunderstood what I meant 

Beg to differ.

> In the past, this has proven to be useful where we could quickly diagnose a few wasm specific issue

This is a special case, you don't need to enforce this path on all contributors (through a github check) and have the additional maintenance overhead on *everyone*. Mozilla mentioned being fully committed to WASM, ideally how this should happen therefore is "tests" at `WASM` (after `-DUSE_WASM_COMPATIBLE_SOURCES=on`; Shouldn't the rogue `size_t` be undetectable during builds anyway?) The github workflows are an absolute no-no.

In general adding more files which impedes development for everyone (also looking at Unified API) needs to stop  as well. This repository somehow seems to be operating with the default of:

* `COMPILE_WASM=off` and `-DUSE_WASM_COMPATIBLE_SOURCES=on`.  (This is a joke, right?) Probably this is why #76 ended up unaware of 6 existing apps? 

This path needs to go away as a whole, saner default set and not add more files to add weight for everyone to carry. Thus two paths -  `COMPILE_WASM=off` and `COMPILE_WASM=on`. Mozilla maintains one actively, we maintain the other. And we respect both configurations build (through github checks, and if possible tests) for harmonious development on both sides (shared ownership and maintenance duties). Anything more is downright unnecessary. Look at the if-else and additional maintenance ladder which follows allowing such a configuration [here](https://github.com/browsermt/bergamot-translator-tests/pull/7). 

> I believe Jerin as well?

Nope, this will save time in the [long-run](https://xkcd.com/974/).

abhi-agg:
> COMPILE_WASM=off and -DUSE_WASM_COMPATIBLE_SOURCES=on. (This is a joke, right?) 

Fyi, `COMPILE_WASM` cmake option compiles code to wasm platform (it involves setting emscripten specific compile options etc.) while `USE_WASM_COMPATIBLE_SOURCES` is a cmake option that enables only those sources (e.g. "reduced marian" + single threading in bergamot-translator) that compile to wasm. Both of these cmake options perform different functions and might be set independent of each other 👍

I can set `USE_WASM_COMPATIBLE_SOURCES` to `off` as well (if that is your only problem).

@kpu @andrenatal Requesting your intervention here 👍

andrenatal:
I'm struggling to understand what's the problem here that's hindering this of being merged. Based on the conversation I had with @abhi-agg today this patch will help us to keep the integrity of our workflow and to identify potential issues in the Wasm port. Is there anything else? Just a brief summary of what's wrong and where is the conflict would help.

.github/workflows/native-custom_marian-ubuntu.yml
kpu: We want to check the reduced Marian is working natively.  Otherwise we'll be in bigger bug territory later.

.github/workflows/native-custom_marian-mac.yml
kpu: We want to see the reduced version works natively.  Resolving.

src/translator/CMakeLists.txt
kpu: The combination of `USE_WASM_COMPATIBLE_MARIAN` (which I think is meant to be a native build that's reduced?) and then defining `WASM` doesn't make sense to me.  To me the define `WASM` means the compilation environment is web assembly (since you've asked to deprecate `__EMSCRIPTEN__`) and it would be fair game to use things that /only/ work in WebAssembly, such as the wormhole.  

If you want to have an in-between Marian that has less baggage and compiles natively, fine, but then be careful about what macro does what.

src/translator/CMakeLists.txt
```diff
@@ -13,10 +13,15 @@ add_library(bergamot-translator STATIC
...
 
 if(COMPILE_WASM)
   # A dirty hack because of marian's bad cmake practices
-  target_compile_definitions(bergamot-translator PUBLIC USE_SSE2 WASM)
+  target_compile_definitions(bergamot-translator PUBLIC USE_SSE2)
```
kpu, line 24: I'm confused, what happens if one doesn't set -DUSE_WASM_COMPATIBLE_MARIAN and does set -DCOMPILE_WASM?

kpu (changes requested):
Previously there were two compilation settings:
1. WebAssembly including removing some Marian code that isn't used by Mozilla
2. Native Marian with all features
This appears to be trying to make a total of three configurations:
1. WebAssembly including removing some Marian code that isn't used by Mozilla
2. Native including removing some Marian code that isn't used by Mozilla
3. Native Marian with all features

I accept the debugging value of scenario 2 and believe it is a good idea.  

However, I do not believe the `WASM` define is the appropriate way to accomplish 2.  The `WASM` define means it is compiling for WebAssembly and allowed to use stuff that only works in WebAssembly i.e. with wormhole et al.  If you want to have scenario 2, then there need to be at least two preprocessor defines with the correct semantics.

kpu:
I largely agree with you @abhi-agg with apology for two pedantic points:
1. The wormhole is even more special than just WASM: it requires a particular flavor of WebAssembly and you'll e.g. not want that for a Chrome build.  
2. Hopefully some submodules won't need touching at all!

abhi-agg:
>     1. The wormhole is even more special than just WASM: it requires a particular flavor of WebAssembly and you'll e.g. not want that for a Chrome build.

Agreed. But is there a way that we can solve it in an elegant way (just now) based on which browser uses it? May be we can go ahead with the proposal for now and solve this problem a bit later in a better way?

>     2. Hopefully some submodules won't need touching at all!
👍

jerinphilip (approved):

abhi-agg:
@kpu I made all the changes as we agreed. Could you please review and approve?

kpu (approved):

Issue #66, opened 2021-03-23
Make marian-decoder-new consume Response instead of Histories


jerinphilip:
#53, #65 requires this as a prerequisite.

Not having `histories_` will break functionality at replacement-decoder for benchmarks and running speed tests. Adjust the marian-decoder replacements to consume processed from history data (translation + annotation). Remove `OutputCollector` dependency. The process can also get rid of the lineNumber information when streamed from a large corpus, which we do not need in a request response mode of operation.

jerinphilip:
@eu9ene Moving discussion to relevant issue.

> My observations:
> 
> wmt20 en, number of lines 1418
> Line 224:
> ```
> As inflation climbs and the cost of living in Oregon soars, these workers are being left to flounder, and management has shown little sympathy for their plight. Contract negotiations that began in February broke down in mid-August. At that point, the bosses had laid out some particularly egregious proposals - a paltry cost-of-living increase, a denial of seniority demands from long-term university workers, and a price hike for staff meals - that workers saw as disrespectful, and even cruel. "There is an unwritten class system at the universities," Fulmer said. "And we are at the bottom of it."
> ```
> 
> My app, number of lines in `de` output 1418 ( `--ssplit-mode` option is not specified)
> Line 224:
> ```
> Während die Inflation steigt und die Lebenshaltungskosten in Oregon in die Höhe klettern, werden diese Arbeiter ins Wunkeln gelassen, und das Management hat wenig Verständnis für ihre Notlage gezeigt. Die im Februar begonnenen Vertragsverhandlungen scheiterten Mitte August. Zu diesem Zeitpunkt hatten die Chefs einige besonders ungeheuerliche Vorschläge vorgelegt - eine dürftige Erhöhung der Lebenshaltungskosten, eine Leugnung der Dienstaltersanforderungen von langjährigen Universitätsarbeitern und eine Preiserhöhung für die Mahlzeiten des Personals - die Arbeiter als respektlos und sogar grausam ansahen. "Es gibt ein ungeschriebenes Klassensystem an den Universitäten", sagte Fulmer. "Und wir sind am unteren Ende".
> ```
> 
> 
> `marian-decode-new`, number of lines in output with `--ssplit-mode sentence` 1428
> Line 224 and 225 (unexpected new line break)
> ```
[...]

PR #76, opened 2021-03-27
Change bergamot app to process stdin texts

eu9ene:
I need a simple app for evaluation that should be able to accept newline-separated texts in stdin, translate them separately and output results to stdout. Basically the same behavior as `marian-decoder`. The motivation to use `bergamot-translator` and high-level API is to be as close to the production setup in Firefox as possible.

@abhi-agg suggested fixing `bergamot-translator-app` to support this functionality.

### DIFF
--- a/app/bergamot-translator-app.cpp
+++ b/app/bergamot-translator-app.cpp
@@ -1,11 +1,13 @@
 /*
  * main.cpp
  *
- * An example application to demonstrate the use of Bergamot translator.
+ * An application which accepts line separated texts in stdin and returns translated ones in stdout.
+ * It is convenient for batch processing and can be used with tools like SacreBLEU.
  *
  */
 
 #include <iostream>
+#include <string>
 
 #include "TranslationModel.h"
 #include "translator/parser.h"
@@ -23,22 +25,10 @@ int main(int argc, char **argv) {
 
   TranslationRequest translationRequest;
   std::vector<std::string> texts;
-  texts.emplace_back(
-      "The Bergamot project will add and improve client-side machine "
-      "translation in a web browser.  Unlike current cloud-based "
-      "options, running directly on users’ machines empowers citizens to "
-      "preserve their privacy and increases the uptake of language "
-      "technologies in Europe in various sectors that require "
-      "confidentiality.");
-  texts.emplace_back(
-      "Free software integrated with an open-source web "
-      "browser, such as Mozilla Firefox, will enable bottom-up adoption "
-      "by non-experts, resulting in cost savings for private and public "
-      "sector users who would otherwise procure translation or operate "
-      "monolingually.  Bergamot is a consortium coordinated by the "
-      "University of Edinburgh with partners Charles University in "
-      "Prague, the University of Sheffield, University of Tartu, and "
-      "Mozilla.");
+
+  for (std::string line; std::getline(std::cin, line);) {
+        texts.emplace_back(line);
+  }
 
   auto results = model->translate(std::move(texts), translationRequest);
 
@@ -46,17 +36,7 @@ int main(int argc, char **argv) {
   //std::vector<TranslationResult> results = futureResults.get();
 
   for (auto &result : results) {
-    std::cout << "[original]: " << result.getOriginalText() << std::endl;
-    std::cout << "[translated]: " << result.getTranslatedText() << std::endl;
-    auto mappings = result.getSentenceMappings();
-    for (auto &p : mappings) {
-      std::string_view src = p.first;
-      std::string_view tgt = p.second;
-
-      std::cout << " [src Sentence]: " << src << std::endl;
-      std::cout << " [tgt Sentence]: " << tgt << std::endl;
-    }
-    std::cout << std::endl;
+    std::cout << result.getTranslatedText() << std::endl;
   }
 
   return 0;


jerinphilip:
> Basically the same behavior as marian-decoder.

Probably not a good idea. marian-decoder has lesser features for Mozilla purposes than the API provided here (sentence-splitting won't be tested, same applies for alignments, quality-scores). 

How it's organized to cover paths here is:
1. `marian-decoder-new` is a replacement marian-decoder, and the check you want. In a regression-test, it runs on WNGT20 and computes BLEU and also checks we don't have hits on speed with any incoming changes.
2. `service-cli` tests `Service` (with alignments, sentence-splitting, 1 doesn't have this).
3. `bergamot-translator-app` tests integration through regression-tests. If tests breaks here and does not break at 2 is when we inspect the high-level API (which I assume is Unified API when you mention this). We need a "blob of text" here, not separated lines.

You can find the above already [here](https://github.com/browsermt/bergamot-translator/blob/main/doc/marian-integration.md).

I recommend you use existing `marian-decoder-new`. [Here](https://github.com/browsermt/bergamot-translator-tests/blob/main/speed-tests/test_wngt20_perf.sh#L64-L75) is an example of how to use it. Use it if you just need an app to report BLEU scores for some production table.  The changes in this PR will break regression-tests (bound to bergamot-translator-app) [here](https://github.com/browsermt/bergamot-translator-tests/blob/65071f09a309dcae5931b607db2138745803775d/tests/basic/test_bergamot_translator_app_intgemm_8bit.cpu-threads.0.sh#L46).

kpu:
What is the exact purpose here?  

Demonstrate how WASM, 1 thread, and not properly batching have decreased speed by over a factor of 10?  

Evaluate quality?

abhi-agg:
Few things regarding this PR:

1. It removes the [hardcoded input for translation](https://github.com/browsermt/bergamot-translator/blob/main/app/bergamot-translator-app.cpp#L26-L41) from bergamot-translator-app and starts receiving this input from stdin. This should happen irrespective of where the evaluation is happening.

2. Doing this change actually makes [bergamot-translator-app](https://github.com/browsermt/bergamot-translator/blob/main/app/bergamot-translator-app.cpp) and [service-cli](https://github.com/browsermt/bergamot-translator/blob/main/app/service-cli.cpp) consistent (because service-cli doesn't have any hardcoded inputs for translation and also accepts input from stdin).

3. For the evaluation we need to work with the TranslationModel class (and not with Service class) and for that [bergamot-translator-app](https://github.com/browsermt/bergamot-translator/blob/main/app/bergamot-translator-app.cpp#L26-L41) (and [bergamot-translator-app-bytearray](https://github.com/browsermt/bergamot-translator/blob/main/app/bergamot-translator-app-bytearray.cpp)) is the right place for this change.

4. It makes bergamot-translator-tests repo also consistent for different tests across different apps. (How? See below.)

> The changes in this PR will break regression-tests (bound to bergamot-translator-app) here.

Agreed. And we can fix them. We already have a [bergamot.in](https://github.com/browsermt/bergamot-translator-tests/blob/65071f09a309dcae5931b607db2138745803775d/data/simple/bergamot.in) test input that we are using in [service-cli test](https://github.com/browsermt/bergamot-translator-tests/blob/65071f09a309dcae5931b607db2138745803775d/tests/basic/test_service-cli_intgemm_8bit.cpu-threads.4.sh#L46). We can use the same input for testing bergamot-translator-app by simply changing [this line](https://github.com/browsermt/bergamot-translator-tests/blob/65071f09a309dcae5931b607db2138745803775d/tests/basic/test_bergamot_translator_app_intgemm_8bit.cpu-threads.0.sh#L46) with [this](https://github.com/browsermt/bergamot-translator-tests/blob/65071f09a309dcae5931b607db2138745803775d/tests/basic/test_service-cli_intgemm_8bit.cpu-threads.4.sh#L46). Infact, this change can make the test scripts [test_bergamot_translator_app*](https://github.com/browsermt/bergamot-translator-tests/tree/65071f09a309dcae5931b607db2138745803775d/tests/basic) consistent with the rest of the test files there (every test file will accept input from bergamot.in) and simplifies the bergamot-translator-tests repo. It is a simple change and @eu9ene or I can do it 👍

jerinphilip:
@abhi-agg: The following is what you missed.

> marian-decoder has lesser features for Mozilla purposes than the API provided here (sentence-splitting won't be tested, same applies for alignments, quality-scores).
> We need a "blob of text" here, not separated lines.

The PR puts forth a Quality Evaluation app, different use-case than what can comprehensively cover above.

abhi-agg:
> @abhi-agg: The following is what you missed.
> 
> > marian-decoder has lesser features for Mozilla purposes than the API provided here (sentence-splitting won't be tested, same applies for alignments, quality-scores).

Wait. I think when @eu9ene wrote "Basically the same behavior as marian-decoder. " in his first comment, he only meant passing the input text from stdin to bergamot-translator-app and nothing more. (@eu9ene please correct me if I misinterpreted)

> > We need a "blob of text" here, not separated lines.

I don't think @eu9ene is proposing to pass separated lines. He is passing `newline-separated texts in stdin` where **texts**, of course, be a "blob of text"

> The PR puts forth a Quality Evaluation app, different use-case than what can comprehensively cover above.

The bergamot-translator-app can be augmented to evaluate all those things that `marian-decoder-new` app is evaluating currently. Right?

All I am saying is that for us, the evaluation would be more correct if we work on the interface (i.e. TranslationModel) rather than on the Service class. (I have started working on unified API so that both Service and TranslationModel class have parity.)

@eu9ene Please correct me if I am completely off here 👍

eu9ene:
@abhi-agg yes, you are correct in all points. 

The purpose of this PR is to provide a convenient interface for final quality evaluation. It should be as close to the production setup as possible and should use the same unified API which is used in the browser extension. I pass the same arguments too. The goal of my evaluation is to get a build that is very similar to the final product, get BLEU scores and compare them to other translation products.

I am using a pipeline with SacreBLEU, similar to this one https://github.com/browsermt/students/blob/master/esen/enes.student.tiny11/eval.sh, that's why I need an app to accept new-line separated texts.

If changing this app is breaking something I can just create another one.

jerinphilip:
> The goal of my evaluation is to get a build that is very similar to the final product, get BLEU scores and compare them to other translation products.

Use `marian-decoder-new`. You get same quality from both. There's no difference. Close to higher level API doesn't change quality, it's a function of the underlying engine not the layers of code in between (unless there are bugs, which is the problem for an integration test-app, not a separate evaluation app).

eu9ene:
@jerinphilip `marian-decoder-new` doesn't work for me because:

1. It does not split output sentences properly, so the output file lines number is different from input one;
2. I still prefer using unified API as a proper abstraction, exactly the same which will be used in production even if it is clear from the code that the result should be the same. You can consider it black-box testing and it is just more future proof.

If my proposal doesn't align with the vision for apps here, I can just add an extra app to the Mozilla fork. It's what I did initially, but @abhi-agg suggested to move it here.

kpu:
> I am using a pipeline with SacreBLEU, similar to this one https://github.com/browsermt/students/blob/master/esen/enes.student.tiny11/eval.sh, that's why I need an app to accept new-line separated texts.

> I still prefer using unified API as a proper abstraction

BLEU expects a sentence in and a sentence out.  Then it scores corresponding sentences in a corpus.  No sentence splitting.  The "unified API" always runs sentence splitting.  

I guess you could treat each sentence as a blob and run it through the splitter, translate separately, and rejoin to get an end-to-end feel?  

> I still prefer using unified API as a proper abstraction

This is self-contradictory.  It was a draft of an API that I didn't ask for and we all agreed would change with implementation.

kpu:
> I have started working on unified API so that both Service and TranslationModel class have parity.

They should be the same class.  One was the draft interface and one was the actual implementation.

eu9ene:
My observations:

wmt20 en, number of lines 1418
Line 224:
```
As inflation climbs and the cost of living in Oregon soars, these workers are being left to flounder, and management has shown little sympathy for their plight. Contract negotiations that began in February broke down in mid-August. At that point, the bosses had laid out some particularly egregious proposals - a paltry cost-of-living increase, a denial of seniority demands from long-term university workers, and a price hike for staff meals - that workers saw as disrespectful, and even cruel. "There is an unwritten class system at the universities," Fulmer said. "And we are at the bottom of it."
```

My app, number of lines in `de` output 1418 ( `--ssplit-mode` option is not specified)
Line 224:
```
Während die Inflation steigt und die Lebenshaltungskosten in Oregon in die Höhe klettern, werden diese Arbeiter ins Wunkeln gelassen, und das Management hat wenig Verständnis für ihre Notlage gezeigt. Die im Februar begonnenen Vertragsverhandlungen scheiterten Mitte August. Zu diesem Zeitpunkt hatten die Chefs einige besonders ungeheuerliche Vorschläge vorgelegt - eine dürftige Erhöhung der Lebenshaltungskosten, eine Leugnung der Dienstaltersanforderungen von langjährigen Universitätsarbeitern und eine Preiserhöhung für die Mahlzeiten des Personals - die Arbeiter als respektlos und sogar grausam ansahen. "Es gibt ein ungeschriebenes Klassensystem an den Universitäten", sagte Fulmer. "Und wir sind am unteren Ende".
```


`marian-decode-new`, number of lines in output with `--ssplit-mode sentence` 1428
Line 224 and 225 (unexpected new line break)
```
Während die Inflation steigt und die Lebenshaltungskosten in Oregon in die Höhe schießen, werden diese Arbeiter ins Wanken gelassen, und das Management hat wenig Verständnis für ihre Notlage gezeigt. Die Vertragsverhandlungen, die im Februar begannen, scheiterten Mitte August. Zu diesem Zeitpunkt hatten die Bosse einige besonders ungeheuerliche Vorschläge aufgestellt - eine dürftige Erhöhung der Lebenshaltungskosten, eine Verweigerung der Dienstaltersanforderungen von langjährigen Universitätsarbeitern und eine Preiserhöhung für Personalprämien - das Arbeiter als respektlos und sogar grausam an den Universitäten sahen.
"Bedermer sagte: "Und wir sind am unteren Ende".
```

`marian-decode-new`, number of lines in output with `--ssplit-mode paragraph` 2103


So, a number of output lines for `marian-decoder-new` does not match a number of input lines which makes it not usable for evaluation pipelines. Maybe I do something wrong?
Also translations are a bit different for my app and `marian-decoder-new` which is likely a side effect of incorrect splitting.

jerinphilip:
> marian-decode-new, number of lines in output with --ssplit-mode sentence 1428
Line 224 and 225 (unexpected new line break)

This is not expected to happen, could be a bug. Try `--max-length-break <big-value>` and see if problem persists. `--ssplit-mode sentence` is expected. If does not work for you let's move the discussion into an issue.

kpu:
> This is not expected to happen, could be a bug. Try `--max-length-break <big-value>` and see if problem persists. `--ssplit-mode sentence` is expected. If does not work for you let's move the discussion into an issue.

Sounds like something for @jerinphilip to investigate.

jerinphilip:
Moved to an issue (https://github.com/browsermt/bergamot-translator/issues/66) which I think is the problem. It's output collector directly tapping into histories as of now. If this is not the case, I request @eu9ene to provide the full args being used to `marian-decoder-new` so I can replicate the issue.

eu9ene:
@jerinphilip @kpu @abhi-agg  my requirements are to evaluate translation quality which FIrefox users will see. For this, I take exactly the same abstraction as used in the extension right now (TranslationModel), exactly the same parameters as used in the extension, use standard wmt datasets and this app does exactly what I need. It splits lines and translates paragraphs separately writing resulting paragraphs as the same number of output lines. I assume the same job would be done by the extension ( of course there is wasm and other factors like html tags which might make results different, but I'd like to not think about those for now).

I'm out of context regarding implementation details. It's hard for me to advise on decisions regarding underlying classes in the issue https://github.com/browsermt/bergamot-translator/issues/66.
I would like to keep it simple on my side and just use abstractions used in extension whatever they are at this point.

My evaluation is already done and the only thing left is this PR. Let's make a decision here. 
There are 3 options:
- approve this PR and merge (someone will have to deal with tests etc);
- copy this code to create a new app specifically for my needs;
- I'm closing this PR and create another app in Mozilla fork.

All three options work perfectly for me. cc @andrenatal

app/bergamot-translator-app.cpp
```diff
@@ -23,40 +25,18 @@ int main(int argc, char **argv) {
...
-      "browser, such as Mozilla Firefox, will enable bottom-up adoption "
-      "by non-experts, resulting in cost savings for private and public "
-      "sector users who would otherwise procure translation or operate "
-      "monolingually.  Bergamot is a consortium coordinated by the "
-      "University of Edinburgh with partners Charles University in "
-      "Prague, the University of Sheffield, University of Tartu, and "
-      "Mozilla.");
+
+  for (std::string line; std::getline(std::cin, line);) {
+        texts.emplace_back(line);
```
kpu, line 30: There should be a cap on the length of this if it's ever to be used for production but since this is a toy it's fine.

jerinphilip:
> (someone will have to deal with tests etc);

@eu9ene Seeing as to how the clean resolution from my side will take some time depending on when I get feedback, I can settle for where you create an app with a different name, like @abhi-agg suggested so it doesn't break tests. 

I hope we are on the same page this app and `getTranslatedText()` does not accomplish what you exactly want, it's just temporarily working for you (for languages with space character separating sentences).

andrenatal:
Hi @eu9ene I was going to say to close this PR and create another app in Mozilla fork if there was still resistance, but I see @kpu merged it, so all good from my end. Thanks!

andrenatal (approved):

PR #80, opened 2021-03-31
Cleanup API: Refactor request on-complete transition

jerinphilip:
(WIP Rework of the `Request -> Response` transition process)

Over here we're not in the small PR making business, we're in moving source from complete to more complete stages business, leaving intermediates stable.

To simplify diff - merge https://github.com/browsermt/bergamot-translator/pull/85 which provides essential bugfixes for what would previously be undefined behaviour, followed by https://github.com/browsermt/bergamot-translator/pull/79 which requires the bugfixes and eliminates the usage of history.

(Edit: This is untangled now)

Incorporating both the above, what this PR does is abstract Request -> Response transition using ResponseBuilder. 
- Request calls ResponseBuilder on completion with histories, and this triggers the construction of Response.  
- ResponseBuilder is also tasked with making certain objects optional like Alignment and QualityScores.
- To test ResponseBuilder, it has to be wired to Service and all the way to command-line apps, requiring edits in these places.
- To be wired to Service in a feature-complete setting, we require the notion of ResponseOptions (in ResponseBuilder), which is analogous to TranslationRequest. 
- So to move from complete sources to complete sources all 4 of the above has to be done.

There are the last few changes that can be moved to a separate collapse PR. behind this, will do by git-magic if required. The structures operated are documented as edits happen, so this PR also includes documentation changes.

We have big PR, yet again.


src/translator/response.cpp
```diff
@@ -6,101 +6,5 @@
...
-      const char *data = &target.text[begin_idx];
-      size_t size = end_idx - begin_idx;
-      targetMappings.emplace_back(data, size);
-    }
-
-    target.addSentence(targetMappings);
-  }
-}
-} // namespace bergamot
+namespace bergamot {} // namespace bergamot
```
abhi-agg, line 9: So this file can be deleted. Right?

src/translator/response.h
```diff
@@ -40,21 +40,22 @@ struct Quality {
 /// AnnotatedText provides an API to access markings of (sub)-word and
 /// sentences boundaries, which are required to interpret Quality and
 /// Alignment (s) at the moment.
-class Response {
+struct Response {
```
abhi-agg, line 43: Few quick questions:
Does it help in some way to convert this class to a struct?
Do you want to keep all the data members of this class public?

app/service-cli.cpp
```diff
@@ -21,8 +21,14 @@ int main(int argc, char *argv[]) {
   std::string input = std_input.str();
   using marian::bergamot::Response;
 
+  marian::Ptr<marian::Options> responseOptions = marian::New<marian::Options>();
```
abhi-agg, line 24: @jerinphilip Should the app `service-cli-bytearray` also be updated?

src/translator/response.h
jerinphilip: I have no strict notions here. Felt very natural to make this a struct and only carrying data/message to be responded with, and placing bindings on things like Annotation. 

If they're private, I'm probably looking at either setters or `ResponseBuilder` being a friend.

src/translator/service.h
```diff
@@ -52,20 +82,55 @@ class Service {
...
+  /// Also useful when one has to set/unset alignments or quality in the
+  /// Response to save compute spent in constructing these objects.
+
+  /// @param [in] source: rvalue reference of the string to be translated /
+  /// @param [in] translationRequest: TranslationRequest (Unified API)
+  /// indicating whether or not to include some member in the Response, also
+  /// specify any / additional configurable parameters.
+
+  std::vector<Response>
+  translateMultiple(std::vector<std::string> &&source,
```
abhi-agg, line 123: Does this API use `PCQueue` and multiple threads internally in the implementation ?

src/translator/response.cpp
jerinphilip: Yes, will do.

app/service-cli.cpp
jerinphilip: When I synchronize with main, yes. Expecting CI to complain when https://github.com/browsermt/bergamot-translator/pull/93 is merged.

src/translator/service.h
jerinphilip: Yes, and no. The mechanism to hide away `PCQueue` as a whole is `ifdefs`. So in one compile switch, yes, the other - no.

abhi-agg: Great. Thanks. So, once we collapse draft API (i.e. Translation* classes) and actual implementation (i.e. Service and Response) classes then we will call `translateMultiple` API for wasm (until we resolve wasm multithreading issue) 👍

src/translator/response.h
kpu: I like having friends.

app/service-cli.cpp
jerinphilip: So the printing code is pulled into the library to avoid duplication. This can be eventually extended maybe to output in a more readable format and to programmatically check cases (no alignments, yes alignments, no quality, quality). Both service-cli and service-cli-bytearray reuses this bit.

Stopping at a minimum invasive edit for now which eliminates duplication.

src/translator/response.cpp
jerinphilip: File is deleted.

src/translator/response.h
jerinphilip: Two things:

1, there's this API being used in a downstream app, where we will break API when members go private and accessors are needed.
2, This PR will become further bloated. Designing the proper API requires an iteration of experimentation ideally with WASM in my opinion, as several nuances like [this one](https://github.com/mozilla/bergamot-translator/issues/8#issuecomment-795929536) remain unanswered. I'm kind-of operating in the blind here without going into WASM which will take longer.

I'm not saying hard no-change here, just that it be deferred so certain other things can progress in parallel. Is this okay for everyone?

XapaJIaMnu: Currently the primary user of this is mozilla, so changing the interface  should be consolidated with @abhi-agg and @motin .
What is the opinion from mozilla side?

abhi-agg: As this PR was intended to bridge the gap b/w Translation* classes and corresponding internal classes (Service, Response), we can go ahead without touching on Quality Scores and Alignments part of this structure.

> Designing the proper API requires an iteration of experimentation ideally with WASM in my opinion

I second this considering the Quality score and Alignment will require some experimentation on JS side 👍

My initial comment was about using `struct` vs a `class` and following a standard way of using the Builder design pattern in the code. I am fine to ignore both of these comments and go ahead 👍 

@jerinphilip Please let us know when the PR is ready for the review again so that we can expedite the merge 👍

jerinphilip: It's ready now. 

I have reverted a few additional DRY attempts to avoid bloating to keep diff minimum. I will be working on them separately after this PR (this will collapse the service-cli, service-cli-bytearray apps into one with a cmdline option and probably make emitting of the `Response` as JSON or something more granular to independently test quality / alignments / raw text / sentences outputs). The only changes now from last review is removal of the now empty `response.cpp` file and fixing service-cli-bytearray to account for `ResponseOptions`/`TranslationRequest`. Both of which are complete. 

Let me know if anything further needs to be done.

kpu:
One concern about friends: can we mock out the system?  It can be useful for users to be able to make their own response object for testing purposes.

app/service-cli-bytearray.cpp
```diff
@@ -27,8 +27,14 @@ int main(int argc, char *argv[]) {
   std::string input = std_input.str();
   using marian::bergamot::Response;
 
+  marian::Ptr<marian::Options> responseOptions = marian::New<marian::Options>();
```
kpu, line 30: Why are we using some weird Options object for what should be a struct?

src/translator/sentence_ranges.h
```diff
@@ -125,9 +124,10 @@ struct AnnotatedText {
   /// constructor is disallowed).
   AnnotatedText(std::string &&text) : text(std::move(text)){};
 
-  AnnotatedText(AnnotatedText &&annotatedBlob)
-      : text(std::move(annotatedBlob.text)),
-        annotation(std::move(annotatedBlob.annotation)) {}
+  // @jerinphilip: We will try the default move constructor for now.
+  // AnnotatedText(AnnotatedText &&annotatedBlob)
+  //     : text(std::move(annotatedBlob.text)),
+  //       annotation(std::move(annotatedBlob.annotation)) {}
```
jerinphilip, line 130: Note to self, cleanup.

src/translator/response_options.h
```diff
@@ -0,0 +1,50 @@
...
+
+  /// Threshold between `[0.0f, 1.0f]` to filter alignments into a sparse
+  /// matrix. Higher value implies stronger filtering leading to provision of
+  /// higher-confidence matches. `1.0f` gives argmax (not the full-dense
+  /// matrix).
+  float alignmentThreshold{0.2f};
+
+  QualityScoreType qualityScoreType{QualityScoreType::FREE};
+  ConcatStrategy concatStrategy{ConcatStrategy::FAITHFUL};
+};
```
jerinphilip, line 45: @kpu Removed the `Options` in favour of a cpp struct, if something needs to stay/go among members or requires rename etc please let know. 

We're looking at collapse of `TranslationRequest` and `ResponseOptions`. I'm deferring collapse here to avoid editing bindings.

app/service-cli-bytearray.cpp
jerinphilip: This is converted into a struct. Resolving here.

jerinphilip:
Please treat this as an explicit request to expedite merge @abhi-agg, @XapaJIaMnu, @kpu.

1. What was previously on the `Options` framework (ResponseOptions) is now a struct, per @kpu's review. 
2. The changes @XapaJIaMnu wants for MSVC compile (https://github.com/browsermt/bergamot-translator/issues/105#issuecomment-826846522) are the last few commits. The sources are moved around between files, so make sense to apply those in the cleanup commit as well.

src/translator/response_options.h
```diff
@@ -0,0 +1,50 @@
...
+
+enum QualityScoreType {
+  /// Provide a free quality-score that comes with the machine-translation model
+  /// itself.
+  FREE,
+
+  /// An expensive quality-score that runs additional computations to determine
+  /// quality of an output.
+  EXPENSIVE
```
abhi-agg, line 23: Just adding that we probably would want to expose these scores at different text granularity as well. e.g. `Word `level, `Sentence` level. But we can update this enum later once we do more experimentation 👍

app/service-cli-bytearray.cpp
```diff
@@ -27,8 +27,14 @@ int main(int argc, char *argv[]) {
...
   using marian::bergamot::Response;
 
+  marian::bergamot::ResponseOptions responseOptions;
+  responseOptions.qualityScores = true;
+  responseOptions.alignment = true;
+  responseOptions.alignmentThreshold = 0.2f;
+
   // Wait on future until Response is complete
-  std::future<Response> responseFuture = service.translate(std::move(input));
+  std::future<Response> responseFuture =
```
XapaJIaMnu, line 36: I assume the old behaviour is preserved. If no `ResponseOptions` is provided you'd have a default behavior that emulates the old one?

XapaJIaMnu (approved):
Looks good

abhi-agg (approved):
looks good to me as well 👍

app/service-cli-bytearray.cpp
jerinphilip: No, defaults are set to off. General consensus among Mozilla and Kenneth is these things are off by default.

jerinphilip: Old behaviour is preserved though. I will collapse service-cli and bytearray (moving output outside) in a PR behind this.

XapaJIaMnu: @jerinphilip are we ready to merge now?

jerinphilip: Squashed and merged. Note that I have gotten rid of the move constructor troubles you are likely to have in windows, I don't have means to test them yet. But most of your patch described in https://github.com/browsermt/bergamot-translator/issues/105#issuecomment-826846522 should be in.

PR #81, opened 2021-04-01
Explictly install gcc version and use 8

kpu:
CI is failing due to some change at github https://github.com/browsermt/bergamot-translator/pull/69/checks?check_run_id=2245404337

### DIFF
--- a/.github/workflows/native-full_marian-ubuntu.yml
+++ b/.github/workflows/native-full_marian-ubuntu.yml
@@ -15,7 +15,7 @@ jobs:
           - name: "Ubuntu CPU-only"
             os: ubuntu-latest
             cuda: ""
-            gcc: 7
+            gcc: 8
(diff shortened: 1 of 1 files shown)


.github/workflows/native-full_marian-ubuntu.yml
```diff
@@ -2,9 +2,9 @@ name: Native (Full Marian) Ubuntu
 
 on:
   push:
-    branches: [ main, ci-sandbox ]
+    branches: [ main, ci-test ]
```
jerinphilip, line 5: My intentions of `ci-sandbox` is to be force-pushed for this sort of situations, when the only way is to debug is push to github runner without creating a mess of commit history and notification spam. I will make the update eventually in a separate PR and update somewhere.

jerinphilip (approved):

PR #85, opened 2021-04-01
Strengthen the Annotation class: Handle empty sentences and tests

jerinphilip:
There's a bug in the current code in main due to undefined behaviour of Annotation class upon inserting empty sentences (there are no word annotations). This can arise in scenarios where the target output ends up being empty due to vocabulary eating characters not present etc. 

image

This came to light while working to remove histories dependency replacing OutputCollector and using Annotation instead, where-in invalid inputs were encountered.

To ensure this works well ahead, the unit tests for Annotation are enhanced with descriptions and the container made to adhere to C++ convention  of half-interval [begin, end), when it comes to sentence-markers to simplify thinking.

The unit tests which worked before (on non-empty sentences are still intact) and more have been added to cover empty sentences, clearly marking in the constructed text what point the empty sentence came up.

This is critical, and is tested in a separate PR with WNGT20. Moving here as a "more focused change".


src/translator/sentence_ranges.cpp
```diff
@@ -6,48 +6,50 @@ namespace marian {
...
+  }
+  return sentenceByteRange;
 }
 
 ByteRange Annotation::word(size_t sentenceIdx, size_t wordIdx) const {
-  size_t offset = sentenceBeginIds_[sentenceIdx];
-  // auto terminals = sentenceTerminals(sentenceIdx);
-  // assert(offset + wordIdx <= terminals.second);
-  return flatByteRanges_[offset + wordIdx];
+  size_t bosOffset = (sentenceIdx == 0) ? 0 : sentenceEndIds_[sentenceIdx - 1];
```
kpu, line 51: What should be a simple fast accessor is turning into compute.  Normally the way to deal with this, for better or worse, is to put 0 at the beginning of sentenceEndIds.

src/translator/sentence_ranges.h
```diff
@@ -46,24 +46,13 @@ class Annotation {
...
-  /// end from the second.
-  std::pair<ByteRange, ByteRange> sentenceTerminals(size_t sentenceIdx) const;
-
-  /// Returns indices of terminal (word) ByteRanges in sentenceIds_ of a
-  /// sentence corresponding to sentenceIdx. The distance can be used to compute
-  /// number of words in a sentence (numWords) and also to construct the
-  /// terminal ByteRanges (sentenceTerminals).
-  std::pair<size_t, size_t> sentenceTerminalIds(size_t sentenceIdx) const;
+  /// Stores indices where sentences end (not inclusive, aligned with C++ half
+  /// interval notions)
```
kpu, line 54: Indices into what?

kpu:
It seems the natural design here is:
```
// Byte offset where each token begins in the string.
std::vector<size_t> tokens_;
// Index into tokens_ where each sentence begins.  
std::vector<size_t> sentenceTokens_;
// Sentence i begins at text[tokens_[sentenceTokens_[i]]] and ends at text[tokens_[sentenceTokens_[i+1]]
```
where both go one longer to allow indexing the end.  And if there's an empty sentence, so be it, they have consecutive integers that are equal in `sentenceTokens_`

Right now there's too many if statements trying to get around the design decisions.

jerinphilip:
@kpu The new design assumes we're not dropping characters in source (which I think we might end up doing, due to sentencepiece). If we can reach consensus on that happy to implement, I find the proposed design far more elegant. `ByteRange` directly maps to `string_views` provided through spiece.

kpu:
I think by "assumes we're not dropping characters in source" you mean there are bytes in the source input with an empty target.  How is that an issue with the code I proposed?  An empty target sentence just has the same beginning and end token index.

jerinphilip:
Look at the screenshot. The (source) due to difference in script doesn't have a ByteRange/string_view corresponding to it. Now this ByteRange is used as a meaningful unit in alignment or quality-estimation.

Consider code-mixed data fed to the input vocab in English (alpha, beta being greek symbols).
* AB\alpha\betaCD -> A B [omitted] C D

We can no longer work with the following as meaningful units for alignments etc.
```
// Byte offset where each token begins in the string.
std::vector<size_t> tokens_;
```

kpu:
I don't see the problem.  
Let's say the output is "a b c d" where "a b", empty, and "c d" are sentences.
tokens_ = {0, 2, 4, 6, 7};
sentenceTokens_ = {0, 2, 2, 4};

jerinphilip:
> Consider code-mixed data fed to the input vocab in English (alpha, beta being greek symbols). A, B, C, D are subwords.
> * AB\alpha\betaCD -> A B [omitted] C D (spiece output)

Without both terminals stored (which will happen when ByteRange -> size_t), I end up with the following unit in alignment/quality:
> * A B\alpha\beta C D

kpu:
If there's input that SentencePiece doesn't have characters for, it will map those to the unknown word.  That the system elected to provide no output when given Greek characters as input does not indicate a problem with SentencePiece and token handling.

jerinphilip:
> it will map those to the unknown word

Yes, but doesn't give me an equivalent `ByteRange` (from what I understand). So the math (with just `size_t`, and not `ByteRange`) works out by grabbing the next words begin. Remember, we're operating on the unnormalized, unoperated text for `string_views` to get subword annotations.

With `size_t`, I get: `[A] [B\alpha\beta] [C] [D]` as subwords. With ByteRange I get `[A] [B] \alpha\beta [C] [D]`, where what's outside square paranthesis is omitted. This also works out well if we have to say faithfully translate emoji or invalid characters.

kpu:
It was your job to map SentencePiece tokens back to byte ranges in the source.  If it doesn't map `<unk>` back to the right span, that's a big to fix there.  And I've seen no evidence that there is such a bug.

kpu (changes requested):
Overcomplicated, should be done with simple token and sentence offsets.

jerinphilip:
So, I tried. Unable to hold the following desideratum (without an if mess). In case there are newlines, the following is a deal-breaker of sorts (a sentence gets the `\n` as it's looking at beginning of next sentence):

```cpp
// Index into tokens_ where each sentence begins.  
// Sentence i begins at text[tokens_[sentenceTokens_[i]]] and ends at text[tokens_[sentenceTokens_[i+1]]]
```
Should I strip `\n` at output and fix unit tests and document API thus?

> where both go one longer to allow indexing the end.

We should probably not violate a constraint that a sentence begins at the start of first-word and ends at end of last word (not at the beginning of next sentence; this means we add a per sentence EoS marker at the minimum, not just one longer for the whole container). 


```cpp
// Sentence i begins at text[tokens_[sentenceTokens_[i]]] and ends at text[tokens_[sentenceTokens_[i+1] - 1]]
```

I  also tried the above with a per-sentence EOS marker, but that makes an `if` mess which we are trying to get rid of. Is there any natural way to deal with this? 

I can also explicitly store the `sentenceEndIds` and `sentenceBeginIds` (but due to the `\n` issue, we also need end of words).

src/translator/sentence_ranges.cpp
jerinphilip: This change is incorporated and improves source. https://github.com/browsermt/bergamot-translator/pull/85/commits/505c4d2c43c9f7544eeb714537ec093d1b670ee2

jerinphilip:
^ The above is just the first of the failure among unit tests I added. However, I realize having only beginning of tokens is not equivalent to having word string_views (`[begin, end)`). 

```cpp
// Byte offset where each token begins in the string. 
std::vector<size_t> tokens_;
// ^ (this is just beginning, I need end as well to avoid spaces).
```

Coming back:
> The new design assumes we're not dropping characters in source

We are dropping characters, at least spaces (among words) and newlines (between sentences). The special sentencepiece unicode character which allows space math does not exist pre-sentencepiece-processing (the source text we're using) and post decoding and hence no `ByteRange` for that.

SentencePiece was probably my overthinking and imagination, my bad there, apologies for throwing a tangent which was a waste of time. 

I have already hammered what would be a `vector<vector<ByteRange>>` into a flat structure, which was reasonable to do due to realloc concerns. However, compressing one more level further (`ByteRange` -> `size_t`) is a maintenance headache (as we loose the `string_view -> ByteRange` one to one mapping niceness) and in all likelihood, wrong. At this stage, we're over-optimizing for space and probably some double indexing elegance.  

As of now the following assumptions hold:

1. Sentence begins at the first word `[begin` and ends at the last word `end)`. 
2. `ByteRange`s one-to-one map with `string_view`s coming out of spiece.
3. Empty sentence is carefully accounted for, marking the exact position with an empty string-view/ByteRange to the point.

There is a unit-test accounting for the above behaviour, passing (in the GitHub CI, full marian). Please let me know if any more invalid inputs that I missed need coverage.

kpu:
We're optimizing for one source of truth instead of a whole bunch of potentially inconsistent ones.  You appear to have stumbled upon a reasonable solution which is that the end-of-sentence token can be mapped to the whitespace between sentences.  This would make sentences contiguous bytes and provide an alignment point for the end of sentence token.

jerinphilip:
I'm 100% convinced `ByteRange` needs to stick (mapping directly from the `string_view`s the container draws from). This means I can no longer work with the following design: 

```
// Byte offset where each token begins in the string.
std::vector<size_t> tokens_;
// Index into tokens_ where each sentence begins.  
std::vector<size_t> sentenceTokens_;
// Sentence i begins at text[tokens_[sentenceTokens_[i]]] and ends at text[tokens_[sentenceTokens_[i+1]]]
```
The design assumes we're not dropping characters between what we want as meaningful units (sentences, words) - as I said, we are - *multiple* spaces for words and *multiple* newlines for sentences. This means we need end of word marker and also end of sentence marker, making both `ByteRanges`, not `size_t`. If you have changes here please let me know, I'm stuck as I cannot work with the above without making a mess worse than when conditionals existed in the previous state (I'll be looping to strip multiple spaces and multiple newlines).

src/translator/sentence_ranges.h
```diff
@@ -46,24 +49,14 @@ class Annotation {
...
-
-  /// Returns indices of terminal (word) ByteRanges in sentenceIds_ of a
-  /// sentence corresponding to sentenceIdx. The distance can be used to compute
-  /// number of words in a sentence (numWords) and also to construct the
-  /// terminal ByteRanges (sentenceTerminals).
-  std::pair<size_t, size_t> sentenceTerminalIds(size_t sentenceIdx) const;
+  /// Stores indices onto flatByteRanges_ of where sentences end (not inclusive,
+  /// aligned with C++ half interval notions). There is a 0 marker to simplify
+  /// sources, indicating where the -1-th sentence ends.
+  std::vector<size_t> sentenceEndIds_;
```
abhi-agg, line 59: Please correct me if I am wrong anywhere below. I am taking a super simple example to understand the Annotation class.

Imagine a text containing 2 sentences: `"Ab cd. Ef gh."`
So in bytes, it is:
| Bytes      | 0   | 1   | 2   | 3   | 4   | 5   | 6   | 7   | 8   | 9   | 10  | 11  | 12  |
|------------|-----|-----|-----|-----|-----|-----|-----|-----|-----|-----|-----|-----|-----|
| Characters | "A" | "b" | " " | "c" | "d" | "." | " " | "E" | "f" | " " | "g" | "h" | "." |

And **assuming** the sentencepiece tokenized the individual sentences as:
Sentence 1 ("Ab cd.") = `"Ab_",    "cd",    "."`
Sentence 2 ("Ef gh.")= `"Ef_",    "gh",    "."`

(I assumed that the white space at byte 6 doesn't belong to any sentence when it comes to tokenization by sentencepiece because sentencepiece deals with only sentences. But I may be wrong.)

So `flatByteRanges_ = [(0,3), (3,5), (5,6), (7,10), (10,12), (12,13)]`

and `sentenceEndIds_ = [0, 6, 13]` which means

Sentence 1 = `(0, 6)`
Sentence 2 = `(6, 13)`

Am I right?

(Of course, all the ranges written above as `(x, x)` means half open interval `[x, x)`)

src/translator/sentence_ranges.cpp
```diff
@@ -6,48 +6,44 @@ namespace marian {
...
+  bosId = sentenceEndIds_[sentenceIdx]; // Half interval, so;
+  eosId = sentenceEndIds_[sentenceIdx + 1];
+  // Difference between eosId and bosId is the number of words.
+  return eosId - bosId;
 }
 
 ByteRange Annotation::sentence(size_t sentenceIdx) const {
-  auto terminals = sentenceTerminals(sentenceIdx);
-  return (ByteRange){terminals.first.begin, terminals.second.end};
+  size_t bosId, eosId;
```
abhi-agg, line 24: If a user wants information for the very 1st sentence of the text, what should be provided as `sentenceIdx`? 0 or 1? Asking because the substring `Idx` of `sentenceIdx` can be interpreted that it starts from 0 and it is a valid value.

It makes sense to add something about the acceptable values for `sentenceIdx` parameter in the API description and clarify this 👍

src/translator/sentence_ranges.h
jerinphilip: I'll try to help you here first conceptually, through composition.

```
typedef ByteRange Word;
typedef vector<Word> Sentence; // vector<ByteRange>, a single sentence
typedef vector<Sentence> AnnotatedText;   // vector<vector<ByteRange>, multiple sentences

AnnotatedText example; 
```

How do I get a sentence ByteRange, for the sentence at index `i`? 
* `(ByteRange){example[i].front().begin, example[i].last().end}`

Annotation is the same thing, except the `vector<vector` hammered into a flat structure to avoid multiple reallocs. This is done by having markers of indices where sentence ends in `flatByteRanges_` in addition.

> sentenceEndIds_ = [0, 6, 13]

You seem to be looking at them as end indices. No, they're indices pointing to the `flatByteRanges` structure. `sentenceEndIds_ = [2, 5]`. 

This additional indirection helps when you want to return `vector<Word>` for a sentence.

jerinphilip: There's an extra sentence marker to convert ifs into accessors at the moment, but the above should be simpler to understand. To the outside world, this container behaves just like `vector<vector<ByteRange>` with multiple sentences.


```
sentenceEndIds_ = [2, 5] => #sentences = [2, 5].size() = 2
```

src/translator/sentence_ranges.cpp
kpu: All public APIs should use 0 indexing.

abhi-agg: Thanks for clarifying. It means for the very 1st sentence of the text, sentenceIdx should be 0. Adding a line in API description just to clear this ambiguity would be helpful 👍

kpu:
As far as I understand, the concern is now multiple instances of whitespace.  What does SentencePiece actually return for that?  

Keep in mind SentencePiece tokens can and do include whitespace.  So e.g. if ``in the`` is a token then so is ``in  the`` which includes two spaces.

src/translator/sentence_ranges.h
abhi-agg: Thanks for explanation. This makes sense to me 👍 

> they're indices pointing to the `flatByteRanges` structure.

Got it.

> `sentenceEndIds_ = [2, 5]`.

I guess, it would be equal to `[0, 2, 5]` (because you are doing `sentenceEndIds_.push_back(0)` in the constructor)? Right? It doesn't matter from the public methods perspective but asking for the sake of correctness 👍

src/translator/sentence_ranges.cpp
abhi-agg: @jerinphilip I don't know whether this PR is the right place for this change or you want to do it in a separate PR. But just mentioning that for 1st sentence/word, `sentenceIdx`/`wordIdx` parameter should be 0 in the public member functions of `Annotation` and `AnnotatedText` classes would be helpful.

I don't want to block this PR just because of this change. Rest of the PR looks good to me 👍

jerinphilip: I'll update the doxygen docs for API along with this PR, there seems to be bigger problems than this at the current stage which kpu's sending me at.

jerinphilip:
@kpu Contiguity asserts/logs are placed on insertion (https://github.com/browsermt/bergamot-translator/pull/90). They hold when working with sentencepiece. So we don't lose text in between with SentencePiece. 

ssplit has '\n' in between, and contiguity doesn't hold. I think I'm filtering out empty sentences from translation, so there can be multiple newlines. We need an end of sentence marker at the very least.

At this stage the question changes to what should be the truth, it can be set accordingly.

src/translator/sentence_ranges.h
jerinphilip: As of now, yes, I think your understanding is correct. numSentences is adjusted to return size of this container  - 1 accordingly to fix the external API.

abhi-agg:
@jerinphilip I am not sure that I followed the conversation above. But, apart from updating the doxygen docs for the API, is there anything else that you still have to add to this PR?

jerinphilip:
Well, me and @kpu are in a bit of disagreement. My excuses have been wrong so far (potentially inconsistent source of truths, which probably needed getting to the bottom of, and that is done),  but I still think this structure should use `ByteRange` (mapping directly with `string_view`s it's built from) and not `size_t` with double indexing, which is the new proposal.

The current status:
1. There's no sentencepiece issue like I've been saying. However, at the risk of being called overengineering - if there are vocabularies which omit the space absent special character, we will be dropping space. I'm not sure if it's best to hard-couple to spiece here still.
2. The multiple newlines from ssplit has not been resolved yet, there needs to be an end of word marker for the last word which doesn't exist in the specified design (the structure given to me to work with allows only `size_t tokens_` going one further for whole, from what I understand not for each sentence). With this it gets messy with conditionals and loops in code (to strip newlines) while the intention of double-accessor is to omit such stuff through accessors. Any `size_t` based edits are stuck until this resolves.
3. I want to reuse `Annotation` for QE, Sheffield has confirmed in slack they're dropping spaces. So we're looking at `ByteRange` instead of `size_t` there and that's another reason to keep `size_t`. I expect them to provide quality scores and annotation, that the API is consistent between alignments and qualityscores should ease integration dev efforts (I think this should operate with subwords with space incorporated though, because you don't have the luxury of space separated "words" in many languages, which brings it to adhering to something like sentencepiece, but that's not up to me). 

I also think `ByteRange`s will be nice to reason with and maintain this source. I don't think the gains from halving storage (0.5x) or removing an if is worth dropping this (at least not right now, if this is a perf issue can come back to it later). So - waiting on inputs now before I can make changes here.

kpu:
I've been convinced by the argument of preserving whitespace between sentences.  (Not necessary to have it finished in the PR.)  A sentence can be the range of bytes that excludes whitespace.  I think that is what you want @jerinphilip though let's do that without if statements.  

Regarding quality estimation using different tokenization, I think that's going to be a bit of a mess and ideally we'd use the same tokenization for everything.  If that doesn't happen, quality estimation will need its own token information.  That can be procrastinated until after this PR.

src/translator/sentence_ranges.cpp
```diff
@@ -6,48 +6,44 @@ namespace marian {
...
+  bosId = sentenceEndIds_[sentenceIdx]; // Half interval, so;
+  eosId = sentenceEndIds_[sentenceIdx + 1];
+  ByteRange sentenceByteRange;
+
+  if (bosId == eosId) {
```
jerinphilip, line 29: @kpu This is the last if remaining, rest have been converted into accessors.

abhi-agg (approved):

src/translator/sentence_ranges.cpp
kpu: I think it would be cleaner if the internal representation had a sequence of contiguous tokens where there's an end-of-sentence token containing whitespace preserved from the source.  Then there wouldn't be a need for this if statement.  

Nonetheless, I am approving this for now because it does in net improve the code.

kpu (approved):
We'll fix the internal part later.

Issue #35, opened 2021-02-22
Texts are treated as sequential blocking requests without async queueing


jerinphilip:
Scouring through bergamot-translator-extension sources, it seems that browser team is trying to break texts into sentences and provide them as individual requests. This can lead to a significant slowdown, referred to in the extension's [release docs](https://github.com/mozilla-extensions/bergamot-browser-extension/releases/tag/v0.2.0).

> Translation speeds are temporarily low - about 150-200 words per second instead of expected 500-600 words per second

No threads/async prevents any of `marian::bergamot::Service`s request collect and pack batches efficiently mechanism, as long as in the current state of unified API's specification. There are potentially two solutions in my mind, enable threading/async queuing mechanisms or collect it as one string and do the processing somewhere in TranslationModel.

https://github.com/browsermt/bergamot-translator/blob/51f702ea6c849687488d6478b3bc428f0ad033ae/src/translator/TranslationModel.cpp#L60-L64

Opening the issue for discussion.
/cc @abhi-agg @motin

abhi-agg:
@jerinphilip We are breaking text into sentences because in current integration, the current ssplit-cpp (with pcre compiled from sources) doesn't split sentences.

Of course, we will send the blob of text as soon as I make ssplit-cpp functional (with pcre2 compiled from sources).

Re: No threads/async
We have already discussed that once we start working on making threads working for wasm, we will enable async requests as well.

abhi-agg:
@jerinphilip No more of breaking text into sentences [via this commit](https://github.com/browsermt/bergamot-translator/commit/462a850d8ae554e835f0e3e64aef897aa3f93dad). So this is resolved.

I am closing this issue.

kpu:
The subject line of the issue is "Texts are treated as sequential blocking requests without async queueing".  That's still a problem right?

abhi-agg:
As I said, once we start working on making threads work for wasm, we can enable async requests as well.

I don't mind keeping this issue open and resolve it once we make threads to work 👍

jerinphilip:
There is a workaround for this currently implemented in a draft PR.

https://github.com/browsermt/bergamot-translator/blob/32285aafb2a0815e730615d74c64625b1fb52008/src/translator/service.cpp#L125-L155

PR #87, opened 2021-04-03
WASM Bindings collapse

jerinphilip:
The PRs which were a prerequisite for this are now complete (#79, #85, #80).

WASM bindings meanwhile have updated with passing memory than filenames (#117). The branch / PR has been brought forward to synchronize with these changes as well.

(A force push cleans the commit history, which was bloated before. Translation* files are deleted)

Checklist (ideally before merge):
- [ ] ~~Sync changes with https://github.com/mozilla/bergamot-translator/pull/39 to generate CircleCI artifact~~.
- [ ] ~~Use artifact once again in https://github.com/mozilla-extensions/bergamot-browser-extension/pull/83 to ensure nothing breaks at extension.~~
- [ ] Seems extension is not up to date yet with the constructors being modified for ByteArray. So first two are probably looking far.
- [ ] Final touch-ups to remove typedefs


wasm/bindings/TranslationModelBindings.cpp
```diff
@@ -6,17 +6,21 @@
...
 
+typedef marian::bergamot::Service TranslationModel;
+typedef marian::bergamot::Response TranslationResult;
+
 // Binding code
 EMSCRIPTEN_BINDINGS(translation_model) {
   class_<TranslationModel>("TranslationModel")
-    .constructor<std::string>()
-    .function("translate", &TranslationModel::translate)
-	  .function("isAlignmentSupported", &TranslationModel::isAlignmentSupported)
```
jerinphilip, line 18: This is [unused](https://github.com/mozilla-extensions/bergamot-browser-extension/search?q=isAlignmentSupported) at extension.

src/translator/service.h
```diff
@@ -144,6 +144,10 @@ class Service {
   std::vector<Response> translateMultiple(std::vector<std::string> &&source,
           TranslationRequest translationRequest);
 
+  /// Returns if model is alignment capable or not. Exists for not breaking things at the extension.
+  // This function, what it does and the name sticks out like a thorn in the entire structure of this class.
+  bool isAlignmentSupported() const { return true; }
```
kpu, line 149: Implement correctly by checking for alignments in the model.

wasm/bindings/TranslationModelBindings.cpp
kpu: Leave it in, implement it correctly

wasm/bindings/TranslationModelBindings.cpp
```diff
@@ -6,17 +6,23 @@
...
-    ;
+      .constructor<std::string>()
+      .function("translate", &TranslationModel::translateMultiple)
+      .function("isAlignmentSupported",
+                &TranslationModel::isAlignmentSupported);
+  // ^ We redirect translateMultiple to translate instead. Sane API is
```
kpu, line 24: What are these comments?

jerinphilip: So the original thing I was told was to have `std::future<Response>` and async operations. `TranslationModel` changes these to a blocking API where they instead feed us multiple texts. I have asked why this because I can't infer from signature alone, now I do guesswork and I read this as "browser needs further granularity in texts but same speed".

So the disparity is:
```haskell
f :: InputType -> OutputType

Service::translate :: Str -> Future[Response]
TranslationModel::translate :: Vector[Str] -> Vector[Response]
```

To arrange parity, the following is what I did:

```haskell
Service::translateMultiple:: Vector[Str] -> Vector[Response]
```

I consider the async future to be the sane API from the feedback, input I have received so far. At wasm bindings, I'm binding `translateMultiple` instead to "translate" and plumbing to test compatibility at extension.

src/translator/service.h
jerinphilip: `options_` is a member in Service now, and `options_->hasAndNotEmpty("alignment")` is used to provide this. It's untested at the moment. 

Let know if there's something from the model I can use which is stronger.

XapaJIaMnu:
Looks good to me, @abhi-agg  would it break your workflow?

wasm/bindings/TranslationResultBindings.cpp
```diff
@@ -6,15 +6,17 @@
 #include <emscripten/bind.h>
 #include <vector>
 
-#include "TranslationResult.h"
+// #include "TranslationResult.h"
```
jerinphilip, line 9: Note to self, remove.

wasm/bindings/TranslationModelBindings.cpp
```diff
@@ -31,9 +35,11 @@ TranslationModel* TranslationModelFactory(const std::string &config,
 EMSCRIPTEN_BINDINGS(translation_model) {
   class_<TranslationModel>("TranslationModel")
     .constructor(&TranslationModelFactory, allow_raw_pointers())
-    .function("translate", &TranslationModel::translate)
+    .function("translate", &TranslationModel::translateMultiple)
```
XapaJIaMnu, line 38: Is this api change?

XapaJIaMnu (approved):
Other than not breaking mozilla's workflow i'm happy

wasm/bindings/TranslationModelBindings.cpp
jerinphilip: We (`Service`) have a future returning API (translate) which is expected to work with async when time comes. It's been deferred for other changes now. `TranslationModel` (Mozilla API) converted this into a blocking API, return `vector<Response>` for texts. 

Collapse thus does `TranslationModel::translate` -> `Service::translateMultiple` (which can handle multiple text -> multiple Responses). This gives them the benefits of proper batching compiling from multiple requests which didn't exist before. The signatures were different enough to create a new function in Service for the same. 

From a layman point of view, they can now translate individual elements (and not whole page) without degrading speed. This is therefore also a workaround for https://github.com/browsermt/bergamot-translator/issues/35 which came as a byproduct.

abhi-agg:
@jerinphilip Since this PR affects the bindings, I would like to be a reviewer of this PR. Therefore, adding myself as a reviewer. I hope that is fine with you 👍

app/bergamot-translator-app-bytearray.cpp
```diff
@@ -20,19 +20,20 @@ int main(int argc, char **argv) {
   std::string config = options->asYamlString();
 
   // Route the config string to construct marian model through TranslationModel
-  TranslationModel model(config, marian::bergamot::getModelMemoryFromConfig(options));
+  marian::bergamot::Service model(
+      config, marian::bergamot::getModelMemoryFromConfig(options));
```
abhi-agg, line 24: We should also pass shortlist by bytes while calling this constructor. Right? Otherwise, it will be using only model as bytes but shortlist will still be loaded from file.

app/bergamot-translator-app-bytearray.cpp
```diff
@@ -20,19 +20,20 @@ int main(int argc, char **argv) {
...
-        texts.emplace_back(line);
+    texts.emplace_back(line);
   }
 
-  auto results = model.translate(std::move(texts), translationRequest);
+  auto results = model.translateMultiple(std::move(texts), translationRequest);
 
   // Resolve the future and get the actual result
-  //std::vector<TranslationResult> results = futureResults.get();
+  // std::vector<TranslationResult> results = futureResults.get();
```
abhi-agg, line 36: We can remove these lines. Right?

app/bergamot-translator-app.cpp
```diff
@@ -21,19 +22,19 @@ int main(int argc, char **argv) {
...
-        texts.emplace_back(line);
+    texts.emplace_back(line);
   }
 
-  auto results = model.translate(std::move(texts), translationRequest);
+  auto results = model.translateMultiple(std::move(texts), translationRequest);
 
   // Resolve the future and get the actual result
-  //std::vector<TranslationResult> results = futureResults.get();
+  // std::vector<TranslationResult> results = futureResults.get();
```
abhi-agg, line 37: Do we need these lines?

app/bergamot-translator-app-bytearray.cpp
jerinphilip: I'd prefer to have @qianqianzhu bring this change in a separate PR (she's working on vocabs and will tread the bytearray test-app territory meanwhile). Does this suggestion work for everyone? 

I'm not changing anything here except my 80 column wrap setting running over Nick's settings, if a closer look is taken.

app/bergamot-translator-app-bytearray.cpp
jerinphilip: Yes, will do. 😃

app/bergamot-translator-app.cpp
```diff
@@ -21,19 +22,19 @@ int main(int argc, char **argv) {
...
   TranslationRequest translationRequest;
   std::vector<std::string> texts;
 
   for (std::string line; std::getline(std::cin, line);) {
-        texts.emplace_back(line);
+    texts.emplace_back(line);
   }
 
-  auto results = model.translate(std::move(texts), translationRequest);
+  auto results = model.translateMultiple(std::move(texts), translationRequest);
```
abhi-agg, line 34: Just saying. This change will require a change in the extension code otherwise extension code will not work out of box.

src/TranslationModel.h
```diff
@@ -1,80 +0,0 @@
...
-// All local project includes
-#include "TranslationRequest.h"
-#include "TranslationResult.h"
-#include "translator/definitions.h"
-#include "translator/service.h"
-
-/* A Translation model that translates a plain (without any markups and emojis)
- * UTF-8 encoded text. This implementation supports translation from 1 source
- * language to 1 target language.
- */
```
abhi-agg, line 26: @jerinphilip Could you add this text in the documentation of Service class somewhere? This is an important piece of information that seems to be missing in Service class (correct me if I am wrong) 👍

src/TranslationModel.h
```diff
@@ -1,80 +0,0 @@
...
-class TranslationModel {
-public:
-  /* Construct the model using the model configuration options as yaml-formatted
-   * string
-   */
-  /**
-   * @param config Marian yml config file in the form of a string
```
abhi-agg, line 33: Same for this (in the constructors of `Service` class that uses `std::string` for `config`)?

app/bergamot-translator-app-bytearray.cpp
qianqianzhu: Yes. Better leave it to me as I am adding vocab bytes as a new argument here.

src/TranslationModel.h
```diff
@@ -1,80 +0,0 @@
...
-   * joined back together and returned in TranslationResult.
-   *
-   * Please refer to the TranslationRequest class to find out what additional
-   * information can be requested. The alignment information can only be
-   * requested if the model supports it (check isAlignmentSupported() API).
-   *
-   * The texts argument will become empty after the execution of this API (each
-   * entry of texts list will be moved to its corresponding TranslationResult
-   * object).
-   */
```
abhi-agg, line 66: I am not sure if this text can be used somewhere in the Service class. Do you see something that is present in this text but not in Service class and can add value there?

app/bergamot-translator-app.cpp
jerinphilip: Haven't I aliased `translate` -> `Service::translateMultiple` in bindings? I'm hoping this is the change you're talking about.

Build is failing in Extension's CircleCI (https://app.circleci.com/pipelines/github/mozilla-extensions/bergamot-browser-extension/204/workflows/ab129537-ad34-403b-b31d-40c626c19707/jobs/206/steps). But I'm guessing these are expected due to the issues discussed in https://github.com/mozilla-extensions/bergamot-browser-extension/pull/83. 

Can you confirm?

app/bergamot-translator-app.cpp
jerinphilip: Removed.

app/bergamot-translator-app.cpp
jerinphilip: image

Well manually downloaded XPI, seems working.

src/TranslationModel.h
jerinphilip: This is already done, marking as resolvd.

src/TranslationModel.h
jerinphilip: Absorbed a little bit that is still valid.

src/TranslationModel.h
jerinphilip: This text is also absorbed.

app/bergamot-translator-app.cpp
abhi-agg: @jerinphilip We can test it with the [wasm test page](https://github.com/browsermt/bergamot-translator/tree/main/wasm#-demo-see-everything-in-action) because that is using the latest byte based API. If it works with that then we are good to go 👍

jerinphilip: image

I have no clue what is happening, but I am assuming bindings transfer is intact, given I/O works?

abhi-agg: > @jerinphilip We can test it with the [wasm test page](https://github.com/browsermt/bergamot-translator/tree/main/wasm#-demo-see-everything-in-action) because that is using the latest byte based API. If it works with that then we are good to go 👍

@jerinphilip Let me help you with testing. I will test it at my end and confirm 👍

abhi-agg: @jerinphilip I tested your change with the fix in https://github.com/browsermt/bergamot-translator/pull/125 and it works 👍

abhi-agg (approved):

app/bergamot-translator-app.cpp
jerinphilip: I have pulled #125 into a [test build](https://github.com/browsermt/bergamot-translator/tree/jp/collapse-yolo), still translates to "schwanger". I guess the transfer is intact from your response and this is something to resolve at my end maybe.

XapaJIaMnu (approved):

PR #91, opened 2021-04-08
Consistent api usage in app

XapaJIaMnu:
@abhi-agg @jerinphilip if you notice the app folder has two versions of the `service-cli` and the `bergamot-app`. One running with a binary file, one running with non-binary file. When changing the API please keep them consistent.

Alternatively, (and even better), we should have just one version of the executables, but as they require different calls to the constructor, and the arguments are being passed to the marian config generator, as opposed to being read by us, it is easier for now to just update 2 files. They are identical, copy paste should work.


jerinphilip:
@XapaJIaMnu Let's separate this, we will have more binaries though, somehow it's undesirable (https://github.com/mozilla/bergamot-translator/pull/38#issuecomment-805640362).

1. Just test source.text = target.text (base test-application)
    a. a variant of this with Model, ShortList and tentatively Vocabulary binary loads which tests the bytearray loads.
2. I will bring independent test-apps for:
    a. Alignment
    b. QualityScores.
    
We can reuse the test-outputs for 1 and 1a. Does this sound like a reasonable thing to do @XapaJIaMnu?

XapaJIaMnu:
That can be done later, for now let's pretend for a bit that our tests run so I can add new features and test them.

app/bergamot-translator-app-bytearray.cpp
```diff
@@ -20,44 +20,22 @@ int main(int argc, char **argv) {
   std::string config = options->asYamlString();
 
   // Route the config string to construct marian model through TranslationModel
-  TranslationModel model(config, marian::bergamot::getModelMemoryFromConfig(options));
+  auto model = std::make_shared<TranslationModel>(config, marian::bergamot::getModelMemoryFromConfig(options));
```
jerinphilip, line 23: Kenneth wanted the shared_pointer off, so reset to the left one? Is the shared_pointer required?

XapaJIaMnu: I just made it consistent with the other non-bytearray file. I can remove it from both

jerinphilip: Reset to variable, no pointer was the call, which is why this bytearray file had no shared-pointer.

XapaJIaMnu: Done

abhi-agg:
> When changing the API please keep them consistent.

Sure 👍

jerinphilip (approved):

abhi-agg (approved):

PR #93, opened 2021-04-08
Moving small tests to GitHub CI

### LINKED ISSUE #92: Transfer tiny tests to GitHub CI
Jenkins@vali is dead. While I bring it back up, it seems reasonable to move some tests to GitHub CI. There's also the plus of ensuring a clean environment regression-tests every single time, unlike Jenkins which caches models etc.

> Many of the tests run quickly on native right? Why aren't these github actions?

Maybe not a bad idea, with 30min debug cycles we're looking at: 

1. Use upload artifacts to generated expected output on GitHub runners.
2. Change expected outputs on bergamot-translator repository with the uploaded artifacts.

jerinphilip:
Fixes #92.

GitHub ubuntu runners are `avx2`, for which we already seem to have outputs. Artifacts are set to be uploaded, so if instruction set on runners change then we can update from uploaded artifacts.

This covers only Ubuntu.


jerinphilip:
Looks like we get some `avx512` as well. @XapaJIaMnu, what to do here? What's the intgemm path? Is there one using avx512 where I add this to test expected?

```
+ /home/runner/work/bergamot-translator/bergamot-translator/bergamot-translator-tests/tools/diff.sh bergamot-bytearray.intgemm_8bit.avx512.out bergamot.intgemm_8bit.avx512.expected
Command: /home/runner/work/bergamot-translator/bergamot-translator/bergamot-translator-tests/tools/diff.sh /home/runner/work/bergamot-translator/bergamot-translator/bergamot-translator-tests/tests/basic/bergamot-bytearray.intgemm_8bit.avx512.out /home/runner/work/bergamot-translator/bergamot-translator/bergamot-translator-tests/tests/basic/bergamot.intgemm_8bit.avx512.expected
diff: bergamot.intgemm_8bit.avx512.expected: No such file or directory
```

XapaJIaMnu:
I'm not sure what you mean by intgemm paths. We have separate 8bit codepaths for ssse3, avx2, avx512 and avx512vnni
There were no outputs created for the avx512 case. You can fetch some from sigyn:
```
/mnt/sigyn0/nbogoych/bergamot-translator/bergamot-translator-tests/tests/basic$ ls | grep .out
bergamot-bytearray.intgemm_8bit.avx512.out
bergamot.intgemm_8bit.avx512.out
service-cli-bytearray.intgemm_8bit.avx512.out
service-cli.intgemm_8bit.avx512.out
```

jerinphilip:
I've configured to upload artifacts, so whatever the runner has will be uploaded. I can download and do things, no more sigyn or anything.

XapaJIaMnu:
You have to put the gold standard tests beforehand, right? Bear in mind that especially when it comes to FP precision, the compiler version does matter.

jerinphilip:
> Bear in mind that especially when it comes to FP precision, the compiler version does matter.

I foresee this changing every now and then (with compiler/OS upgrades and whatnot), so just keeping the hooks/automation to manually inspect and move it around as well. I'll never get an M1-mac, maybe GitHub has a runner? There's a human moving this around (here -> brt -> bergamot-translator) who's supposed to check (I don't think I'm qualified yet given all the devil in details that surfaces when I start on this, feel free to take up the mantle). The compilers / OS here and the generated outputs eventually reach consistency / fixed-point? 

We'll configure the parameters (compilers, arch as var) into one workflow and output file-names and have it cover the ground? I'm only concerned if something changes (it might when GitHub runners create upgrades and chaos, where we'll try to inspect and justify) between commits at outputs. Not that it works covers all architectures.

jerinphilip:
The run of workflow corresponding to modifications are in: 
* https://github.com/browsermt/bergamot-translator/actions/runs/732886781

This completes the build + regression-test pipeline on Jenkins previously moved over here. There are a few more instruction set output to be accounted for in the tests-repo perhaps.

.github/workflows/native-full_marian-ubuntu.yml
```diff
@@ -87,34 +61,89 @@ jobs:
         CC=/usr/bin/gcc-${{ matrix.gcc }} CXX=/usr/bin/g++-${{ matrix.gcc }} CUDAHOSTCXX=/usr/bin/g++-${{ matrix.gcc }} \
         cmake .. \
           -DBoost_ARCHITECTURE=-x64 \
-          -DCMAKE_BUILD_TYPE=Release \
+          -DCMAKE_BUILD_TYPE=Slim \
```
abhi-agg, line 64: Any specific reason to change it to `Slim`?

jerinphilip: Binaries (artifacts uploaded) are 2GB with debug symbols. It's `Slim` that provides no `-g` and hence tinier artifacts. GitHub CI doesn't require `-g` I hope. :-)

.github/workflows/native-full_marian-ubuntu.yml
```diff
@@ -87,34 +61,89 @@ jobs:
...
           -DCOMPILE_CPU=${{ matrix.cpu }} \
           -DCOMPILE_CUDA=${{ matrix.gpu }} \
-          -DCOMPILE_EXAMPLES=on \
-          -DCOMPILE_SERVER=on \
+          -DCOMPILE_EXAMPLES=off \
```
abhi-agg, line 67: `COMPILE_EXAMPLES` is set to `on` [here](https://github.com/browsermt/bergamot-translator-tests#jenkins). Am I missing something?

jerinphilip: Don't need the example binaries over here. I copied over the old marian workflow files while I was less knowledgeable regarding the ecosystem to do something which was minimum invasive. 

Since the tests at browsermt/marian-dev covers that these build anyway and I'm not using the binaries for any tests here, I see no point in generating these (some time is saved as well in build I hope). Will set them back if and when required in future (ie if I use them in tests).

The overall build time in this path has been brought down from 30m to 19m with these optimizations (as this was making progress in this PR slow).

abhi-agg: Makes sense 👍 

In this case, we should update the [README](https://github.com/browsermt/bergamot-translator-tests#readme) in bergamot-translator-tests repository because it is misleading 😄

.github/workflows/native-full_marian-ubuntu.yml
abhi-agg: Marian [enables NDEBUG](https://github.com/browsermt/marian-dev/blob/master/CMakeLists.txt#L266) for `Slim` builds which will remove [asserts()](https://en.cppreference.com/w/cpp/error/assert) in your binaries.

I understand your problem though. May be keep a release build and a separate workflow for slim builds just for the regression tests?

I am not sure what is the right call here.

kpu: `Release` is set by Marcin in an attempt to minimize support e-mails.  IMHO Release should not have assertions or `-g` either.  But I agree it's good to have perhaps a separate job with less expensive testing that verifies assertions are still met.

jerinphilip: > I agree it's good to have perhaps a separate job with less expensive testing that verifies assertions are still met.

So I clubbed the regression test into the same workflow avoiding the problem of big binaries. In addition, the native minimal marian config has been collapsed (so we'll only have two workflow files sharing the specs for regression-tests and install, varying cmake build in the matrix strategy). Sorry for the additional review load, but this should move the state of tests to an overall better state I hope. 

I request @abhi-agg to revisit https://github.com/browsermt/bergamot-translator-tests/pull/7 (I've pushed a few changes behind the previous state, and we're now ready to merge this). I'm placing the the README update behind this.

With this, I'm also allowing a tag-based regression-testing for the minimal/custom marian (testing only bergamot-translator-app, switching the build by ifdefs through CMake options).

abhi-agg: > I request @abhi-agg to revisit [browsermt/bergamot-translator-tests#7](https://github.com/browsermt/bergamot-translator-tests/pull/7) (I've pushed a few changes behind the previous state, and we're now ready to merge this).

Approved 👍 

> I'm placing the the README update behind this.

Awesome 👍

.github/workflows/native-mac.yml
```diff
@@ -0,0 +1,125 @@
...
+              TEST_TAGS: ""
+
+          - name: "minimal-marian"
+            os: macos-10.15
+            cmake: 
+              CMAKE_BUILD_TYPE: "Release"
+              COMPILE_TESTS: "ON"
+              USE_WASM_COMPATIBLE_SOURCE: "OFF"
+              USE_FBGEMM: "OFF"
+              USE_STATIC_LIBS: "OFF"
```
abhi-agg, line 46: `USE_WASM_COMPATIBLE_SOURCE` and `USE_STATIC_LIBS` will be `ON` for custom marian.

.github/workflows/native-ubuntu.yml
```diff
@@ -0,0 +1,146 @@
...
+              TEST_TAGS: ""
+
+          - name: "minimal-marian"
+            os: ubuntu-latest
+            gcc: 8
+            cpu: 'ON'
+            gpu: 'OFF'
+            cmake:
+              CMAKE_BUILD_TYPE: "Release"
+              COMPILE_TESTS: "OFF"
```
abhi-agg, line 48: `COMPILE_TESTS` will be `ON` for this. Right?

Additionally, having "minimal-marian" here exactly the same as "minimal-marian" of `native-mac.yml` would be nicer. This one is up to you though 👍

jerinphilip: Wasn't sure of the implications, `ON` works for me. Thanks for pointing out. Will try to set it accordingly. 

I thought the idea was to just have the bare minimum required - i.e the library for WASM (I'm unsure what works in the minimal marian-path due to the CMake switches).

.github/workflows/native-mac.yml
jerinphilip: There's a comment above for mac file complaining about Static libs and it was OFF. My bad regarding the USE_WASM_COMPATIBLE_SOURCE switch. Thanks for pointing out. Will fix that one.

jerinphilip:
```
    # Openblas location is exported explicitly because openblas is keg-only,
    # which means it was not symlinked into /usr/local/.
    # CMake cannot find BLAS on GitHub runners if Marian is being compiled
    # statically, hence USE_STATIC_LIBS=off
```

The above is from source before. I had build failures without this. Although it seems to be working in minimal/custom marian without the need of this I'll try the variations here as well.

.github/workflows/native-ubuntu.yml
```diff
@@ -0,0 +1,146 @@
...
+    # Configure hooks to upload artifacts if the build is Slim
+    - name: Upload build artifacts
+      uses: actions/upload-artifact@v2
+      with: 
+        name: bergamot-translator-artifacts
+        path: |
+            ./build/app/service-cli*
+            ./build/app/bergamot-translator-app*
+            ./build/app/marian-decoder-new
+      if: matrix.cmake.CMAKE_BUILD_TYPE == 'Slim'
```
abhi-agg, line 128: I understood now. Regression tests are running for "full-marian-slim" only. Right?

.github/workflows/native-ubuntu.yml
abhi-agg: I think if COMPILE_TESTS=ON works for "minimal-marian" then we can keep it ON. One less thing to worry 👍

.github/workflows/native-ubuntu.yml
jerinphilip: No, they're running for everything. It is only artifact upload is reduced to 'slim' (I don't want to upload big binaries with debug symbols). Not that since regression-tests are now in the same matrix entry, I don't need to upload for regression-tests.

This is more a future hook not complete, we can cut this out if it's causing any harm (my plans are to ship executables compatible with multiple OS (FYI, @kpu). Maybe a REST server that can be downloaded directly and used in user-testing to compare with WASM single thread as well?

abhi-agg: I think I am not clear then what is happening 😄

I was under the impression that uploading build artifacts step has something to do with running regression-tests. But looks like it is not?

kpu: I'm confused too.  That demo about a multiplatform executable from somebody on the Internet was fun and all, but what does it have to do with the regression tests?

jerinphilip: So, when I set to Slim in a two workflow mode, it was suggested we maintain Release. The problem with Release are really big binaries which I don't want to move around and long-compile times.

1. Solve the big binary problem by keeping build and regression-test as steps in the same file. I have two builds here, Release and Slim now. We no longer require upload artifacts for the binaries, they're available in the same workflow-run.
2. Now we have growing list of files per workflow and inconsistencies between them. As a lazy programmer, I don't like this. I pick common parameters and collapse them into GitHub CI Matrix. While I'm at this, I see I can improve maintainability and have regression-tests in native as well with minimum effort (there's only one place which requires edits). So I collapse native into the same file as well. 
3. I choose to have some upload-artifacts code left from before, than cut it off, I keep it seeing the possibilities. Not that it immediately needs to be done, just putting out the possibilities of leaving it around. That is the upload code we're commenting on.

abhi-agg: Okay. In this case, my suggestion would be to remove the uploading artifacts step from both of these files if it is not serving any purpose right now. It is a super small code that can be added later as and when it is needed. We will keep the confusion away this way. However, I leave the decision up to you 👍

jerinphilip: I guess I'll remove it to avoid confusion for now. :D

.github/workflows/native-mac.yml
jerinphilip: So, seems this is leading to failures, but not from places expected (blas):
https://github.com/browsermt/bergamot-translator/runs/2333377746
Does this need more investigation?

.github/workflows/native-ubuntu.yml
jerinphilip: So compile tests seem to be offending sqlite is my conclusion, can I disable them for minimal marian?

abhi-agg: Go ahead then 👍

.github/workflows/native-ubuntu.yml
```diff
@@ -0,0 +1,135 @@
...
+
+          - name: "minimal-marian"
+            os: ubuntu-latest
+            gcc: 8
+            cpu: 'ON'
+            gpu: 'OFF'
+            test_tags: "'#wasm'"
+            cmake:
+              CMAKE_BUILD_TYPE: "Release"
+              COMPILE_TESTS: "OFF" # Minimal marian has no sqlite support and COMPILE_TEST=ON fails.
```
abhi-agg, line 49: @jerinphilip One final ask. Could you please add USE_STATIC_LIBS: "ON" here as well? It will make both mac and ubuntu workflow files consistent. It is a nit so you can ignore it as well.

Rest of the PR looks great. I can approve it 👍

jerinphilip: So mac was the exception, because of the BLAS issue. The default is USE_STATIC_LIBS: ON in the repo, and ubuntu workflows don't use the switch. I can add a note if necessary, as this key is not used anywhere. The note is supposed to exist for Mac I think which is the exception due to BLAS issues. For Mac, I  believe the minimal marian is circumventing this in the other mode using supplied BLAS at the moment.

https://github.com/browsermt/bergamot-translator/blob/2e5daac978f535b14bf32997e563f3bf430efbfb/CMakeLists.txt#L24

.github/workflows/native-mac.yml
```diff
@@ -0,0 +1,116 @@
...
+          brew install openblas protobuf coreutils
+
+    # Openblas location is exported explicitly because openblas is keg-only,
+    # which means it was not symlinked into /usr/local/.
+    # CMake cannot find BLAS on GitHub runners if Marian is being compiled
+    # statically, hence USE_STATIC_LIBS=off
+    - name: Configure CMake
+      run: |
+        export LDFLAGS="-L/usr/local/opt/openblas/lib"
+        export CPPFLAGS="-I/usr/local/opt/openblas/include"
```
abhi-agg, line 74: Just one comment. As openblas is not used when `USE_WASM_COMPATIBLE_SOURCE` is `ON`, probably setting LDFLAGS and CPPFLAGS conditionally makes sense (the same way as you are doing with installing MKL in `native-ubuntu.yml` file).

I am suggesting it because I am not sure how having these flags set affect when USE_WASM_COMPATIBLE_SOURCE is ON and I don't want to take a risk here 👍 
What do you think?

.github/workflows/native-ubuntu.yml
abhi-agg: Got it.

.github/workflows/native-mac.yml
jerinphilip: Using `$GITHUB_ENV` and an `if` to ensure this doesn't trigger in the `WASM_COMPATIBLE_SOURCES` workflow. This should achieve the intended effect?

abhi-agg: I think it did the job 👍

.github/workflows/native-mac.yml
```diff
@@ -0,0 +1,120 @@
...
+            os: macos-10.15
+            test_tags: ""
+            cmake: 
+              CMAKE_BUILD_TYPE: "Slim"
+              COMPILE_TESTS: "ON"
+              USE_WASM_COMPATIBLE_SOURCE: "OFF"
+              USE_STATIC_LIBS: "OFF"
+              USE_FBGEMM: "OFF"
+              COMPILE_SERVER: "OFF"
+              COMPILE_EXAMPLES: "OFF"
```
abhi-agg, line 37: @jerinphilip So, now there are 4 configs in CI:

`full-marian`
`full-marian-slim`
`minimal-marian`
`wasm`

Do you still want to keep `full-marian-slim` now that you are not uploading any artifacts (because of which this was added in the first place)?

I am fine with keeping it if you have reasons for it. But in this case, CI will be doing 8 checks (instead of 6 if you don't) which will make it slower.

jerinphilip: CI launches everything in parallel, Slim is faster to check. I don't have much strong opinions here, I just used Slim for faster turnaround cycles while editing only CI. Note that I was doing a lot of trial and error in this PR and 7 mins gain seemed nice at the point :D.

image

It's quite simple to add remove at the current stage of the yaml (with matrix), so just let me know what is the call.

abhi-agg: I completely understand your perspective here 👍 
If you are fine with it then let's remove `full-marian-slim`.

jerinphilip:
Reminder before final approval:

- [ ] Update README in bergamot-translator-tests repo (https://github.com/browsermt/bergamot-translator-tests/pull/21)
- [ ] Point submodule to master

abhi-agg:
> Reminder before final approval:
> 
>     * [ ]  Update README in bergamot-translator-tests repo ([browsermt/bergamot-translator-tests#21](https://github.com/browsermt/bergamot-translator-tests/pull/21))
> 
>     * [ ]  Point submodule to master

@jerinphilip Do you want to include these in this PR or a separate PR?

jerinphilip:
Let's include it, seems tiny.

abhi-agg (approved):
@jerinphilip Thanks for doing all the changes 👍

abhi-agg:
@kpu @XapaJIaMnu Could any one of you please review and approve the PR? We need 1 more approval to merge it. Thanks.

.github/workflows/native-ubuntu.yml
```diff
@@ -0,0 +1,122 @@
...
+    - name: Install MKL
+      run: |
+        wget -qO- "https://apt.repos.intel.com/intel-gpg-keys/GPG-PUB-KEY-INTEL-SW-PRODUCTS-2019.PUB" | sudo apt-key add -
+        sudo sh -c "echo deb https://apt.repos.intel.com/mkl all main > /etc/apt/sources.list.d/intel-mkl.list"
+        sudo apt-get update -o Dir::Etc::sourcelist="/etc/apt/sources.list.d/intel-mkl.list"
+        sudo apt-get install -y --no-install-recommends intel-mkl-64bit-2020.0-088
+      if: matrix.cmake.USE_WASM_COMPATIBLE_SOURCE == 'OFF'
+
+    # The script simplifies installation of different versions of CUDA
+    - name: Install CUDA
```
XapaJIaMnu, line 72: Wait, we test GPU as well? Weren't all of our tests CPU?

.github/workflows/native-ubuntu.yml
```diff
@@ -0,0 +1,122 @@
...
+        cmake .. \
+          -DCMAKE_BUILD_TYPE=${{ matrix.cmake.CMAKE_BUILD_TYPE }}\
+          -DCOMPILE_TESTS=${{ matrix.cmake.COMPILE_TESTS }}\
+          -DCOMPILE_EXAMPLES=${{ matrix.cmake.COMPILE_EXAMPLES }} \
+          -DCOMPILE_SERVER=${{ matrix.cmake.COMPILE_SERVER }} \
+          -DUSE_WASM_COMPATIBLE_SOURCE=${{ matrix.cmake.USE_WASM_COMPATIBLE_SOURCE }} \
+
+    - name: Compile bergamot-translator
+      working-directory: build
+      run: make -j2
```
XapaJIaMnu, line 92: Is this the limit of the CL machine capacity?

.github/workflows/wasm-custom_marian-ubuntu.yml
```diff
@@ -2,9 +2,9 @@ name: WASM (Custom Marian) Ubuntu
 
 on:
   push:
-    branches: [ main ]
+    branches: [ main, ci-sandbox ]
```
XapaJIaMnu, line 5: I didn't folow the discussion  beforehand, but why do we have the separate extra branch here (ci-sandbox)?

XapaJIaMnu (commented):
Several small questions before approving.

.github/workflows/wasm-custom_marian-ubuntu.yml
jerinphilip: So far I can only trial and error through GitHub CI (no SSH or shell) to get a script working. This (a) pollutes commit history (b) spams people subscribed to PR while I do edits here. After enough of this, I figured it's easier to leave another branch permanently where CI can be triggered by commits, and branch also allows force pushes (not main, for obvious reasons).

In essence use the branch if you can't pull request against main and trigger CI.

.github/workflows/native-ubuntu.yml
jerinphilip: I understand this is not future proof. I'm trying `make -j` and will make it stick if it works. Hoping we won't run into out of memory issues. 

Limits are [here](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources).

.github/workflows/native-ubuntu.yml
jerinphilip: Removed to avoid confusion. This was carried over from marian workflows.

.github/workflows/native-ubuntu.yml
XapaJIaMnu: Then let's keep -j2
j3 will most likely crash with 7 GB of ram and -j is not a good idea:

from the make manual
> If the ‘-j’ option is followed by an integer, this is the number of recipes to execute at once; this is called the number of job slots. If there is nothing looking like an integer after the ‘-j’ option, there is no limit on the number of job slots. The default number of job slots is one, which means serial execution (one thing at a time).

.github/workflows/native-ubuntu.yml
XapaJIaMnu: yes, we should then not put cuda, it will speed up the setup time.

XapaJIaMnu (approved):
Good now

.github/workflows/native-ubuntu.yml
jerinphilip: I reverted the commit which did the trial, so I'm guessing if all green we're good. Marking this as resolved.

abhi-agg (approved):

PR #94, opened 2021-04-08
Rudimentary validator for binary files

XapaJIaMnu:
Validate that the binary file you are trying to load is not truncated.

### DIFF
--- a/src/translator/byte_array_util.cpp
+++ b/src/translator/byte_array_util.cpp
@@ -76,7 +76,7 @@ bool validateBinaryModel(AlignedMemory& model, uint64_t fileSize) {
   }
 }
 
-} // Annonymous namespace
+} // Anonymous namespace
 
 AlignedMemory loadFileToMemory(const std::string& path, size_t alignment){
   uint64_t fileSize = filesystem::fileSize(path);
--- a/src/translator/byte_array_util.cpp
+++ b/src/translator/byte_array_util.cpp
@@ -5,6 +5,8 @@
 namespace marian {
 namespace bergamot {
 
+namespace {
+
 // This is a basic validator that checks if the file has not been truncated
 // it basically loads up the header and checks
 
@@ -73,6 +75,8 @@ bool validateBinaryModel(AlignedMemory& model, uint64_t fileSize) {
     return false;
   }
 }
+
+} // Annonymous namespace
 
 AlignedMemory loadFileToMemory(const std::string& path, size_t alignment){
   uint64_t fileSize = filesystem::fileSize(path);
--- a/src/translator/byte_array_util.cpp
+++ b/src/translator/byte_array_util.cpp
@@ -5,6 +5,75 @@
 namespace marian {
 namespace bergamot {
 
+// This is a basic validator that checks if the file has not been truncated
+// it basically loads up the header and checks
+
+// This struct and the getter are copied from the marian source, because it's located
+// inside src/common/binary.cpp:15 and we can't include it.
+struct Header {
+  uint64_t nameLength;
+  uint64_t type;
+  uint64_t shapeLength;
+  uint64_t dataLength;
+};
+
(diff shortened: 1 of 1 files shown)


src/translator/byte_array_util.cpp
```diff
@@ -5,6 +5,75 @@
 namespace marian {
 namespace bergamot {
 
+// This is a basic validator that checks if the file has not been truncated
+// it basically loads up the header and checks
+
+// This struct and the getter are copied from the marian source, because it's located
+// inside src/common/binary.cpp:15 and we can't include it.
```
jerinphilip, line 12: /// to be picked up by doxygen, consistent with what exists in source.

XapaJIaMnu: I don't want this to be picked up by doxygen? It's an internal function, not exposed by the headers and only really meant to be read by the developers.

src/translator/byte_array_util.cpp
```diff
@@ -5,6 +5,75 @@
...
+template <typename T>
+const T* get(const void*& current, uint64_t num = 1) {
+  const T* ptr = (const T*)current;
+  current = (const T*)current + num;
+  return ptr;
+}
+
+bool validateBinaryModel(AlignedMemory& model, uint64_t fileSize) {
+  const void * current = &model[0];
+  uint64_t memoryNeeded = sizeof(uint64_t)*2; // We keep track of how much memory we would need if we have a complete file
```
jerinphilip, line 30: wrap lines with comments to 80 characters is standard? My config overwrites this autoformat and I always end up bearing the brunt.

XapaJIaMnu: Dude, we're not 1998, we have 4k monitors. Just. No. All modern projects (including marian) have abandoned that.

src/translator/byte_array_util.cpp
kpu: Anonymous namespace please.  

The standard for documentation is higher in this repo than it is elsewhere, sorry @XapaJIaMnu.

src/translator/byte_array_util.cpp
kpu: I'm ok with longer lines.

src/translator/byte_array_util.cpp
XapaJIaMnu: Sorry, I genuinely do not understand how I should document this better. I'm went through the repo and looked at various files: (this one comes to mind) https://github.com/browsermt/bergamot-translator/blob/main/src/translator/request.h

Do I need to add the `defines` bit on top?

jerinphilip: I think what @kpu means is to either leave it in an anonymous namespace if it's local and don't bother documenting, or fit it with doxygen compatible documentation.

[This](https://github.com/browsermt/bergamot-translator/blob/jp/response-builder/src/translator/request.h) is the request which is documented. The target is a sphinx doc build, a demo of which looks like [this](https://jerinphilip.github.io/bergamot-translator/api/namespace_marian__bergamot.html).

XapaJIaMnu: Ok that should be fixed now.

src/translator/byte_array_util.cpp
```diff
@@ -5,6 +5,79 @@
...
+
+  // If this final check passes, the file is at least big enough to contain the model
+  if (fileSize >= memoryNeeded) {
+    return true;
+  } else {
+    return false;
+  }
+}
+
+} // Annonymous namespace
```
kpu, line 79: } // namespace

kpu (approved):

abhi-agg (approved):
Approving this change (as Kenneth already approved + CI is passing + All the unresolved conversations seem to be resolved) so that we can merge it 👍

PR #98, opened 2021-04-13
Fix the target_include_directories

### LINKED ISSUE #97: Compilation as a submodule is broken
https://github.com/browsermt/bergamot-translator/blob/3daa024eb303b52e483f326b26d6c976b42b2fe9/src/translator/CMakeLists.txt#L31

So, as far as I can tell, the syntax is wrong. There must be just one `PUBLIC`.

Second  `${CMAKE_SOURCE_DIR}` breaks when the project is included as a submodule. We should instead use `${PROJECT_SOURCE_DIR}` which will refer to the most recent `project` declaration.

XapaJIaMnu:
Fixes #97

### DIFF
--- a/src/translator/CMakeLists.txt
+++ b/src/translator/CMakeLists.txt
@@ -29,5 +29,5 @@ endif(COMPILE_WASM)
 target_link_libraries(bergamot-translator marian ssplit)
 
 target_include_directories(bergamot-translator
-    PUBLIC ${CMAKE_SOURCE_DIR}
-    PUBLIC ${CMAKE_SOURCE_DIR}/src)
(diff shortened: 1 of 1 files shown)


jerinphilip:
> So, as far as I can tell, the syntax is wrong. There must be just one PUBLIC.

Is this covered here? (Asking because automation will close that issue which also brings this up).

XapaJIaMnu:
> > So, as far as I can tell, the syntax is wrong. There must be just one PUBLIC.
> 
> Is this covered here? (Asking because automation will close that issue which also brings this up).

yes

jerinphilip (approved):

abhi-agg (approved):

PR #99, opened 2021-04-14
Clarify misleading comment

XapaJIaMnu:
The boolean passed to the `ExpressiongGraph` class actually sets the inference mode, it doesn't deal with any optimisations.


jerinphilip (approved):
/cc @ugermann 
https://github.com/ugermann/mts/blob/a9103e63b9a34304afaf8c572ded0a48ec3b7a44/src/service/common/translation_worker.cpp#L7

kpu (approved):

Issue #105, opened 2021-04-21
Does not compile on windows megathread


XapaJIaMnu:
Hi @kpu @jerinphilip,

Here I'll list the specific issues that prevent the codebase from compiling on Windows and discuss/propose/implement workarounds.

1) https://github.com/browsermt/bergamot-translator/blob/c00c263f8f8e1eb02ecbac3c59acdbe591f4fe0a/src/translator/multifactor_priority.h#L6 `<sys/time.h>` is a unix header, not available on win32. This doesn't seem to be used anywhere, should we delete the `multifactor_priority.*` files?
2) https://github.com/browsermt/bergamot-translator/blob/c00c263f8f8e1eb02ecbac3c59acdbe591f4fe0a/src/translator/sentence_ranges.cpp#L35 and a few other places in the file. MSVC compiler doesn't like syntax like:  `(ByteRange){START, END};` and triggers the `C4576: a parenthesized type followed by an initializer list is a non-standard explicit type conversion syntax`. This looks like a warning, is it safe to ignore? Can we use alternative syntax?
3) https://github.com/browsermt/bergamot-translator/blob/c00c263f8f8e1eb02ecbac3c59acdbe591f4fe0a/src/translator/response.h#L63 MSVC doesn't like this, not sure why. `C2280 'marian::bergamot::Response &marian::bergamot::Response::operator =(const marian::bergamot::Response &)': attempting to reference a deleted function`
4) https://github.com/browsermt/bergamot-translator/blob/c00c263f8f8e1eb02ecbac3c59acdbe591f4fe0a/src/translator/response.cpp#L68 similar to 2), including also line 78 in the same file.
5) https://github.com/browsermt/bergamot-translator/blob/c00c263f8f8e1eb02ecbac3c59acdbe591f4fe0a/src/translator/pcqueue.h#L20 this is not so much of an issue, as much as it is wishful thinking. Getting boost on windows is a bit annoying, is it possible to replace this with some lightweight header like: https://github.com/DanielTillett/Simple-Windows-Posix-Semaphore ? If not, i guess we should add this to the `cmake` dependencies because currently we don't look for boost.

jerinphilip:
1. That whole file can be deleted for now.
2. I'm using an old shorthand I know from C, it's just initing a strict. Can change it to something MSVC won't complain about.
[...]

PR #106, opened 2021-04-21
Windows PCQueue support without Boost

kpu:
Also undo formatting changes somebody made from my upstream version

PCQueue started out on https://github.com/kpu/preprocess/blob/master/util/pcqueue.hh and somebody changed the formatting from there.  I've undone that.  

Lightly tested in https://github.com/kpu/pcqueue/actions

Partially does #105.


XapaJIaMnu (approved):
Let's get this in for some heavier weight testing.

jerinphilip (approved):

PR #108, opened 2021-04-22
Windows workflows and mac framework accelerate draft

XapaJIaMnu:
This pull request updates the marian-dev submodule to the latest version of browsermt and restores the windows workflow.

The marian-dev changes allow to use the apple accelerate framework when running on apple machine and allows for compilation with apple clang. I also fixed they way `BLAS_FOUND` was being defined because it was preventing BLAS_FOUND to be set in some rare cases.

Finally, the windows workflow is fixed up, but will fail until some ssplit-cpp changes land. Those will go to a separate pull request, as they involve a few small changes to the CMakeLists.txt.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -39,8 +39,8 @@ jobs:
     - name: Prepare vcpkg
       uses: lukka/run-vcpkg@v4
       with:
-        vcpkgArguments: protobuf pcre2 #@TODO separate versions?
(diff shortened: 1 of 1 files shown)


jerinphilip (approved):
As I understand this PR updates marian-dev with apple fixes and brings the workflow file for windows with some cleanups into a state where it's improved.

Windows builds are expected to be fixed eventually when issues resolved. Approving the incremental changes here to go into main source.

kpu (approved):

PR #109, opened 2021-04-25
Full windows support with ssplit from browsermt, not a fork

XapaJIaMnu:
Support will materialise here, hopefully with the power of github workflows. This includes #108


XapaJIaMnu:
We're hitting these issue on mac:
```
wasm-ld: error: unknown argument: -search_paths_first
wasm-ld: error: unknown argument: -headerpad_max_install_names
```
https://github.com/emscripten-core/emscripten/issues/10896

Those are somehow being passed to the LDFLAGS when compiling pcre2 as external project. Fml.

XapaJIaMnu:
@kpu @jerinphilip @abhi-agg  this pull request introducs the following features:
1) On native marian mac, the default BLAS framework used is Apple Accelerate
2) ssplit-cpp uses a proper External project for pcre2, for those that want to no use the internal pcre2
3) Windows workflow + windows support. Several changes in marian were done in order to introduce it.
4) Proper MSVC compilation.

There are some issues still. There is a bug when using the Ninja generator on windows (https://github.com/lukka/run-cmake/issues/47 @ugermann  is debugging and opening bug reports left right and center until we figure out whose fault this is) , so as a workaround we use Visual Studio 2019 generator. The downside of this generator, is that it doesn't respect cmake variables such as `CMAKE_BUILD_TYPE` and those need to be provided by a supplementary CmakeSettings.json. I have not yet found a way to specify them at the command line, so if you're building outside the Visual Studio IDE, you'd default to a debug build. This will all be fixed when we move back to the ninja generator.

jerinphilip (dismissed):

jerinphilip (dismissed):
ssplit-cpp points to a sub-branch, which is okay for now given all the other improvements. The change of branch to main for submodule as I understand from an offline conversation will come separately.

Other than that, this looks good to me.

CMakeLists.txt
```diff
@@ -9,6 +9,28 @@ project(bergamot_translator CXX C)
...
 
+# Note that with CMake MSVC build, the option CMAKE_BUILD_TYPE is automatically derived from the key
+# 'configurationType' in CMakeSettings.json configurations
+if(NOT CMAKE_BUILD_TYPE)
+  message(WARNING "CMAKE_BUILD_TYPE not set; setting to Release")
+  set(CMAKE_BUILD_TYPE "Release")
+endif()
+#MSVC can't seem to pick up correct flags otherwise:
+if(MSVC)
+  add_definitions(-DUSE_SSE2=1) # Supposed to fix something in the sse_mathfun.h but not sure it does
```
abhi-agg, line 20: @XapaJIaMnu I believe, you don't need to set this definition here because marian submodule already sets this if the compiler is MSVC ([see this](https://github.com/browsermt/marian-dev/blob/master/CMakeLists.txt#L140))

src/QualityScore.h
```diff
@@ -8,6 +8,7 @@
 
 #include <string>
 #include <vector>
+#include "translator/definitions.h"
```
abhi-agg, line 11: This file can be removed [as I mentioned in my comment](https://github.com/browsermt/bergamot-translator/pull/109/files?file-filters%5B%5D=.h#r623949653).

CMakeLists.txt
XapaJIaMnu: Empirically, I think it does it if the generator in Ninja. In practise, this crashed every time for me when using the microsoft visual studio generator ;/. I think this would be fixed once this is fixed https://gitlab.kitware.com/cmake/cmake/-/issues/22133

src/translator/definitions.h
```diff
@@ -28,4 +28,21 @@ typedef AlignedVector<char> AlignedMemory;
 } // namespace bergamot
 } // namespace marian
 
+// @TODO at the moment the usage of string_view in this repository is a hot mess and a disaster waiting to happen.
+// ssplit uses std::string_view if the compiler supports c++17, else falls back to c++11 and absl::string_view
+// bergamot-translator uses, depending on the source file std::string_view (which will break if ssplit-cpp uses
+// absl::string_view) and marian::string_view which is an export of (confusingly) the sentencepiece module that
+// marian has. marian::string_view is our addition to the marian fork, which will make merging even nicer. Not.
+// This is just an ugly patchwork that allos gcc5, our lowest targetted gcc to run. We don't actually try to run
+// on older compilers.
```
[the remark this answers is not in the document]
jerinphilip: So, I was convinced what Nick is doing here is the minimum invasive edit, restricting to what this PR describes it is doing. We can delete the files in another one.

ssplit-cpp I think still maintains an `std::string_view` dependency, far as I understand, although it can change.

CMakeLists.txt
abhi-agg: Okay.

EDIT: It seems to me that `marian-dev` submodule in this PR points to some other than `master` (please correct me if I am wrong). Is it intended?

XapaJIaMnu: Hmmm, no... This is probably my bad, I will fix it before merging.

src/QualityScore.h
XapaJIaMnu: I'll leave removing this file to you in the subsequent pull request? Once you merge it you can remove the unnecessary files.

PR #115, opened 2021-04-27
Faithful to source-structure translation

### LINKED ISSUE #103: Preserve whitespace between sentences
Currently all whitespace between source sentences is normalized to space.  We should be preserving the source whitespace in the target.  I think this was requested by @motin but I can't find the original issue.  Anyway I support it.

jerinphilip:
Fixes #103. Ready for review.

Target translated text is now constructed by inserting translated sentences where source-sentences existed before, effectively replicating the structure (paragraph, DOM tags, whatever) from source. If someone provides string_views to extract sentences (here, sentence-splitter from ssplit-cpp), the string-view rebase math is worked out to extract the filler text in between sentences, copying them over to translated text.


jerinphilip:
Looks reasonable, for now.

```
jphilip@var:~/code/bergamot-translator-test-scripts$ cat paragraph.in  | /mnt/Storage/jphilip/bergamot-build/app/service-cli -m /home/jphilip/code/bergamot-translator-test-scripts/models/deen/e
nde.student.tiny11/model.intgemm.alphas.bin --vocabs /home/jphilip/code/bergamot-translator-test-scripts/models/deen/ende.student.tiny11/vocab.deen.spm /home/jphilip/code/bergamot-translator-te
st-scripts/models/deen/ende.student.tiny11/vocab.deen.spm --ssplit-mode paragraph --alignment soft --beam-size 1 --skip-cost --shortlist /home/jphilip/code/bergamot-translator-test-scripts/mode
ls/deen/ende.student.tiny11/lex.s2t 50 50 --int8shiftAlphaAll --cpu-threads 4 --max-length-break 1024 --mini-batch-words 1024 -w 128 2> /dev/null
[original]: The Project Gutenberg EBook of Crime and Punishment, by Fyodor Dostoevsky

This eBook is for the use of anyone anywhere at no cost and with
almost no restrictions whatsoever.  You may copy it, give it away or
re-use it under the terms of the Project Gutenberg License included
with this eBook or online at www.gutenberg.org


[translated]: Das Projekt Gutenberg EBook of Crime and Punishment, von Fjodor Dostojewski

Dieses eBook ist für die Verwendung von jedermann überall ohne Kosten und mit
fast gar keine Einschränkungen.  Sie können es kopieren, verschenken oder
wiederverwenden Sie es unter den Bedingungen des Projekts Gutenberg Lizenz enthalten
mit diesem eBook oder online unter www.gutenberg.org
```

XapaJIaMnu:
Ubuntu tests fail..?

jerinphilip:
@XapaJIaMnu I'm editing how the outputs are generated so they're expected to? I need to find out why Mac is failing though.

The solution here is to update output (or input to remove trailing newline/space). 😄

XapaJIaMnu:
OK so the tests had paragraphs in them. What else do you want to do in here that it is marked as wip

kpu:
I pulled this submodule into the demo I'm giving of translateLocally.

jerinphilip:
I have manually updated all outputs to have a trailing `\n` at https://github.com/browsermt/bergamot-translator-tests/pull/24. The submodule is updated here and the tests are passing except for windows.

src/translator/sentence_ranges.cpp
```diff
@@ -99,5 +99,42 @@ string_view AnnotatedText::asStringView(const ByteRange &byteRange) const {
   return string_view(data, size);
 }
 
+string_view AnnotatedText::pre(size_t sentenceIdx) const {
+  // Find start of filler-text before, there's a corner case when there's no
+  // sentence before.
+  const char *start{nullptr};
```
XapaJIaMnu, line 105: the C++ style is usually = nullptr;

src/translator/sentence_ranges.cpp
```diff
@@ -99,5 +99,42 @@ string_view AnnotatedText::asStringView(const ByteRange &byteRange) const {
...
+  if (sentenceIdx == 0) {
+    // If first sentence, filler begins at start of whole-text.
+    start = &(text[0]);
```
XapaJIaMnu, line 108: unnecessary brackets here

jerinphilip: I'm worried this might lead to segfaults. [0], is there a safe way to avoid the indexing and get a char*?

XapaJIaMnu (approved):
SOme minor comments, but other than that it looks good. I didn't look too closely at the code. 

Please change the C style initialisation because sometimes MSVC complains about it. It's better to have the pull request that enables the windows workflow merged first so that we can see when windows compilation is broken.

src/translator/sentence_ranges.cpp
XapaJIaMnu: `const char * start = nullptr`

XapaJIaMnu: Oh sorry wrong comment. The [] takes precedence before &: https://en.cppreference.com/w/cpp/language/operator_precedence

src/translator/sentence_ranges.cpp
kpu: Why are we even initializing if both branches of if initialize it?

src/translator/sentence_ranges.cpp
kpu: text.data()

src/translator/sentence_ranges.cpp
jerinphilip: Removed brackets initialization altogether.

src/translator/sentence_ranges.cpp
jerinphilip: x.data() is used inside the container now.

src/translator/sentence_ranges.cpp
```diff
@@ -99,5 +99,42 @@ string_view AnnotatedText::asStringView(const ByteRange &byteRange) const {
   return string_view(data, size);
 }
 
+string_view AnnotatedText::pre(size_t sentenceIdx) const {
+  // Find start of filler-text before, there's a corner case when there's no
+  // sentence before.
+  const char *start;
```
XapaJIaMnu, line 105: Very minor comment, it's good practise to assign a ptr to `nullptr` when you declare it, but you don't actually assign its value right away.

XapaJIaMnu (approved):
Ready for merge I think

abhi-agg:
> I have manually updated all outputs to have a trailing `\n` at [browsermt/bergamot-translator-tests#24](https://github.com/browsermt/bergamot-translator-tests/pull/24).

@jerinphilip Do we have a test for the faithful translation in [bergamot-translator-tests](https://github.com/browsermt/bergamot-translator-tests)? I was going through [browsermt/bergamot-translator-tests#24](https://github.com/browsermt/bergamot-translator-tests/pull/24) and I couldn't find any. I may be wrong though.

If not, you can just change the existing tests to introduce some extra whitespaces b/w sentences of a paragraph just like you showed in https://github.com/browsermt/bergamot-translator/pull/115#issuecomment-827750573 👍

src/translator/sentence_ranges.h
```diff
@@ -151,6 +151,14 @@ struct AnnotatedText {
   /// Returns a string_view representing sentence corresponding to sentenceIdx.
   string_view sentence(size_t sentenceIdx) const;
 
+  /// Returns the string_view of the filler between the previous sentence and
+  /// sentence described by sentenceIdx
+  string_view pre(size_t sentenceIdx) const;
+
+  /// Returns the string_view of the filler after the sentence corresponding to
+  /// sentenceIdx and the next sentence.
+  string_view post(size_t sentenceIdx) const;
```
abhi-agg, line 160: @jerinphilip Do you need both `post` and `pre` methods? Based on my understanding, `pre(n) = post(n-1)`. Right? Or did I miss something?

jerinphilip: Any alternative ideas, 

1. I can have an API with `.filler(i)`, where `i = -1..N-1 or 0..N+1`, while the sentences range from `[0, N-1]`. Use this to update the algorithm. I don't know if it is sane to have -1 or N+1 and create confusion while reading code. To me I prefer having `pre(i)` where I know it's filler before and `post(i)` where I know it's filler after while reading outside, not having to look up documentation.
2. I can return a pair of `(pre, post)` as `surroundings(i)` for the `i`-th sentence. Whoever calls the function outside decides which one to use to build their algorithm (faithful translation here). This is neater for readability, imo if you want a single function.

kpu: I think 0 to N+1 is perfectly sensible here.

jerinphilip: Changed with `gap(i)`.

jerinphilip:
> If not, you can just change the existing tests to introduce some extra whitespaces b/w sentences of a paragraph just like you showed in

@abhi-agg Went for this, checkout https://github.com/browsermt/bergamot-translator-tests/pull/24 now.

src/translator/sentence_ranges.cpp
jerinphilip: Set back to nullptr.

abhi-agg (dismissed):

jerinphilip:
Rerequest of approvals @abhi-agg, @XapaJIaMnu? Apologies for the extra trouble - the conflict plus sync to submodule dismissed reviews.

abhi-agg (approved):

XapaJIaMnu (approved):

Issue #112, opened 2021-04-27
Parsing config string into Options invokes config validator for byte array based constructor


abhi-agg:
Currently, the [constructor of the Service class](https://github.com/browsermt/bergamot-translator/blob/main/src/translator/service.h#L54-L55) invokes [parseOptions](https://github.com/browsermt/bergamot-translator/blob/main/src/translator/service.h#L56).

The [parseOptions](https://github.com/browsermt/bergamot-translator/blob/main/src/translator/parser.h#L34-L65) function performs 2 responsibilities right now:

1. It parses the `config` string to `marian::Options` object (which it should do because Marian works with this object internally almost everywhere)
2. And runs the [ConfigValidator](https://github.com/browsermt/bergamot-translator/blob/main/src/translator/parser.h#L61-L62) on the parsed `marian::Options` object to check its validity

My question is pertaining to its 2nd responsibility: Should the [ConfigValidator](https://github.com/browsermt/bergamot-translator/blob/main/src/translator/parser.h#L61-L62) step be executed at all while [constructing Service](https://github.com/browsermt/bergamot-translator/blob/main/src/translator/service.h#L54-L55) with model, shortlist (and vocab in future) files as bytes?

I believe we don't need the validation step for this constructor because [ConfigValidator for translation mode](https://github.com/browsermt/marian-dev/blob/master/src/common/config_validator.cpp#L49-L66) checks:
1. Whether model file and vocab files are provided and
2. The path of these files are valid.

We don't need these checks if we are providing these files as bytes.

**My proposal is**: Since the [constructor](https://github.com/browsermt/bergamot-translator/blob/main/src/translator/service.h#L54-L55) is designed for the cases when model, shortlist file (and vocab in future) can be provided as byte array, we can remove the validator step from it (of course, it would mean [no optional arguments](https://github.com/browsermt/bergamot-translator/blob/main/src/translator/service.h#L55) in this constructor but I can make that change in the PR for the issue https://github.com/browsermt/bergamot-translator/issues/75).

@kpu @XapaJIaMnu @jerinphilip I can provide this fix if the proposal sounds good to you. Please let me know if I missed something here 👍 
cc @andrenatal

[...]

PR #116, opened 2021-04-28
Control validating the config options via a boolean flag

abhi-agg:
- parseOptions() function now validates the parsed options
   based on the validate argument
 
 The PR fixes https://github.com/browsermt/bergamot-translator/issues/112

### DIFF
--- a/src/translator/parser.h
+++ b/src/translator/parser.h
@@ -31,7 +31,7 @@ inline marian::ConfigParser createConfigParser() {
 }
 
 inline std::shared_ptr<marian::Options>
-parseOptions(const std::string &config) {
+parseOptions(const std::string &config, bool validate = true) {
   marian::Options options;
 
   // @TODO(jerinphilip) There's something off here, @XapaJIaMnu suggests
@@ -58,8 +58,11 @@ parseOptions(const std::string &config) {
   options.parse(config);
   YAML::Node configCopy = options.cloneToYamlNode();
 
-  marian::ConfigValidator validator(configCopy);
-  validator.validateOptions(marian::cli::mode::translation);
+  if (validate) {
+    // Perform validation on parsed options only when requested
+    marian::ConfigValidator validator(configCopy);
+    validator.validateOptions(marian::cli::mode::translation);
+  }
 
   return std::make_shared<marian::Options>(options);
 }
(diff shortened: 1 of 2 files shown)


src/translator/service.h
```diff
@@ -84,7 +84,7 @@ class Service {
   explicit Service(const std::string &config,
                    AlignedMemory modelMemory = AlignedMemory(),
                    AlignedMemory shortlistMemory = AlignedMemory())
-      : Service(parseOptions(config), std::move(modelMemory),
+      : Service(parseOptions(config, false), std::move(modelMemory),
```
jerinphilip, line 87: `/*validate=*/false`, since it's not obvious from reading here, if that's okay for you. ([ref](https://google.github.io/styleguide/cppguide.html#Function_Argument_Comments))

abhi-agg: Done.

jerinphilip (approved):

src/translator/parser.h
```diff
@@ -31,7 +31,7 @@ inline marian::ConfigParser createConfigParser() {
 }
 
 inline std::shared_ptr<marian::Options>
-parseOptions(const std::string &config) {
+parseOptions(const std::string &config, bool validate = true) {
```
XapaJIaMnu, line 34: This default parameter here is true (as expected)

src/translator/service.h
```diff
@@ -84,7 +84,7 @@ class Service {
   explicit Service(const std::string &config,
                    AlignedMemory modelMemory = AlignedMemory(),
                    AlignedMemory shortlistMemory = AlignedMemory())
-      : Service(parseOptions(config), std::move(modelMemory),
+      : Service(parseOptions(config, /*validate=*/false), std::move(modelMemory),
```
XapaJIaMnu, line 87: The comment here suggests the default is false when it's true.

XapaJIaMnu (changes requested):
Consistent comment and code behaviour

src/translator/service.h
abhi-agg: @XapaJIaMnu Do you want me to remove `/*validate=*/` from here?

XapaJIaMnu:
Sorry my bad didn't see the comment doesn't extend to the value

Issue #75, opened 2021-03-26
Adapt JS bindings for WASM module to pass files as byte array from browser extension


jerinphilip:
Current bindings only support the single `std::string` config constructor. The adjustment which succeeds at least in builds in the WASM - `TranslationModel` boundary needn't wait until the feature is developed in marian end to speed things up. Ideally, this part should be testable using stubs so that we know that if something fails we are to look in marian or the WASM boundary.

https://github.com/mozilla/bergamot-translator/issues/8 complains about `std::string_view` exports (which I hope is just absence of first class support) while conceptually a `string_view` is only `const void* + size_t`. If you need to test with the actual pointer, @XapaJIaMnu has already gotten it till WASM-TranslationModel boundary (https://github.com/browsermt/bergamot-translator/pull/55/commits/7da6392013cfc69bf87b5aa51da93a5daa995fcb) with the model in current main.

https://github.com/browsermt/bergamot-translator/blob/e0dca1ba1b5d345a1c81a8932addd711f37b51c7/wasm/bindings/TranslationModelBindings.cpp#L14-L22

> Refactoring const void *model_memory, size_t model_memory_size with a MemoryGift struct or something can be done later once I generate JS bindings.

Please prioritize this and bring this asap so #69 gets timely feedback.

EDIT: Files in question here are model, vocabulary and shortlist files.

XapaJIaMnu:
So what i'm thinking is having something like
```c++
class marian_model {
   explicit marian_model(void * model_bytes, size_t model_bytes_size, void * shortlist, size_t shortlist_size,  void * vocab, size_t vocab size) {
                 validate_model_integrity();
                 validate_vocab_integrity();
                 validate_shortlist_integrity();
                 }
    marian_model() = delete;
```
Maybe include the config to it as well, and do a sanity-check that all the files are correct. Or use checksums for validation..

kpu:
I think you meant const void *

abhi-agg:
@jerinphilip There is another issue in this repo https://github.com/browsermt/bergamot-translator/issues/59 that you created and sounds similar to me. I am closing that one.

abhi-agg:
- [x] Write bindings for AlignedMemory class
- [x] Adapt bindings of TranslationModel class to use AlignedMemory for passing model and shortlist bytes
- [x] Modify wasm test page to:
    - [x] Load both model and shortlist file from local machine (instead of emscripten’s emulated file system)
[...]

PR #117, opened 2021-04-28
JS bindings for loading model and shortlist files as bytes

abhi-agg:
The PR fixes https://github.com/browsermt/bergamot-translator/issues/75

1. Adds JS bindings for `AlignedMemory` class and modifies bindings of `TranslationModel` class to use `AlignedMemory`
2. Modifies wasm test page to use the new bindings
3. Updated README with appropriate instructions

### DIFF
--- a/wasm/bindings/TranslationModelBindings.cpp
+++ b/wasm/bindings/TranslationModelBindings.cpp
@@ -9,76 +9,51 @@
 #include "TranslationModel.h"
 
 using namespace emscripten;
-/*
-std::unique_ptr<std::size_t> createSizeTAlignedMemory() {
-  std::cout << "createSizeTAlignedMemory" << std::endl;
-  return std::unique_ptr<std::size_t>(new std::size_t(9));
-}
-
-EMSCRIPTEN_BINDINGS(aligned_sizet_memory) {
-  function("createSizeTAlignedMemory", &createSizeTAlignedMemory);
-}
-
-AlignedUint8Memory* createAlignedUint8Memory(std::size_t size, std::size_t alignment) {
-  std::cout << "unique ptr createAlignedUint8Memory" << std::endl;
-  return new AlignedUint8Memory(size, alignment);
-}
-*/
 
 val getBytes(marian::bergamot::AlignedMemory& alignedMemory) {
-  return val(typed_memory_view(alignedMemory.size(), alignedMemory.as<unsigned char>()));
+  return val(typed_memory_view(alignedMemory.size(), alignedMemory.as<char>()));
 }
 
 void print(marian::bergamot::AlignedMemory& alignedMemory) {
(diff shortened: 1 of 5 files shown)


andrenatal (dismissed):

wasm/bindings/TranslationModelBindings.cpp
```diff
@@ -10,10 +10,30 @@
...
+    .function("size", &marian::bergamot::AlignedMemory::size)
+	  .function("getByteArrayView", &getByteArrayView)
+    ;
+}
+
+TranslationModel* TranslationModelFactory(const std::string &config,
+                                          marian::bergamot::AlignedMemory* modelMemory,
+                                          marian::bergamot::AlignedMemory* shortlistMemory) {
+  if (modelMemory == nullptr || shortlistMemory == nullptr) {
+    return nullptr;
```
XapaJIaMnu, line 29: Do you ever check for that nullptr down the line? You should have a graceful failure

wasm/test_page/bergamot.html
```diff
@@ -93,22 +111,53 @@
...
+      log(`${languagePair} file download took ${(Date.now() - start) / 1000} secs`);
+
+      // Construct AlignedMemory objects with downloaded buffers
+      var alignedModelMemory = constructAlignedMemoryFromBuffer(modelBuffer, 256);
+      var alignedShortlistMemory = constructAlignedMemoryFromBuffer(shortListBuffer, 64);
+
+      // Instantiate the TranslationModel
+      if (translationModel) translationModel.delete();
+      console.debug("Creating TranslationModel with config:", modelConfigWithoutModelAndShortList);
+      translationModel = new Module.TranslationModel(modelConfigWithoutModelAndShortList, alignedModelMemory, alignedShortlistMemory);
```
XapaJIaMnu, line 157: This could be null here

XapaJIaMnu (changes requested):
Gracefully fail when model construction fails.

wasm/bindings/TranslationModelBindings.cpp
abhi-agg: We don't expect these to be nullptr. I kept it just to be cautious. I will remove it from here. 👍

wasm/test_page/bergamot.html
abhi-agg: The object returned will never be null once I fix the comment above.

XapaJIaMnu:
I think it's good to have a check for failure. I'm not sure what happens in your pipeline when the download fails. I think I saw some code that covers this case

abhi-agg:
> I think it's good to have a check for failure. I'm not sure what happens in your pipeline when the download fails. I think I saw some code that covers this case

You are right about both handling the download failure and a check in place in the test page for this 👍

XapaJIaMnu (approved):

motin (approved):
Approving without testing the changes (will be done once it is possible).

PR #118, opened 2021-04-28
Updating brt-tests to rely on linux less

jerinphilip:
WIP: Mac tests are skipping, let's fix that one first.


jerinphilip:
Macs steps are no longer skipping, they're differing in outputs for minimal and full configurations. I have got one to work for now, but the other one is failing. Both resolve to `ssse3` on lookup of CPU features to get respective output.

The difference between the two builds are only
`(USE_STATIC_LIBS, WASM_COMPATIBLE_SOURCE) = (on, on) | (off, off)` looks like. 

Any idea what might be happening @XapaJIaMnu, @abhi-agg?

XapaJIaMnu:
The mac build uses openblas or the apple accelerate framework, not mkl. It is expected that scores are slightly different.

As for the wasm compatible sources build, compilers on different systems can produce slightly different fp32 results.

jerinphilip:
Abandoned, being worked on in 127 though.

PR #121, opened 2021-04-29
Functional wasm test page

abhi-agg:
This PR enables using wasm test page for inference by passing models and shortlist as bytes (instead of files) to the translation engine

### DIFF
--- a/wasm/test_page/bergamot.html
+++ b/wasm/test_page/bergamot.html
@@ -65,7 +65,7 @@ En consecuencia, durante el año 2011 se introdujeron 180 proyectos de ley que r
   const downloadAsArrayBuffer = async(url) => {
     const response = await fetch(url);
     if (!response.ok) {
-      throw Error(`HTTP ${response.status} - ${response.statusText}`);
+      throw Error(`Downloading ${url} failed: HTTP ${response.status} - ${response.statusText}`);
     }
     return response.arrayBuffer();
   }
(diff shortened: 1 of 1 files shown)


wasm/test_page/bergamot.html
```diff
@@ -156,7 +156,7 @@
       console.debug("Creating TranslationModel with config:", modelConfigWithoutModelAndShortList);
       translationModel = new Module.TranslationModel(modelConfigWithoutModelAndShortList, alignedModelMemory, alignedShortlistMemory);
     } catch (error) {
-      console.error(error);
+      log(error);
```
jerinphilip, line 159: Not console.log?

abhi-agg: Nopes. Otherwise it log the failure silently and you will see nothing on the actual page 👍

jerinphilip:
Question: Can I use this to test the updated WASM bindings code in #87?

jerinphilip (approved):

abhi-agg:
> Question: Can I use this to test the updated WASM bindings code in #87?

Most likely yes (please follow [this step of README](https://github.com/browsermt/bergamot-translator/tree/main/wasm#-demo-see-everything-in-action)) :)

If it doesn't then I can help you there 👍

XapaJIaMnu (approved):

PR #122, opened 2021-04-29
Enable vocabs pass as byte arrays

qianqianzhu:
Enable loading SentencePiece vocabs from the buffer.
- Update `marian-dev` submodule to add `loadFromSerialized()` in vocabulary side. 
- Plumb vocabs byte arrays into `Service`.


qianqianzhu:
The vocab bytes is slightly different from the model and the shortlist, as it is used to initialise `vocabs_` before the `Service` object (it seems not needed for service afterwards).  Also `LoadFromSerializedProto` from the SentencePiece module needs to wrap the actual vocab bytes with `string_view`. 
So I am not sure about the ownership of the vocab bytes: should `service` own the actual vocab bytes? Or is it better to pass the `string_view`  pointers to service (my current solution)?
@kpu @jerinphilip any ideas?

src/translator/byte_array_util.h
```diff
@@ -7,6 +7,9 @@ namespace bergamot {
 AlignedMemory loadFileToMemory(const std::string& path, size_t alignment);
 AlignedMemory getModelMemoryFromConfig(marian::Ptr<marian::Options> options);
 AlignedMemory getShortlistMemoryFromConfig(marian::Ptr<marian::Options> options);
+void getVocabsMemoryFromConfig(marian::Ptr<marian::Options> options,
+                               std::vector<AlignedMemory>& vocabsMemory,
+                               std::vector<string_view>& vocabsMemorySV);
```
jerinphilip, line 12: This should be `AlignedMemory`, not `string_view`. Reasons 1) Consistency, API person outside shouldn't work with 2 structures doing same thing. 2) `AlignedMemory` + WASM glue has been worked out already over [here](https://github.com/browsermt/bergamot-translator/blob/2788116f8b10f7edd2168ae7fb0be0f059681c8b/wasm/bindings/TranslationModelBindings.cpp#L13-L23).

/cc @abhi-agg

src/translator/service.cpp
```diff
@@ -6,33 +6,46 @@
...
   return vocabs;
 }
 
 namespace marian {
 namespace bergamot {
 
-Service::Service(Ptr<Options> options, AlignedMemory modelMemory, AlignedMemory shortlistMemory)
-    : requestId_(0), vocabs_(std::move(loadVocabularies(options))),
+Service::Service(Ptr<Options> options, AlignedMemory modelMemory, AlignedMemory shortlistMemory,
+                 const std::vector<marian::string_view> *vocabsMemoryView)
```
jerinphilip, line 43: This should be `AlignedMemory`, not `string_view`. Reasons 1) Consistency, API person outside shouldn't work with 2 structures doing same thing. 2) `AlignedMemory` + WASM glue has been worked out already over [here](https://github.com/browsermt/bergamot-translator/blob/2788116f8b10f7edd2168ae7fb0be0f059681c8b/wasm/bindings/TranslationModelBindings.cpp#L13-L23).

/cc @abhi-agg

src/translator/service.cpp
```diff
@@ -6,33 +6,46 @@
...
-    if (m.second) { // new: load the vocab
-      m.first->second = marian::New<marian::Vocab>(options, i);
-      m.first->second->load(vfiles[i]);
+  std::vector<marian::Ptr<marian::Vocab const>> vocabs;
+  if(vocabsMemoryView != nullptr){
+    // load vocabs from buffer
+    vocabs.resize(vocabsMemoryView->size());
+    for (size_t i = 0; i < vocabs.size(); i++) {
+      marian::Ptr<marian::Vocab> vocab = marian::New<marian::Vocab>(options, i);
+      vocab->loadFromSerialized((*vocabsMemoryView)[i]);
```
jerinphilip, line 17: convert AlignedMemory -> string_view here somehow? You won't have to copy the underlying memory I'm guessing? Restrict this conversion troubles to minimum interaction with sentencepiece (I'm guessing method signature there is what leads to troubles). string_view will be alive while AlignedMemory is alive.

As with others, we accept `AlignedMemory` and `Service` owns `AlignedMemory` (for vocab).

jerinphilip:
@qianqianzhu I've quickly responded with the minimal understanding I have working with spiece and the rest. Also take a look at the comments left on [`bc42a37` (#109)](https://github.com/browsermt/bergamot-translator/pull/109/commits/bc42a372da6571882225e7c40dbc66b2cd184b79) regarding `string_view`s. 

I see that the buffers aren't required after vocabs are loaded, so feel free to not `std::move(..)` it to a member variable (of Service), so the memory passed will be freed after use in construction.

kpu:
The memory is moved into Service.  So Service can clear it after initializing vocabulary.

kpu:
And agree it may as well be AlignedMemory for consistency.

src/translator/service.cpp
abhi-agg: I agree on this.

src/translator/service.h
```diff
@@ -83,9 +84,11 @@ class Service {
   /// @param [in] shortlistMemory byte array of shortlist (aligned to 64)
   explicit Service(const std::string &config,
                    AlignedMemory modelMemory = AlignedMemory(),
-                   AlignedMemory shortlistMemory = AlignedMemory())
+
+                   AlignedMemory shortlistMemory = AlignedMemory(),
+                   const std::vector<string_view> *vocabsMemoryView = nullptr)
```
abhi-agg, line 89: Can it be `const std::vector<AlignedMemory> vocabMemory`? It would be consistent with rest of the 2 arguments 👍

src/translator/service.cpp
qianqianzhu: Many thanks for your feedback. I changed `string_view` to `AlignedMemory`.

src/translator/service.h
qianqianzhu: Yes. This is fixed.

src/translator/service.cpp
```diff
@@ -6,33 +6,48 @@
...
 
 namespace marian {
 namespace bergamot {
 
-Service::Service(Ptr<Options> options, AlignedMemory modelMemory, AlignedMemory shortlistMemory)
-    : requestId_(0), vocabs_(std::move(loadVocabularies(options))),
+Service::Service(Ptr<Options> options, AlignedMemory modelMemory, AlignedMemory shortlistMemory,
+                 std::vector<AlignedMemory> vocabMemorySet)
+    : requestId_(0),
+      vocabMemorySet_(std::move(vocabMemorySet)),
```
kpu, line 45: Why keep it at all?  Just forward it to vocabs_ and then free the memory once loaded.

src/translator/service.h
```diff
@@ -73,7 +74,7 @@ class Service {
   /// wherein empty memory is passed and internal flow defaults to file-based
   /// model, shortlist loading.
   explicit Service(Ptr<Options> options)
-      : Service(options, AlignedMemory(), AlignedMemory()) {}
+      : Service(options, AlignedMemory(), AlignedMemory(), {}) {}
```
jerinphilip, line 77: I agree with @kpu now, default constructors are okay - get rid of this double tax; Probably a note in docs/comments in the other constructor that `AlignedMemory()` corresponds to empty memory and leads to loading from files supplied through config. 

This is my bad, sorry. 🤦‍♂️

src/translator/service.h
```diff
@@ -83,9 +84,10 @@ class Service {
   /// @param [in] shortlistMemory byte array of shortlist (aligned to 64)
```
jerinphilip, line 84: Missing documentation for the new param.

src/translator/byte_array_util.h
```diff
@@ -7,6 +7,8 @@ namespace bergamot {
 AlignedMemory loadFileToMemory(const std::string& path, size_t alignment);
 AlignedMemory getModelMemoryFromConfig(marian::Ptr<marian::Options> options);
 AlignedMemory getShortlistMemoryFromConfig(marian::Ptr<marian::Options> options);
+void getVocabsMemoryFromConfig(marian::Ptr<marian::Options> options,
+                               std::vector<AlignedMemory>& vocabMemorySet);
```
jerinphilip, line 11: @qianqianzhu Could you `s/vocabMemorySet/vocabMemories` or something so we don't confuse it to be an `std::set` from variable name, if it's not too much?

src/translator/service.cpp
```diff
@@ -6,33 +6,48 @@
 #include <utility>
 
 inline std::vector<marian::Ptr<const marian::Vocab>>
-loadVocabularies(marian::Ptr<marian::Options> options) {
+loadVocabularies(marian::Ptr<marian::Options> options, const std::vector<marian::bergamot::AlignedMemory>& vocabMemorySet) {
   // @TODO: parallelize vocab loading for faster startup
-  auto vfiles = options->get<std::vector<std::string>>("vocabs");
-  // with the current setup, we need at least two vocabs: src and trg
-  ABORT_IF(vfiles.size() < 2, "Insufficient number of vocabularies.");
```
jerinphilip, line 13: Do we have this check somewhere in the aligned memory case? Should be useful, I hope?

src/translator/byte_array_util.h
kpu: I think she is using the word set correctly in the sense that it only has the set of vocabularies i.e. the same vocab for source and target is only loaded once.

src/translator/byte_array_util.cpp
```diff
@@ -104,5 +104,15 @@ AlignedMemory getShortlistMemoryFromConfig(marian::Ptr<marian::Options> options)
...
 }
 
+void getVocabsMemoryFromConfig(marian::Ptr<marian::Options> options,
+                               std::vector<AlignedMemory>& vocabMemorySet){
+  auto vfiles = options->get<std::vector<std::string>>("vocabs");
+  ABORT_IF(vfiles.size() < 2, "Insufficient number of vocabularies.");
+  vocabMemorySet.resize(vfiles.size());
+  for (size_t i = 0; i < vfiles.size(); ++i) {
+    vocabMemorySet[i] = loadFileToMemory(vfiles[i], 64);
+  }
```
jerinphilip, line 114: @kpu This is not a set, @qianqianzhu is double loading it even if the files are same.

src/translator/byte_array_util.cpp
```diff
@@ -104,5 +104,15 @@ AlignedMemory getShortlistMemoryFromConfig(marian::Ptr<marian::Options> options)
   return loadFileToMemory(shortlist[0], 64);
 }
 
+void getVocabsMemoryFromConfig(marian::Ptr<marian::Options> options,
+                               std::vector<AlignedMemory>& vocabMemorySet){
+  auto vfiles = options->get<std::vector<std::string>>("vocabs");
+  ABORT_IF(vfiles.size() < 2, "Insufficient number of vocabularies.");
```
jerinphilip, line 110: Check should be in loadVocabularies (if mozilla or an API consumer passes only one vocab, we should abort with readable error).

src/translator/byte_array_util.cpp
qianqianzhu: Here is my current struggle. I tried to avoid duplicate loading but failed. As here we need to provide the number of vocabularies. If you have any ideas, please tell me.

jerinphilip: This is tricky territory. 

I think we should take from the callee (Mozilla) a `vector<AlignedMemory> vocabMemories` of unique vocabulary memories and a `vector<size_t> vocabIndices` indicating the index where we load (each vocabs) to not double copy the memory (also not double load).  

So if we're reusing source and target vocabularies as `esen`, we'll have `vocabMemories.size() = 1`, and `vocabIndices.size() = 2`.

With this we can build a replicate the current hashmap behaviour with filenames on `vocabIndices` avoiding duplication between loading from files/memory as well I think.

/cc @abhi-agg

qianqianzhu: @jerinphilip many thanks for your suggestion. (1) Introducing `vector<size_t> vocabIndices` is indeed a good way to avoid double loading.

Following your thoughts, if we will have a pair of vocabularies (source and target), we can just deal with `vocabMemories.size() = 1` and `vocabMemories.size() = 2` separately. (2) When `vocabMemories.size() = 1`, we initialise source and target vocabs from the same buffer. When `size = 2`, we initialise source and target vocabs from different buffers.

Or we can (3) use `std::vector<marian::Ptr<AlignedMemory>> vocabMemories` to avoid double loading. I hope introducing `shared_ptr` will not affect the WASM JS binding (@abhi-agg). 

@kpu @jerinphilip @abhi-agg Let me know which way works best.

XapaJIaMnu:
@qianqianzhu please fix the windows build.

src/translator/service.h
```diff
@@ -60,7 +60,7 @@ class Service {
   /// @param shortlistMemory byte array of shortlist (aligned to 64)
```
abhi-agg, line 68: The line above in documentation says `modelMemory` is Optional. Is it correct?

qianqianzhu: This line is outdated. For this constructor, `modelMemory` is compulsory. I updated the documentation in the newest documentation.

src/translator/byte_array_util.cpp
```diff
@@ -89,13 +89,17 @@ AlignedMemory loadFileToMemory(const std::string& path, size_t alignment){
...
+  ABORT_IF(models.size() != 1, "Loading multiple binary models is not supported for now as it is not necessary.");
+  marian::filesystem::Path modelPath(models[0]);
+  ABORT_IF(modelPath.extension() != marian::filesystem::Path(".bin"), "The file of binary model should end with .bin");
+  AlignedMemory alignedMemory = loadFileToMemory(models[0], 256);
+  bool check = options->get<bool>("check-bytearray",true);
+  if (check) {
+    ABORT_IF(!validateBinaryModel(alignedMemory, alignedMemory.size()),
+             "The binary file is invalid. Incomplete or corrupted download?");
+  }
+  return alignedMemory;
```
jerinphilip, line 102: This path is not fired when mozilla calls through API, we want to give them the ability to toggle check-bytearray as well, yes? Can't we have the check/validate code in one place, i.e inside Service consistent with where shortlist is being checked as well?

This function is so far only used in service-cli right?

qianqianzhu: OK. I will move to Service.

src/translator/service.cpp
```diff
@@ -6,33 +6,51 @@
...
+    ABORT_IF(vfiles.size() < 2, "Insufficient number of vocabularies.");
+    vocabs.resize(vfiles.size());
+    std::unordered_map<std::string, marian::Ptr<marian::Vocab>> vmap;
+    for (size_t i = 0; i < vocabs.size(); ++i) {
+      auto m = vmap.emplace(std::make_pair(vfiles[i], marian::Ptr<marian::Vocab>()));
+      if (m.second) { // new: load the vocab
+        m.first->second = marian::New<marian::Vocab>(options, i);
+        m.first->second->load(vfiles[i]);
+      }
+      vocabs[i] = m.first->second;
```
jerinphilip, line 37: Abstract the `load*` function somehow and eliminate the redundancies? The load method called is the only difference between the two, correct? Preferable or overengineering?

src/translator/byte_array_util.cpp
abhi-agg: I agree on this one.

src/translator/service.cpp
qianqianzhu: I will have a try here to eliminate the redundancies.

src/translator/byte_array_util.cpp
```diff
@@ -5,8 +5,6 @@
 namespace marian {
 namespace bergamot {
 
-namespace {
```
jerinphilip, line 8: What happened here? I thought @XapaJIaMnu was told to put externally unrequired functions / struct (Header, some other stuff like `get`) in an anonymous namespace.

jerinphilip: Context: https://github.com/browsermt/bergamot-translator/pull/94#discussion_r610526930

src/translator/byte_array_util.cpp
```diff
@@ -27,8 +25,8 @@ const T* get(const void*& current, uint64_t num = 1) {
   return ptr;
 }
 
-bool validateBinaryModel(AlignedMemory& model, uint64_t fileSize) {
-  const void * current = &model[0];
+bool validateBinaryModel(const AlignedMemory *model, uint64_t fileSize) {
+  const void * current = model->begin();
```
jerinphilip, line 29: What's happening here? Aren't references preferable over pointers because the former cannot be null?

kpu: A void reference seems icky to me.

src/translator/byte_array_util.cpp
abhi-agg: @qianqianzhu: 2 questions:

1. Does marian avoid loading the vocabularies twice into memory when both the vocabulary files passed to its command line parameter `--vocabs` are the same?

2. If yes, how does it decide to do so? By checking whether both file names are same?

From the API consumer perspective, I am probably missing the point of having `vocabIndices`. Providing single entry in `vocabMemories` but 2 in `vocabIndices` mean the consumer of the API should be aware that the 2 vocab files are same and hence create only 1 AlignedMemory object. If the API consumer is expected to know this then probably we could get away with `vocabIndices` completely. Would following work:

1. When both vocab files are exactly the same (e.g. `esen` case), the API consumer will create only 1 `AlignedMemory` object, create a vector with this object and call the `Service` constructor.

2. When both vocab files are different, the consumer will create 2 `AlignedMemory` objects, create a vector with these 2 objects and call the `Service` constructor.

Let me know if I missed something here 👍
cc @jerinphilip @kpu

src/translator/service.h
abhi-agg: Cool. Thanks.

src/translator/byte_array_util.cpp
qianqianzhu: Sorry I will move the header and get in an anonymous namespace. `validateBinaryModel()` function is needed for BatchTranslator, so we can have the validation code in one place.

src/translator/byte_array_util.cpp
qianqianzhu: I was trying to call `validateBinaryModel` in `BatchTranslator` where the model bytes is passed as const AlignedMemory*. I will change to `const AlignedMemory &model`.

src/translator/service.cpp
```diff
@@ -6,33 +6,51 @@
...
   return vocabs;
 }
 
 namespace marian {
 namespace bergamot {
 
-Service::Service(Ptr<Options> options, AlignedMemory modelMemory, AlignedMemory shortlistMemory) 
-    : requestId_(0), options_(options), vocabs_(std::move(loadVocabularies(options))),
+Service::Service(Ptr<Options> options, AlignedMemory modelMemory, AlignedMemory shortlistMemory,
+                 std::vector<AlignedMemory> vocabMemories, std::vector<size_t> vocabIndices)
```
jerinphilip, line 47: @qianqianzhu We have 3, maybe 4 arguments related to bytearray now, one more (shortlist) incoming. Now might be a good time to:

* https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#i23-keep-the-number-of-function-arguments-low

src/translator/service.h
```diff
@@ -64,10 +64,15 @@ class Service {
...
+  /// of a model.bin.
   /// @param shortlistMemory byte array of shortlist (aligned to 64)
+  /// @param vocabMemories vector of unique vocabulary memories (aligned to 64)
+  /// @param vocabIndices vector of vocabulary memory indices where each vocabulary
+  /// are loaded
   explicit Service(Ptr<Options> options, AlignedMemory modelMemory,
-                   AlignedMemory shortlistMemory);
+                   AlignedMemory shortlistMemory,
+                   std::vector<AlignedMemory> vocabMemories,
+                   std::vector<size_t> vocabIndices);
```
abhi-agg, line 75: I left my comment regarding `vocabIndices` [here](https://github.com/browsermt/bergamot-translator/pull/122/files#r625791112).

src/translator/byte_array_util.cpp
qianqianzhu: Sorry I missed this comment. 

@abhi-agg For marian, it does not care whether the vocabularies are the same or not. It just loads the files even they are the same. You can see here: https://github.com/browsermt/marian-dev/blob/master/src/translator/translator.h#L231-L238

For bergamot-translator, we did optimise the loading part: we avoid double loading by checking the filenames. https://github.com/browsermt/bergamot-translator/blob/main/src/translator/service.cpp#L19-L22

src/translator/service.cpp
qianqianzhu: Do you want to wrap all AlignedMemory objects together?

src/translator/byte_array_util.cpp
qianqianzhu: Need some help here to make a decision. @kpu @jerinphilip

src/translator/service.cpp
jerinphilip: I'm thinking a self describing structure like the following (I gave quick names, you can probably come up with something easier) which should make code in a lot of places neater. 

```
struct VocabBundle {
	std::vector<AlignedMemory> uniqueVocabs;
	set::vector<size_t> vocabIndices;
};

struct MemoryBundle {
	AlignedMemory model;
	AlignedMemory shortlist;
	VocabBundle vocabs;
	AlignedMemory ssplitPrefixFile;
};
```
Of course you don't have `ssplitPrefixFile` yet, but you can see how further changes become simplified with this structure. There's also discussions surrounding if MemoryBundle (or a subset) will be the single binary Mozilla provides us.

XapaJIaMnu: Would be nice to have, but I think it's outside of the scope of this pr. Refactoring may come later.
Technically, we could have arbitrary number of models, shortlists and vocabularies loaded too (marian supports that), but we don't support that.

src/translator/byte_array_util.cpp
XapaJIaMnu: @qianqianzhu still an issue, or did you choose a solution?

qianqianzhu: I will go for `std::vector<marian::Ptr<AlignedMemory>>` if it is ok @XapaJIaMnu.

abhi-agg: @qianqianzhu Would passing just `std::vector<AlignedMemory>` not work?

The API consumer will pass `std::vector<AlignedMemory>` containing unique AlignedMemory objects.
Listing down some scenarios:

Scenario 1:
Source vocabs = "vocab1", "vocab2", "vocab3"
Target vocabs = "vocab4"

You will have 4 unique entries in `std::vector<AlignedMemory>`

Scenario 2:
Source vocabs = "vocab1"
Target vocabs = "vocab1"

You will have 1 unique entry in `std::vector<AlignedMemory>`

Did I miss any other scenario? If not then this solution should work?

qianqianzhu: @abhi-agg you are right. Thought `shared_ptr` would be easier to manage the objects.

qianqianzhu: Wait. @abhi-agg How about the following scenario:
Source vocabs = "vocab1", "vocab2", "vocab1"
Target vocabs = "vocab1"

abhi-agg: @qianqianzhu I don't have any issue with using `shared_ptr` in the API but this discussion opened up a topic for clarification that probably you, @kpu, @XapaJIaMnu or @jerinphilip can answer :)

1. Can marian work with multiple source vocabularies?
    1. Reason of confusion: The [code](https://github.com/browsermt/marian-dev/blob/master/src/translator/translator.h#L231-L238) + [data members](https://github.com/browsermt/marian-dev/blob/master/src/translator/translator.h#L213-L214) lead me to believe that yes it is possible. However, I couldn't find any example in [marian documentation](https://marian-nmt.github.io/docs/) that uses multiple source vocabularies. The [--vocabs option](https://github.com/browsermt/marian-dev/blob/master/src/common/config_parser.cpp#L643-L644) is indeed a vector<string> but it is not mentioned anywhere whether the user can pass more than 2 entries for it.
 
2. I think the answer to this one is "No" but for the sake of completion asking here, can marian work with multiple target vocabularies?

If marian is not supposed to work with multiple source vocabularies then we don't need to use `shared_ptr` in the API because it wouldn't be needed. If both source and target vocab is same, `vector<AlignedMemory>` will contain single entry otherwise 2 entries :) However, I am still fine with `shared_ptr` 👍

kpu: I think we should use `shared_ptr` (and not marian::Ptr which is the same thing but exposes a Marian internal).  

Marian can use different vocabularies for source and target.  More generally, Marian supports multiple inputs each of which can have its own vocabulary.  This has been used for example to do http://statmt.org/wmt21/triangular-mt-task.html .

abhi-agg: >  Marian can use different vocabularies for source and target. More generally, Marian supports multiple inputs each of which can have its own vocabulary.

This clears few things. Thanks.

For the case when the user specifies only 1 entry for `--input` option, can the user specify multiple source and target vocabularies?

src/translator/service.cpp
jerinphilip: I'm ok with doing the struct instead of multiple args in a different PR.

qianqianzhu: I will leave this refactoring later (in a new PR).

src/translator/service.h
```diff
@@ -76,24 +78,28 @@ class Service {
   ///
   /// This is equivalent to a call to:
   /// ```cpp
-  ///    Service(options, AlignedMemory(),  AlignedMemory())
+  ///    Service(options, AlignedMemory(), AlignedMemory(), {})
   /// ```
   /// wherein empty memory is passed and internal flow defaults to file-based
-  /// model, shortlist loading.
+  /// model, shortlist loading. AlignedMemory() corresponds to empty memory
   explicit Service(Ptr<Options> options)
```
XapaJIaMnu, line 85: This is probably not your fault, but any reason why we have 3 different constructors as opposed to using default arguments?

qianqianzhu: I guess the reason for the 3 constructors here is to keep a purely file-based one and a bytearray-based one separated. We can get rid of different constructors in a later PR (addressing https://github.com/browsermt/bergamot-translator/pull/122#discussion_r625868489)

abhi-agg: There are 2 classes of constructors right now:

1. One that uses `std::string` to pass config ([constructor 1](https://github.com/browsermt/bergamot-translator/blob/main/src/translator/service.h#L92-L94))
    1. This is the constructor that browser extension currently uses (as it is independent of marian internal structs)
2. Other class that uses `marian::Options` to pass config ([constructor 2](https://github.com/browsermt/bergamot-translator/blob/main/src/translator/service.h#L83-L84) and [constructor 3](https://github.com/browsermt/bergamot-translator/blob/main/src/translator/service.h#L69-L70)). I believe @jerinphilip can throw some light on why we need these ones 👍 
    1. I think we can merge [constructor 2](https://github.com/browsermt/bergamot-translator/blob/main/src/translator/service.h#L83-L84) and [constructor 3](https://github.com/browsermt/bergamot-translator/blob/main/src/translator/service.h#L69-L70) into one
    2. Probably, we should make this constructor private because `Service` class now is the main API class and we wouldn't want to expose marian internal structs in public APIs. However, I am fine with keeping it public if there is a reason for it.

src/translator/byte_array_util.cpp
abhi-agg: @qianqianzhu Is this resolved?

qianqianzhu: Yes!

src/translator/service.h
```diff
@@ -76,24 +78,28 @@ class Service {
...
-  /// @param [in] shortlistMemory byte array of shortlist (aligned to 64)
+  /// @param [in] modelMemory byte array (aligned to 256!!!) that contains the
+  /// bytes of a model.bin. Optional. AlignedMemory() corresponds to empty memory
+  /// @param [in] shortlistMemory byte array of shortlist (aligned to 64). Optional.
+  /// @param [in] vocabMemories vector of vocabulary memories (aligned to 64). Optional.
   explicit Service(const std::string &config,
                    AlignedMemory modelMemory = AlignedMemory(),
-                   AlignedMemory shortlistMemory = AlignedMemory())
+                   AlignedMemory shortlistMemory = AlignedMemory(),
+                   std::vector<Ptr<AlignedMemory>> vocabsMemories = {})
```
abhi-agg, line 98: I know that we are using `Ptr<Options>` which although is same as std::shared_ptr<Options>, however could you please use `std::shared_ptr<>` for `vocabsMemories` (in the Service class at least)?

XapaJIaMnu:
I think the comments on this PR are minor and any issues would refer to refactoring that needs to happen. We can open separate issues and prs for those.

XapaJIaMnu (approved):

src/translator/service.h
qianqianzhu: Yes. Fixed.

src/translator/service.h
```diff
@@ -76,24 +78,28 @@ class Service {
...
   /// Construct Service from a string configuration.
   /// @param [in] config string parsable as YAML expected to adhere with marian
   /// config
-  /// @param [in] model_memory byte array (aligned to 256!!!) that contains the
-  /// bytes of a model.bin. Optional.
-  /// @param [in] shortlistMemory byte array of shortlist (aligned to 64)
+  /// @param [in] modelMemory byte array (aligned to 256!!!) that contains the
+  /// bytes of a model.bin. Optional. AlignedMemory() corresponds to empty memory
+  /// @param [in] shortlistMemory byte array of shortlist (aligned to 64). Optional.
+  /// @param [in] vocabMemories vector of vocabulary memories (aligned to 64). Optional.
```
abhi-agg, line 94: @qianqianzhu One final question. When the source and target vocabulary is exactly the same file, does the API expect 2 entries in `vocabMemories` or single entry? Documenting it makes sense 👍

qianqianzhu: > When the source and target vocabulary is exactly the same file, does the API expect 2 entries in vocabMemories or single entry?
The API will expect 2 entries. These two entries (shared pointers) share the same AlignedMemory object. Added to the documentation.

src/translator/service.cpp
qianqianzhu: I think `loadVocabularies()` should not be in `Service` (maybe put in vocab.h or somewhere). Also this function needs refactoring. I will fix it in a new PR (together with the memory bundle).

qianqianzhu:
Quickly run the regression tests. AVX2 is ok.

src/translator/service.h
jerinphilip: > This is probably not your fault

This is my fault, and part of me still thinks the fault should stay. NULL values (empty `AlignedMemory`) as default argument is what should go away (this is kludge added to get green tick on CI not breaking a late `TranslationModel`), with `Service` accepting only legit bytearray payloads thus making way for two constructors (one from options with file-loads and one from options + bytearray payload). I'm not going to count the delegated constructor, it has only a minimal maintenance overhead. 

> We can get rid of different constructors in a later PR

Mark this as resolved?

src/translator/byte_array_util.cpp
qianqianzhu: > For the case when the user specifies only 1 entry for --input option, can the user specify multiple source and target vocabularies?

I don't know. But my guessing is no as one input corresponds to one vocabulary. Multiple source vocabularies for one input source seems not necessary. @XapaJIaMnu knows better than me.

kpu: Every system you will encounter @abhi-agg has exactly two vocabulary slots: one for source and one for target.  These vocabularies can be the same.  

If you provide a single vocabulary slot, shame on you that's an error.

abhi-agg: > Every system you will encounter @abhi-agg has exactly two vocabulary slots: one for source and one for target. These vocabularies can be the same.
> 
> It essentially never makes sense to have multiple vocabularies for the same input and anyway Marian doesn't support it.

This explanation clears everything for me. Thanks again 👍

abhi-agg (approved):
@qianqianzhu Thank you for the awesome work 👍

jerinphilip (approved):
I'd prefer the `Ptr`/`shared_ptr` in `vector<Ptr<AlignedMemory>>` to not be there, but that's what one gets on enforcing one argument. 

Other than that, I'm okay with the redundancies (which I understand will be in taken care of later) given the feature (vocab byte array loading) working and approve this PR in the current state.

jerinphilip:
> Quickly run the regression tests. AVX2 is ok.

@qianqianzhu FYI, green ticks in CI now include regression-tests run (take a look at the CI log), at least the ones that can run on a tiny system.

PR #123, opened 2021-04-30
Improve inference speed for model and shortlist bytes

abhi-agg:
1. Loading Model and Shortlist as bytes caused a reduction in inference speed
    1. Added the missing `gemm-precision` option in config of wasm test page that improved it
2. Improved wasm README
    1. The README is updated to show the usage of JS APIs

### DIFF
--- a/wasm/README.md
+++ b/wasm/README.md
@@ -17,11 +17,37 @@ gunzip models/*/*
 
 ```js
 // The model configuration as YAML formatted string. For available configuration options, please check: https://marian-nmt.github.io/docs/cmd/marian-decoder/
-// This example captures the most relevant options: model file, vocabulary files and shortlist file
-const modelConfig = "{\"models\":[\"/esen/model.esen.npz\"],\"vocabs\":[\"/esen/vocab.esen.spm\",\"/esen/vocab.esen.spm\"],\"shortlist\":[\"/esen/lex.esen.s2t\"],\"beam-size\":1}";
+// This example captures some of the most relevant options
+const modelConfig = `vocabs:
+  - /esen/vocab.esen.spm
+  - /esen/vocab.esen.spm
+beam-size: 1
+normalize: 1.0
+word-penalty: 0
+max-length-break: 128
+mini-batch-words: 1024
+workspace: 128
+max-length-factor: 2.0
+skip-cost: true
+cpu-threads: 0
+quiet: true
+quiet-translation: true
+gemm-precision: int8shift
+`;
+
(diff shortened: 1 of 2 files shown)


wasm/README.md
```diff
@@ -17,11 +17,37 @@ gunzip models/*/*
...
+word-penalty: 0
+max-length-break: 128
+mini-batch-words: 1024
+workspace: 128
+max-length-factor: 2.0
+skip-cost: true
+cpu-threads: 0
+quiet: true
+quiet-translation: true
+gemm-precision: int8shift
```
XapaJIaMnu, line 35: For maximum performance this should be `int8shiftAlphaAll` or `int8shiftAlpha`  (not sure which of the two since I haven't tested inside the WASM environment) https://github.com/browsermt/students/blob/3344c53660acc160714ff8394c8155db65cf9c13/eten/enet.student.tiny11/config.intgemm8bitalpha.yml#L20

wasm/test_page/bergamot.html
```diff
@@ -127,6 +127,7 @@
 cpu-threads: 0
 quiet: true
 quiet-translation: true
+gemm-precision: int8shift
```
XapaJIaMnu, line 130: For maximum performance this should be `int8shiftAlphaAll` or `int8shiftAlpha`  (not sure which of the two since I haven't tested inside the WASM environment) https://github.com/browsermt/students/blob/3344c53660acc160714ff8394c8155db65cf9c13/eten/enet.student.tiny11/config.intgemm8bitalpha.yml#L20

XapaJIaMnu (changes requested):
You're still using suboptimal parameters

wasm/README.md
abhi-agg: @XapaJIaMnu I am not using `int8shiftAlphaAll` intentionally because wasm is having issue for EN->DE model with `int8shiftAlphaAll` (it fails with `Error: We did not find an alpha in the model named: F0::Wemb_QuantMultA.`).

I have to create an issue to investigate it.

However, I didn't want to loose out on the inference speed up (almost 2x) that comes by using `int8shift`. That is why, I pushed the change with `int8shift` and would like to merge it. The change with `int8shiftAlphaAll` will follow later.

I hope that explains :)

XapaJIaMnu: Are you packaging the right model? Can you link me to the model you are using, i can probably fix it quickly.

XapaJIaMnu (approved):

motin (approved):

Issue #119, opened 2021-04-29
Testing / CD meta issue


jerinphilip:
1. ~~Collapse x and x-bytearray apps into one through a command line flag. There's a lot of redundancy in the printing code for both and stuff is requiring edits in both places. (#120)~~
2. (In Progress: #184) There's one output app where outputs for all features are bound to regression tests. I would prefer to break this down to feature level so edits are not all over the place while we parallelize development which affects each of these features:
    a. source text -> translated text
    b. source sentence mappings -> translated sentence-mappings (checking sentence alignments)
    c. quality
    d. alignments
3. BRT is bound to hardware / instructions available. This is not pleasant as generation and moving around when code is updated to change output is a lot of manual labour at the moment. Once there are 4 x #instructions (ssse3, avx, avx2, avx512, avx512vnni = 5 at the moment) is a lot of files to store as expected ground truths. Something needs to change here so things can be automated. https://github.com/kpu/intgemm/issues/84 this can solve part of the issue. Nick describes this also being a function of the compiler and BLAS supplied.
4. As of now we're testing code here with only the en -> de pipeline. For purposes of a model zoo (which I'm guessing is the ideal direction), it's best if we generalize this for other language directions and changes in models should also trigger errors in regression tests if something fails (https://github.com/browsermt/students or https://github.com/mozilla-applied-ml/bergamot-models).

XapaJIaMnu:
As to point 2)
I think a simple sentence mapping a BLEU score with some tolerance of 0.1 would probably work just as well to detect breakage.

As to point 3)
1. Floating point precision issues. When adding/mutliplying floating point numbers the results may vary based on the compiler used/OS used/library version used/CPU hardware used. Currently all of our models go through a BLAS provider (AppleAccelerate/MKL/OpenBLAS). We have ssse3-avx512vnni architectures and I know of no reliable way to limit the instruction generation to be only of specific range (eg up to ssse3) when linking with the libraries, because they do runtime CPU detection and have dedicated codepaths for various architectures. Some flags exist, eg:
```
MKL_ENABLE_INSTRUCTIONS=SSE4_2
MKL_ENABLE_INSTRUCTIONS=AVX
MKL_ENABLE_INSTRUCTIONS=AVX2
MKL_ENABLE_INSTRUCTIONS=AVX512
```
But I am not sure if SSE4_2 MKL codepath would produce the same result on a very old processor, as well as on a very new processor (BUT IT MAY HELP).
[...]

PR #127, opened 2021-05-03
Making bytearray a commandline switch

jerinphilip:
- [x] Remove `bergamot-app-bytearray` (This is moot since TranslationModel is gone and service-cli-bytearray achieves the intended outcome).
- [x] Collapse `service-cli{,-bytearray}`. This is a command line switch now.
- [x] Update brt scripts accordingly.

There's a precision issue from #119, which will be taken up separately. For now, mac tests are restricted in BRT to have only one test.


src/translator/parser.h
```diff
@@ -23,9 +23,13 @@ inline marian::ConfigParser createConfigParser() {
...
       "Maximum input tokens to be processed in a single sentence.", 128);
 
+  cp.addOption<bool>(
+      "--bytearray", "Bergamot Options",
+      "Enable loads through byte-arrays to enable testing of byte-array codepath (false by default)", false);
+
   cp.addOption<bool>(
       "--check-bytearray", "Bergamot Options",
-      "Flag holds whether to check the content of the bytearray (true by default)", true);
+      "Flag holds whether to check the content of the bytearray (false by default)", false);
```
jerinphilip, line 32: @qianqianzhu I changed the default behaviour here. Setting `--bytearray` to false and `--check-bytearray` default to true leads to validation issues. So to ensure that the cli always also validates the blobs, I manually change the parsed option to true in the other place (https://github.com/browsermt/bergamot-translator/pull/127/files#r626677517).

XapaJIaMnu: Can we get a better solution. I don't like the hardcoded  `  options->set<bool>("check-bytearray", true);` even if it's just in the test binary.

qianqianzhu: I see your point. What I meant https://github.com/browsermt/bergamot-translator/pull/127#discussion_r626677517 here is that we do not need to manually set `--check-bytearray` to true.

app/service-cli.cpp
```diff
@@ -14,7 +15,21 @@
...
+  // Prepare memories for model and shortlist
+  marian::bergamot::AlignedMemory modelBytes, shortlistBytes;
+
+  if (options->get<bool>("bytearray")) {
+    // Load legit values into bytearrays.
+    modelBytes = marian::bergamot::getModelMemoryFromConfig(options);
+    shortlistBytes = marian::bergamot::getShortlistMemoryFromConfig(options);
+
+    // Service command line always has check-bytearray on.
+    options->set<bool>("check-bytearray", true);
```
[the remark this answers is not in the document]
jerinphilip: I have changed this to use `--check-bytearray`. Default being always check and default being loaded from bytearray. The scripts have explicit arguments to match old behaviour. Is this ok?

src/translator/parser.h
jerinphilip: @XapaJIaMnu Take a look now?

qianqianzhu (approved):

XapaJIaMnu (approved):

PR #128, opened 2021-05-04
Improved wasm scripts and README

abhi-agg:
- Build the wasm artifacts in the top level build folder instead of build/wasm folder
- Robust scripts to patch wasm artifacts and starting the http server
    - **There is no change in what they were doing before**
    - Just added an argument to accept the folder where the wasm artifacts reside
- Updated root level README
    - To not package model and lexical files in wasm binary since they are provided at run time
        - It avoids bloating up the wasm binary significantly

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,19 +36,18 @@ Bergamot translator provides a unified API for ([Marian NMT](https://marian-nmt.
     cd bergamot-translator
     ```
 
-3. Download files (only required if you want to package files in wasm binary)
+3. Download files (only required if you want to perform inference using build artifacts)
 
-    This step is only required if you want to package files (e.g. models, vocabularies etc.)
-    into wasm binary. If you don't then just skip this step.
+    It packages the vocabulary files into wasm binary, which is required only if you want to perform inference.
+    The compilation commands will preload these files in Emscripten’s virtual file system.
 
-    The build preloads the files in Emscripten’s virtual file system.
-
-    If you want to package bergamot project specific models, please follow these instructions:
+    If you want to package bergamot project specific files, please follow these instructions:
     ```bash
-    mkdir models
     git clone --depth 1 --branch main --single-branch https://github.com/mozilla-applied-ml/bergamot-models
+    mkdir models
     cp -rf bergamot-models/prod/* models
     gunzip models/*/*
+    find models \( -type f -name "model*" -or -type f -name "lex*" \) -delete
     ```
 
 4. Compile
@@ -59,14 +58,14 @@ Bergamot translator provides a unified API for ([Marian NMT](https://marian-nmt.
         ```
 
     2. Compile the artefacts
-        * If you want to package files into wasm binary then execute following commands (Replace `FILES_TO_PACKAGE` with the path of the
-        directory containing the files to be packaged in wasm binary)
+        * If you want to package files into wasm binary then execute following commands (Replace `FILES_TO_PACKAGE` with the
+        directory containing all the files to be packaged)
 
             ```bash
             emcmake cmake -DCOMPILE_WASM=on -DPACKAGE_DIR=FILES_TO_PACKAGE ../
             emmake make -j
             ```
-            e.g. If you want to package bergamot project specific models (downloaded using step 3 above) then
+            e.g. If you want to package bergamot project specific files (downloaded using step 3 above) then
             replace `FILES_TO_PACKAGE` with `../models`
 
         * If you don't want to package any file into wasm binary then execute following commands:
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -36,19 +36,18 @@ Bergamot translator provides a unified API for ([Marian NMT](https://marian-nmt.
     cd bergamot-translator
     ```
 
-3. Download files (only required if you want to package files in wasm binary)
+3. Download files (only required if you want to perform inference using build artifacts)
```
motin, line 39: Why else would anyone want to use the build artifacts? :) Also, I think it is a bit inaccurate, since inference is possible without the vocabs, it is just slower. Maybe something like "to improve the translation speed"?

README.md
```diff
@@ -36,19 +36,18 @@ Bergamot translator provides a unified API for ([Marian NMT](https://marian-nmt.
...
-
-    If you want to package bergamot project specific models, please follow these instructions:
+    If you want to package bergamot project specific files, please follow these instructions:
     ```bash
-    mkdir models
     git clone --depth 1 --branch main --single-branch https://github.com/mozilla-applied-ml/bergamot-models
+    mkdir models
     cp -rf bergamot-models/prod/* models
     gunzip models/*/*
+    find models \( -type f -name "model*" -or -type f -name "lex*" \) -delete
```
jerinphilip, line 50: So vocabs are now being kept and loaded into data? Others are deleted to reduce binary size if I understand correctly?

README.md
```diff
@@ -36,19 +36,18 @@ Bergamot translator provides a unified API for ([Marian NMT](https://marian-nmt.
     cd bergamot-translator
     ```
 
-3. Download files (only required if you want to package files in wasm binary)
+3. Download files (only required if you want to perform inference using build artifacts)
 
-    This step is only required if you want to package files (e.g. models, vocabularies etc.)
-    into wasm binary. If you don't then just skip this step.
+    It packages the vocabulary files into wasm binary, which is required only if you want to perform inference.
```
motin, line 41: Suggest simplifying the language here. `which is required only if you want to perform inference` to `which is required to improve the translation speed`

motin (commented):
Looks good overall!

motin (approved):

README.md
motin: Correct, at least in relation to the wasm test page in this repo. The extension also loads vocabs dynamically, to further decrease the binary size.

jerinphilip (approved):

README.md
abhi-agg: > Why else would anyone want to use the build artifacts? :) 

I agree with you on that one. I will remove it.

> Also, I think it is a bit inaccurate, since inference is possible without the vocabs, it is just slower. Maybe something like "to improve the translation speed"?

@motin Inference is not possible without vocabs. So that statement is correct 👍

README.md
abhi-agg: As per my comment above, this comment in invalid. So, ignoring this one 👍

README.md
abhi-agg: Yep. You are right. Others are being loaded at run time. Once vocabs can also be passed as bytes, we will stop including vocabs at build time 👍

wasm/CMakeLists.txt
```diff
@@ -23,6 +23,7 @@ endif()
 set_target_properties(bergamot-translator-worker PROPERTIES
                         SUFFIX ".js"
                         LINK_FLAGS ${LINKER_FLAGS}
-                        )
+                        RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}
```
jerinphilip, line 26: Why did we require this change? This is what's breaking tests?

abhi-agg: Yes. Submitted the fix for it with the latest commit.

README.md
kpu: I think you're confusing vocabs and lexical shortlists.  Vocabs are required.  Lexical shortlists are optional.

motin: Ah yes indeed I am! Thanks @abhi-agg and @kpu for clarifying!

motin (approved):

jerinphilip (approved):

abhi-agg:
@jerinphilip Although doing squash and merge might be a good idea while merging some PRs, I would have preferred rebase and merge for this one because all commits here were modular and if we ever have to revert a specific commit, it would have been easier 👍

In general, I believe there is no single correct answer when it comes to which merge method to use, may be we can leave the responsibility of merging on the PR creator 👍

PR #129, opened 2021-05-04
Extension desired changes

motin:
This PR includes a couple of changes that makes the glue code compatible with the current bergamot-browser-extension implementation. 

WorkerFS is used to mount remotely downloaded model files (including vocab files) in the emscripten's file system so that no files needs to be bundled in the wasm artifacts. 

Avoiding node.js-code in Emscripten glue-code allows the glue code to be integrated properly in the extension's webpack build configuration.


abhi-agg (approved):

abhi-agg:
@jerinphilip Pushing this change as it doesn't affect native part and we need it for extension asap 👍

PR #135, opened 2021-05-08
Rewrite annotation class to remove corner cases

kpu:
This cleans up Annotation to store just offsets where tokens begin and explicitly model the gaps between sentences (as well as before and after).  

Empty sentences and empty space are no longer a special case, they just happen naturally.  https://www.youtube.com/watch?v=UfJljXugLA0

Now that the accessors are simpler, they are in the headers so the compiler can reason about them.  

Still TODO is the two methods for filling Annotation (give it a string or append stuff) could use better cohesiveness and documentation.

### DIFF
--- a/src/translator/annotation.cpp
+++ b/src/translator/annotation.cpp
@@ -20,7 +20,7 @@ void AnnotatedText::appendSentence(string_view prefix, std::vector<string_view>:
   // Appending sentence text.
   std::size_t offset = text.size();
   if (begin != end) {
-    text.append(begin->data(), (end - 1)->data() + (end - 1)->size() + 1);
+    text.append(begin->data(), (end - 1)->data() + (end - 1)->size());
   }
   for (std::vector<string_view>::iterator token = begin; token != end; ++token) {
     offset += token->size();
--- a/src/translator/annotation.cpp
+++ b/src/translator/annotation.cpp
@@ -1,16 +1,16 @@
 #include "annotation.h"
 #include <cassert>
-#include <iostream>
 
 namespace marian {
 namespace bergamot {
 
 AnnotatedText::AnnotatedText(std::string &&t) : text(std::move(t)) {
   // Treat the entire text as a gap that recordExistingSentence will break.
-  annotation.token_begin_.back() = t.size();
+  annotation.token_begin_.back() = text.size();
 }
 
 void AnnotatedText::appendSentence(string_view prefix, std::vector<string_view>::iterator begin, std::vector<string_view>::iterator end) {
+  assert(annotation.token_begin_.back() == text.size());
   // We'll be adding tokens from the sentence and another gap.
   annotation.token_begin_.reserve(annotation.token_begin_.size() + (end - begin) + 1);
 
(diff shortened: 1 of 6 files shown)


src/translator/annotation.cpp
```diff
@@ -1,130 +1,68 @@
...
-  size_t offset = text.size(); // Get size before to do ByteRange arithmetic
-  text += reference;           // Append reference to text
-  std::vector<ByteRange> sentence;
-  for (auto &wordView : wordRanges) {
-    size_t thisWordBegin = offset + wordView.data() - reference.data();
-    sentence.push_back(
-        ByteRange{thisWordBegin, thisWordBegin + wordView.size()});
+  if (begin != end) {
+    text.append(begin->data(), (end - 1)->data() + (end - 1)->size());
+    assert(offset == text.size()); // Catch people without consecutive tokens.
```
XapaJIaMnu, line 28: People?

XapaJIaMnu (approved):

src/translator/annotation.h
```diff
@@ -122,7 +137,31 @@ struct AnnotatedText {
...
+  /// Append the whitespace at the end of input. string_view must not be in
+  /// text.
+  void appendEndingWhitespace(string_view whitespace);
+
+  /// Record the existence of a sentence that is already in text.  The
+  /// iterators are over string_views for each token that must be in text
+  /// already.  This function must be called to record sentences in order.
+  /// Normally the beginning of the sentence can be inferred from
+  /// tokens_begin->data() but the tokens could be empty, so sentence_begin is
+  /// required to know where the sentence is.
```
jerinphilip, line 160: I think we're going to have at least EOS - which should point to the end of the active sentence being inserted, so can get rid of the extra sentence_begin eventually.

src/translator/annotation.h
```diff
@@ -17,83 +17,99 @@ struct ByteRange {
...
+    return gap_[sentenceIdx + 1] - gap_[sentenceIdx] - 1 /* minus the gap */;
+  }
 
   /// Returns a ByteRange representing `wordIdx` in sentence indexed by
   /// `sentenceIdx`. `wordIdx` follows 0-based indexing, and should be less than
   /// `.numWords()` for `sentenceIdx` for defined behaviour.
-  ByteRange word(size_t sentenceIdx, size_t wordIdx) const;
+  ByteRange word(size_t sentenceIdx, size_t wordIdx) const {
+    size_t tokenIdx = gap_[sentenceIdx] + 1 + wordIdx;
+    return ByteRange {token_begin_[tokenIdx], token_begin_[tokenIdx + 1]};
```
jerinphilip, line 65: Since we're doing asserts, there's an assert missing here described loosely as:

```
gap_[sentenceIdx] + 1 + wordIdx < gap_[sentenceIdx+1] + 1
```

XapaJIaMnu:
So.... Should this be merged?

XapaJIaMnu:
@jerinphilip ?

kpu:
As discussed on the plenary call, quality estimation will be returning piece-level information.  These might be "upgraded" to word-level for display.  We'd have to define word again, given that we don't have a tokenizer that e.g. splits , or ! off a word.  Space is possibly ok for space-separated languages; I guess piece will do for unsegmented languages.  

So there isn't a quality estimation grounds to block this particular representation.

jerinphilip (approved):
We'll live with `size_t`s instead of `ByteRange`s.

PR #138, opened 2021-05-09
Change WASM_COMPATIBLE_SOURCE=OFF by default

kpu:
The default was WASN_COMPATIBLE_SOURCE=ON COMPILE_WASM=OFF which is a
testing configuration, not a sensible default for native or wasm.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -35,7 +35,7 @@ include(CMakeDependentOption)
 
 # Project specific cmake options
 option(COMPILE_WASM "Compile for WASM" OFF)
-option(USE_WASM_COMPATIBLE_SOURCE "Use wasm compatible sources" ON)
+option(USE_WASM_COMPATIBLE_SOURCE "Use wasm compatible sources" OFF)
 option(COMPILE_TESTS "Compile bergamot-tests" OFF)
 
 SET(PACKAGE_DIR "" CACHE STRING "Directory including all the files to be packaged (pre-loaded) in wasm builds")


XapaJIaMnu:
@abhi-agg remind me again why do we have `WASM_COMPATIBLE_SOURCES` and `COMPILE_WASM`. Does this break your workflow?

jerinphilip:
> I suppose the useful part of this is being able to natively test a "reduced" Marian. Though I'm not convinced it's really a reflective test given the changes on top to disable threads. And I still don't understand why Mozilla keeps disabling threads when the plan is to use them.


* https://github.com/browsermt/bergamot-translator/pull/74#discussion_r602369482

> Having capability to compile "reduced-marian" on native platform is helpful in debugging the wasm runtime failures by first reproducing the same failures on native platform with "reduced-marian" native binaries. In the past, this has proven to be useful where we could quickly diagnose a few wasm specific issues (e.g. inconsistent size of size_t data type on wasm and native platform). Therefore, I am proposing strongly to keep 3rd workflow.



* https://github.com/browsermt/bergamot-translator/pull/74#issuecomment-809264006

kpu:
Indeed @jerinphilip this change preserves the ability to compile `-DUSE_WASM_COMPATIBLE_SOURCE=ON` natively, just not as the default.  I don't see how a change in default is in conflict with anything you quoted above.

jerinphilip (approved):
> @jerinphilip can quote me all he wants above... 

Confused. The response was context to answer Nick's query, in the interest of speeding this up? 

I have no problem with what this PR does and am for it since the beginning.

> ~~This path needs to go away as a whole~~, saner default set ~~and not add more files to add weight for everyone to carry~~.

kpu:
Oh I understand the conversation now.

abhi-agg (approved):

CMakeLists.txt
```diff
@@ -4,6 +4,10 @@ if (POLICY CMP0074)
   cmake_policy(SET CMP0074 NEW) # CMake 3.12
 endif ()
 
+if (POLICY CMP0077)
```
XapaJIaMnu, line 7: Could you comment here and specify what this policy accomplishes?

Issue #77, opened 2021-03-29
Collapse draft API and actual implementation


kpu:
The following class pairs should collapse into one class.  

1. `Service` and `TranslationModel`
2. `Response` and `TranslationResult`
3. `Request` and `TranslationRequest`

My understanding is the `Translation` classes came from the "Unified API" which was a draft of what the API could look like.  During that exercise, we agreed that the API would change with actual implementation.  It always does.  An API written without implementation inevitably does not capture the full complexity of the task.  It also adds unnecessary complexity with functionality that neither the backend nor front end actually uses.  Moreover, it does not consider what is efficient to pass and there's no point to multiple format conversions.  

Ideally a skeleton API class should be subsumed into the implemented class.  There will be no converter between them, which is just a source of bugs.  

My understanding is nobody is calling alignments and quality estimation, so it's easiest to merge there.  

@jerinphilip Let's avoid arguing about casing for file names and try to avoid reinvention where possible.  

@abhi-agg A working implementation should not be taxed because its design differs.  

@motin @mlopatka @andrenatal Happy to discuss.

jerinphilip:
**On Collapse**
1. `TranslationRequest` and `marian::bergamot::Request` cannot be collapsed, they're nowhere near parity and never will be (Request has marian internal structs, it's not going anywhere else). `TranslationRequest` doesn't even have the input string, it's an options dict on what they need in addition to translation.
2. `marian::bergamot::Response`, `TranslationResult` should be largely collapsable after resolving 3 issues (1 of which is solved in a pending PR). 
3. `marian::bergamot::Service`, `TranslationModel` is probably collapsable, given `TranslationModel` doesn't do anything as of now. This will however needs propagation due to `isAlignmentSupported` being present in places at extension. `TranslationModel` has a blocking API (no `std::future<Response>`, perhaps offends WASM?) with ugly arguments (`vector<string>`) which creates issues with batching. 

[...]

PR #139, opened 2021-05-09
Collapsing TranslationRequest -> ResponseOptions

jerinphilip:
Towards #77.

`QualityScore.h` was only used by `TranslationRequest` before (after `TranslationResult` merged into `Response`). This can now be deleted (https://github.com/browsermt/bergamot-translator/pull/109#discussion_r623957470).

For a first-step `TranslationRequest` is typedef-ed as `ResponseOptions` to see if it's a clean collapse.

TODO:
- [x] Copy over salvageable documentation

Edit: I figure the following can be a separate PR, idea is configurability of responseOptions out so extension can iterate. Choose one of the following.

  - [x] **A**: Add a JSON/YAML constructor using marian-internals. This will need to be replaced with a faster parsing mechanism or direct access to the struct containing `ResponseOptions/TranslationRequest`. (#141)
  - [x] **B**: Provide struct TranslationRequest access directly to WASM (#140)
  
Unsure if both **A** and **B** can be retained.


wasm/bindings/TranslationRequestBindings.cpp
```diff
@@ -5,7 +5,9 @@
 
 #include <emscripten/bind.h>
 
-#include "TranslationRequest.h"
+#include "response_options.h"
+
+typedef marian::bergamot::ResponseOptions TranslationRequest;
```
XapaJIaMnu, line 10: Bad. Refactor cleanly.

XapaJIaMnu (changes requested):
No typedef hacks please. If we can't change it because we're breaking the plugin, negotiate the change with the plugin. Else don't change it.

wasm/bindings/TranslationRequestBindings.cpp
jerinphilip: I have to draw the line somewhere - if I have to make it clean I have to clean it in extension as well by the rationale. This is the cleanest it in my perspective, so the binding semantics below stay consistent (TranslationRequest -> "TranslationRequest") in JS. The plugin can do their part later? 

My objective here is to get the two different entities "TranslationRequest" | "ResponseOptions" into one of Service so incoming concurrency changes don't wonder why there exist two.

src/translator/service.h
```diff
@@ -137,13 +136,13 @@ class Service {
...
-  /// indicating whether or not to include some member in the Response, also
-  /// specify any additional configurable parameters.
+  /// @param [in] translationRequest: ResponseOptions indicating whether or not
+  /// to include some member in the Response, also specify any additional
+  /// configurable parameters.
 
   std::vector<Response>
   translateMultiple(std::vector<std::string> &&source,
-                    TranslationRequest translationRequest);
+                    ResponseOptions responseOptions);
```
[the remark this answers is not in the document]
jerinphilip: bools and enums only as of now, about 5. WASM previously was pass by value so I'm keeping the signature.

wasm/bindings/TranslationRequestBindings.cpp
abhi-agg: Created a separate issue https://github.com/browsermt/bergamot-translator/issues/149 for this.

abhi-agg:
> No typedef hacks please. If we can't change it because we're breaking the plugin, negotiate the change with the plugin. Else don't change it.

Created a separate issue https://github.com/browsermt/bergamot-translator/issues/149 to address clean refactoring

abhi-agg (approved):

XapaJIaMnu:
@abhi-agg before we approve this, can you confirm that we have agreed on the same name at #149 ?

XapaJIaMnu:
Ping @abhi-agg

abhi-agg:
> Ping @abhi-agg

@XapaJIaMnu I suggested few options there and we need to decide which one we opt for. I pinged everyone again. It is a decision that all of us need to take together 👍

XapaJIaMnu:
Jerin has proposed name here, I think that name is fine, if you agree I can merge.

abhi-agg:
> Jerin has proposed name here, I think that name is fine, if you agree I can merge.

@XapaJIaMnu Sorry but I am a bit lost here. I saw only 1 comment from Kenneth on the issue. Did I miss something?

XapaJIaMnu:
`ResponseOptions` got renamed to `TranslationRequest` other bits of the code should follow so we can remove the `typedef`

XapaJIaMnu (approved):
Ok, I guess let's start renaming. Let's get rid of the `typedef` ASAP. @abhi-agg please take the necessary steps to rename things in the other repos.

abhi-agg:
> `ResponseOptions` got renamed to `TranslationRequest` other bits of the code should follow so we can remove the `typedef`

@XapaJIaMnu Do you mean, `ResponseOptions` is renamed back to `TranslationRequest` in this PR? I don't see it here.

> Ok, I guess let's start renaming. Let's get rid of the `typedef` ASAP. @abhi-agg please take the necessary steps to rename things in the other repos.

I don't see a clear decision. Which [option](https://github.com/browsermt/bergamot-translator/issues/149#issue-890174034) did we decide on finally?

XapaJIaMnu:
TranslationRequest got renamed to response options

PR #140, opened 2021-05-09
WASM struct bindings for TranslationRequest

jerinphilip:
Looks like WASM compiled and no need of any parsing.


jerinphilip:
Abandoned, somebody else's problem now.

PR #141, opened 2021-05-09
Bindings constructor parse from YAML

jerinphilip:
Parse from YAML version, expensive right now due to allocations


kpu:
Oh no, the YAML form of request options keeps coming back.  Why??????

jerinphilip:
> Oh no, the YAML form of request options keeps coming back. Why??????

So, #139 is a no-YAML solution, you can let it in as-is. This PR (#141) is behind it. There's also an alternate of #140 behind #139. I gain some WASM exp++ even if this doesn't go in. Fair enough?

kpu (changes requested):
Blocking what is apparently yaml practice, then muting.

jerinphilip:
Abandoned.

PR #142, opened 2021-05-10
Change ResponseBuilder to accept callback instead of future

### LINKED ISSUE #126: Callback instead of future?
In @XapaJIaMnu 's use case there's a UI thread.  Currently we have a future and waiting on that future means blocking, which means another thread whose sole purpose is to wait on the future then notify the UI thread.  This is a bit silly and could be fixed with a callback.  
Mozilla isn't using the future now anyway (that sentence sounds funny) and didn't care if it was a callback IIRC.  
So we may as well change it to a callback.

jerinphilip:
Fixes #126.


XapaJIaMnu:
Ready for review or abandon?

jerinphilip:
Making batcher thread-safe is what I should do before this goes in ideally, correct me if I'm wrong. I'm working on this separately. This completes one part, but leaves it on the user to do the `.wait()`, which I'm doing moving the future code into a lambda passed as the callback for a hack to complete one part of it.

kpu:
Why is this blocking on making batcher thread safe?  I get that making batcher thread-safe is more important, but I don't understand how these depend on each other.

jerinphilip:
> Why is this blocking on making batcher thread safe?

Earlier state std::future/std::promise ensures the user gets sane error messages and waits until the promise sets the future to get a valid value. There's plenty room for errors when the callback change comes in on the client side, I should probably leave a note somewhere this is not thread-safe. 

Marking this as ready for review. Tests pass, given mozilla has only blocking workflow (so callbacks are triggered on same thread and calls are blocking). `translateLocally` if it uses the callback will need to be careful for a while.

src/translator/service.h
```diff
@@ -91,8 +91,9 @@ class Service {
   /// @param [in] responseOptions: Options indicating whether or not to include
   /// some member in the Response, also specify any additional configurable
   /// parameters.
-  std::future<Response> translate(std::string &&source,
-                                  ResponseOptions options = ResponseOptions());
+  void translate(std::string &&source,
```
XapaJIaMnu, line 94: Callback not in the documentation. 

Also, translate has a callback, but translateMultiple doesn't?

src/translator/service.cpp
```diff
@@ -56,51 +56,49 @@ Service::translateMultiple(std::vector<std::string> &&inputs,
   // We queue the individual Requests so they get compiled at batches to be
   // efficiently translated.
   std::vector<std::future<Response>> responseFutures;
-  for (auto &input : inputs) {
-    std::future<Response> inputResponse =
-        queueRequest(std::move(input), responseOptions);
-    responseFutures.push_back(std::move(inputResponse));
+  for (size_t i = 0;  i < inputs.size(); i++) {
```
XapaJIaMnu, line 59: I don't quite understand what is happening here. When translateMultiple is hit, you mimic the old behaviour using a callback under the hood?

XapaJIaMnu: When is translateMultiple being used? How about a callback be executed on a vector of Responses, as opposed to this?

XapaJIaMnu (changes requested):
Several places are missing documentation for the added callback parameter. Unfortunately I don't think they are visible from this window, you need to clone the project yourself.

translateMultiple's current solution is not pretty. Why not have a callback that acts on the vector of responses instead?

src/translator/service.h
jerinphilip: I'll break extension and bindings if I change translateMultiple (translateMultiple is aliased as translate for extension at bindings). I have not heard from extension for a while regarding their async strategy to be able to modify this into callback. 

/cc @abhi-agg @motin

XapaJIaMnu: We could provide a thin wrapper for them probably, but it's not ideal. Let's hear from mozilla ppl. I'd rather not put a temporary hack that would become permanent later.

jerinphilip: Fixed documentation, leaving open until extension resolves the other half.

src/translator/service.cpp
jerinphilip: The thing is @kpu has only a `blockIfWASM()`, while `--cpu-threads 0` previously supported a blocking workflow on even non-WASM. The bergamot-translator-app compiled here thus passed tests.

I have two options here:
1. Disable tests corresponding to blocking workflow on `--cpu-threads 0` on non-WASM. The code will work without issues.
2. Do some future trickery to get those tests to pass as well. This sequentially sets the promise and future in the `blockIfWASM()` case becoming moot.

Would you like me to go with 1? Only Mozilla uses `translateMultiple(...)`. This means the `std::promise/future` clutter here can go. I like to keep 2. It's a mess either way, because of a collapse of two different behaviours into the same class.

jerinphilip: @XapaJIaMnu's suggestions here require (i) negotiating with Mozilla to change signature (ii) Remove the adjustments changing brt, changing the run here also fighting even worse contraptions. 

Unless there are critical problems with going through an additional promise + future, I'm proposing to close this PR here. I'll be doing the test-adjustments restructuring things in a systematic way in a later PR if it's OK. 

All this PR does is change future to callback for https://github.com/browsermt/bergamot-translator/issues/126, which is complete.

app/marian-decoder-new.cpp
```diff
@@ -34,10 +34,16 @@ int main(int argc, char *argv[]) {
   std::string input = std_input.str();
   using marian::bergamot::Response;
 
-  // Wait on future until Response is complete
-  std::future<Response> responseFuture = service.translate(std::move(input));
+  marian::bergamot::ResponseOptions responseOptions;
```
qianqianzhu, line 37: `responseOptions` is not initialised but passed to `translate()`. Is that a better option that we use a default value for `responseOptions` in `translate()`? So we don't need `responseOptions` here.

kpu:
This needs to be talked about in the context of priorities and cancellation.  

Really want we want is something like

`Job translate(std::string &&src, std::function<void(Response &&)>, const ResponseOptions &options)`

Where Job is an object that allows cancellation and priority changes.

app/marian-decoder-new.cpp
jerinphilip: Removed `responseOptions`.

jerinphilip:
> `Job translate(std::string &&src, std::function<void(Response &&)>, const ResponseOptions &options)`
>
> Where Job is an object that allows cancellation and priority changes.  

We have something like the following lying around in some branch.

<details>
<summary>Details</summary>


```cpp

enum class JobState {
	  RUNNING,   ///< In pipeline to be scheduled in a batch, can be cancelled / amended
	  REJECTED,  ///< Service cannot take more jobs due to bounded memory constraints
	  COMPLETED, ///< Finished successfully
	  FAILED     ///< Failed due to whatever unforseeable error.
};

class Job {
public:
	  void wait(); // Wait until over barrier, do we still need after we decided no future?
	      
	  void cancel(){ service_.cancel(requestId_)};
	  void amend(size_t priority) { service_.amend(requestId, priority) };
	  
	  JobState state(){ return state_; };
	  float progress() { 
	  	return total_?
				(100.0f*completed_.load())/float(total_):
				0; 
	 }
private:
	  const Service &service_; ///< To inform cancel or amend to service.
	  size_t requestId_; ///< Identify the request associated with this Job / Tracker
	  
	  JobState state_{QUEUED};
	  
	  // The following can be updated by multiple workers.
	  std::atomic<size_t> completed_{0};   
	  void update(size_t sentences){ /* guard? atomic access? */ completed_+=sentences; }
	  
	  // Total units in translation (sentences?), to be set by Service, only once.
	  size_t total_{0};
	  
	  friend class Service; ///< Only Service can update state_, completed_;
};
```

</details>

@kpu Are you suggesting doing this in this PR?

kpu:
I think there does need to be a function that supports multiple text inputs as a single job.  For now though, we can kill translateMultiple on native.

src/translator/service.cpp
```diff
@@ -56,45 +56,45 @@ std::vector<Response> Service::translateMultiple(std::vector<std::string> &&inpu
...
   // efficiently translated.
   std::vector<std::future<Response>> responseFutures;
-  for (auto &input : inputs) {
-    std::future<Response> inputResponse = queueRequest(std::move(input), responseOptions);
-    responseFutures.push_back(std::move(inputResponse));
+  for (size_t i = 0; i < inputs.size(); i++) {
+    // https://stackoverflow.com/a/33437008/4565794
+    // @jerinphilip is not proud of this.
+    Ptr<std::promise<Response>> responsePromise = New<std::promise<Response>>();
+    responseFutures.push_back(std::move(responsePromise->get_future()));
```
kpu, line 63: Ugh, let's just not support translateMultiple on native.

jerinphilip: So going for translateMultiple only on WASM with ifdefs kills the only test which tests mozilla's codepath is intact and works (on native). I will have to skip this to get green -> implying minimal marian has no tests also the same for Mozilla code path. 

WASM workflow only checks for builds on CI, no tests for whether intended translation comes for a given input or not. I'd rather have test coverage for this codepath at the cost of having a promise + future + wait. 

I'm trying to work out a solution on #177. This branch is reset to the last state where all tests pass.

src/translator/service.cpp
```diff
@@ -56,45 +56,45 @@ std::vector<Response> Service::translateMultiple(std::vector<std::string> &&inpu
   // We queue the individual Requests so they get compiled at batches to be
   // efficiently translated.
   std::vector<std::future<Response>> responseFutures;
-  for (auto &input : inputs) {
-    std::future<Response> inputResponse = queueRequest(std::move(input), responseOptions);
-    responseFutures.push_back(std::move(inputResponse));
+  for (size_t i = 0; i < inputs.size(); i++) {
+    // https://stackoverflow.com/a/33437008/4565794
+    // @jerinphilip is not proud of this.
+    Ptr<std::promise<Response>> responsePromise = New<std::promise<Response>>();
```
kpu, line 62: The `std::vector<Response>` could be grown to the appropriate size then written to asynchronously by the callbacks without a lock.  Then use a semaphore (or condition variable) to wait for completion.  There's no need for `future` nor a need to fill them.

kpu: However, were you to make `translateMultiple` exist only for WASM, then there would be no need for this, so I suggest doing that first, provided all regression tests pass.

Issue #133, opened 2021-05-07
Refactor Service


qianqianzhu:
PR https://github.com/browsermt/bergamot-translator/pull/122 left some refactorings in `Service` to be done for better design (I will work on this issue):
  1. refactor `loadVocabularies()` and wrap `vector<Ptr<Vocab const>>` in a new container. So we can take `loadVocabularies()` out of `Service`. `loadVocabularies()` should not belong to `Service`.
  2. bundle all byte arrays in a new structure, like `MemoryBundle`.  As suggested by @jerinphilip:
  ```cpp
struct MemoryBundle {
	AlignedMemory model;
	AlignedMemory shortlist;
	std::vector<std::shared_ptr<AlignedMemory>> vocabs;
	AlignedMemory ssplitPrefixFile;
};
```
  3. refactor Service constructors to remove redundancies

2 and 3 will affect the WASM JS bindings. @abhi-agg need your input here. 

This issues is open for discussion, so feel free to leave any comments.

abhi-agg:
>     2. bundle all byte arrays in a new structure, like `MemoryBundle`.  As suggested by @jerinphilip:
> ```c++
> struct MemoryBundle {
>   AlignedMemory model;
>   AlignedMemory shortlist;
>   std::vector<std::shared_ptr<AlignedMemory>> vocabs;
>   AlignedMemory ssplitPrefixFile;
> };
> ```

I am fine with this.

>     3. refactor Service constructors to remove redundancies

In my opinion:

1. All the constructors using `Options` can be made private (as we didn't want to expose marian internals in public APIs) unless there is a good reason to keep them public. @jerinphilip might have something to add here.

2. For each category of constructors (i.e `Options` based and [string](https://github.com/browsermt/bergamot-translator/blob/main/src/translator/service.h#L97) based), we should have 2 separate constructors: one for loading only by bytes and another . Something like this below:
    1. One that loads only by files. e.g. `Service(const std::string &config)`
        1. It will call `parseOptions(config, /*validate=*/true)` because we would want to validate the model and vocab file names etc.
    2. Another that loads only by bytes. e.g. `Service(const std::string &config, MemoryBundle memoryBundle)`
        1. It will call `parseOptions(config, /*validate=*/false)`

jerinphilip:
> 1. All the constructors using Options can be made private... unless there is a good reason to keep them public

[...]

PR #143, opened 2021-05-11
Refactor vocabs in Service

qianqianzhu:
This PR address https://github.com/browsermt/bergamot-translator/issues/133 (1) which aims to take loadVocabularies() out of Service. 
Introduce `vocabs.h` to include `loadVocabularies()` functionalities.

### DIFF
--- a/src/translator/batch_translator.cpp
+++ b/src/translator/batch_translator.cpp
@@ -22,16 +22,16 @@ void BatchTranslator::initialize() {
   bool check = options_->get<bool>("check-bytearray",false); // Flag holds whether validate the bytearray (model and shortlist)
   if (options_->hasAndNotEmpty("shortlist")) {
     int srcIdx = 0, trgIdx = 1;
-    bool shared_vcb = vocabs_->source().front() == vocabs_->target();
+    bool shared_vcb = vocabs_->sources().front() == vocabs_->target(); // vocabs_->sources().front() is invoked as we currently only support one source vocab
     if (shortlistMemory_->size() > 0 && shortlistMemory_->begin() != nullptr) {
       slgen_ = New<data::BinaryShortlistGenerator>(shortlistMemory_->begin(), shortlistMemory_->size(),
-                                                   vocabs_->source().front(), vocabs_->target(),
-                                                     srcIdx, trgIdx, shared_vcb, check);
+                                                   vocabs_->sources().front(), vocabs_->target(),
+                                                   srcIdx, trgIdx, shared_vcb, check);
     }
     else {
       // Changed to BinaryShortlistGenerator to enable loading binary shortlist file
       // This class also supports text shortlist file
-      slgen_ = New<data::BinaryShortlistGenerator>(options_, vocabs_->source().front(),
+      slgen_ = New<data::BinaryShortlistGenerator>(options_, vocabs_->sources().front(),
                                                     vocabs_->target(), srcIdx,
                                                     trgIdx, shared_vcb);
     }
@@ -97,7 +97,7 @@ void BatchTranslator::translate(Batch &batch) {
   std::vector<Ptr<SubBatch>> subBatches;
   for (size_t j = 0; j < maxDims.size(); ++j) {
     subBatches.emplace_back(
-        New<SubBatch>(batchSize, maxDims[j], vocabs_->source().at(j)));
+        New<SubBatch>(batchSize, maxDims[j], vocabs_->sources().at(j)));
   }
 
   std::vector<size_t> words(maxDims.size(), 0);
--- a/src/translator/batch_translator.cpp
+++ b/src/translator/batch_translator.cpp
@@ -10,7 +10,7 @@ namespace marian {
 namespace bergamot {
 
 BatchTranslator::BatchTranslator(DeviceId const device,
-                                 std::vector<Ptr<Vocab const>> &vocabs,
+                                 Vocabs &vocabs,
                                  Ptr<Options> options,
                                  const AlignedMemory* modelMemory,
                                  const AlignedMemory* shortlistMemory)
@@ -22,17 +22,17 @@ void BatchTranslator::initialize() {
   bool check = options_->get<bool>("check-bytearray",false); // Flag holds whether validate the bytearray (model and shortlist)
   if (options_->hasAndNotEmpty("shortlist")) {
     int srcIdx = 0, trgIdx = 1;
-    bool shared_vcb = vocabs_->front() == vocabs_->back();
+    bool shared_vcb = vocabs_->source().front() == vocabs_->target();
     if (shortlistMemory_->size() > 0 && shortlistMemory_->begin() != nullptr) {
       slgen_ = New<data::BinaryShortlistGenerator>(shortlistMemory_->begin(), shortlistMemory_->size(),
-                                                     vocabs_->front(), vocabs_->back(),
+                                                   vocabs_->source().front(), vocabs_->target(),
                                                      srcIdx, trgIdx, shared_vcb, check);
     }
     else {
       // Changed to BinaryShortlistGenerator to enable loading binary shortlist file
       // This class also supports text shortlist file
-      slgen_ = New<data::BinaryShortlistGenerator>(options_, vocabs_->front(),
-                                                    vocabs_->back(), srcIdx,
+      slgen_ = New<data::BinaryShortlistGenerator>(options_, vocabs_->source().front(),
+                                                    vocabs_->target(), srcIdx,
                                                     trgIdx, shared_vcb);
     }
   }
@@ -97,7 +97,7 @@ void BatchTranslator::translate(Batch &batch) {
   std::vector<Ptr<SubBatch>> subBatches;
   for (size_t j = 0; j < maxDims.size(); ++j) {
     subBatches.emplace_back(
-        New<SubBatch>(batchSize, maxDims[j], vocabs_->at(j)));
+        New<SubBatch>(batchSize, maxDims[j], vocabs_->source().at(j)));
   }
 
   std::vector<size_t> words(maxDims.size(), 0);
@@ -117,7 +117,7 @@ void BatchTranslator::translate(Batch &batch) {
   auto corpus_batch = Ptr<CorpusBatch>(new CorpusBatch(subBatches));
   corpus_batch->setSentenceIds(sentenceIds);
 
-  auto trgVocab = vocabs_->back();
+  auto trgVocab = vocabs_->target();
   auto search = New<BeamSearch>(options_, scorers_, trgVocab);
 
   auto histories = std::move(search->search(graph_, corpus_batch));
--- a/src/translator/batch_translator.cpp
+++ b/src/translator/batch_translator.cpp
@@ -116,9 +116,8 @@ void BatchTranslator::translate(Batch &batch) {
 
   auto corpus_batch = Ptr<CorpusBatch>(new CorpusBatch(subBatches));
   corpus_batch->setSentenceIds(sentenceIds);
-
-  auto trgVocab = vocabs_->target();
-  auto search = New<BeamSearch>(options_, scorers_, trgVocab);
+  
+  auto search = New<BeamSearch>(options_, scorers_, vocabs_->target());
 
   auto histories = std::move(search->search(graph_, corpus_batch));
   batch.completeBatch(histories);
--- a/src/translator/batch_translator.cpp
+++ b/src/translator/batch_translator.cpp
@@ -14,7 +14,7 @@ BatchTranslator::BatchTranslator(DeviceId const device,
                                  Ptr<Options> options,
                                  const AlignedMemory* modelMemory,
                                  const AlignedMemory* shortlistMemory)
-    : device_(device), options_(options), vocabs_(&vocabs),
+    : device_(device), options_(options), vocabs_(vocabs),
     modelMemory_(modelMemory), shortlistMemory_(shortlistMemory) {}
 
 void BatchTranslator::initialize() {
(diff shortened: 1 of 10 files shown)


XapaJIaMnu (dismissed):

src/translator/vocabs.h
```diff
@@ -0,0 +1,68 @@
+#pragma once
+
+namespace marian {
+namespace bergamot {
+
+class VocabsGenerator {
```
jerinphilip, line 6: This class can be `Vocabs`.

src/translator/vocabs.h
```diff
@@ -0,0 +1,68 @@
...
+public:
+  VocabsGenerator(Ptr<Options> options): options_(options){}
```
jerinphilip, line 8: Should be aware of only Filenames / AlignedMemory and not aware of `Ptr<Options>` - which are two modes of construction.

src/translator/vocabs.h
```diff
@@ -0,0 +1,68 @@
...
+    std::unordered_map<std::string, Ptr<Vocab>> vmap;
+    for (size_t i = 0; i < vocabs_.size(); ++i) {
+      auto m = vmap.emplace(std::make_pair(vocabPaths[i], Ptr<Vocab>()));
+      if (m.second) { // new: load the vocab
+        m.first->second = New<Vocab>(options_, i);
+        m.first->second->load(vocabPaths[i]);
+      }
+      vocabs_[i] = m.first->second;
+    }
+  }
```
jerinphilip, line 43: load corresponds to two modes of construction overloaded by arguments - filenames / ByteArrays.

src/translator/vocabs.h
```diff
@@ -0,0 +1,68 @@
...
+  // load from file
+  void load(const std::vector<std::string>& vocabPaths){
+    // with the current setup, we need at least two vocabs: src and trg
+    ABORT_IF(vocabPaths.size() < 2, "Insufficient number of vocabularies.");
+    vocabs_.resize(vocabPaths.size());
+    std::unordered_map<std::string, Ptr<Vocab>> vmap;
+    for (size_t i = 0; i < vocabs_.size(); ++i) {
+      auto m = vmap.emplace(std::make_pair(vocabPaths[i], Ptr<Vocab>()));
+      if (m.second) { // new: load the vocab
+        m.first->second = New<Vocab>(options_, i);
```
jerinphilip, line 38: Ah, you need options. Never mind, my bad.

src/translator/vocabs.h
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  std::vector<Ptr<Vocab const>> getVocabs(){
+    return vocabs_;
+  }
+
+private:
+  std::vector<Ptr<Vocab const>> vocabs_;
+  Ptr<Options> options_;
+};
```
jerinphilip, line 52: This class can have two methods - `source()` and `target()` which provides me source and target vocabulary. At places, I'm using the following syntax. I'd rather this knowledge be contained within this class.

```cpp
srcVocab = vocabs.front()
tgtVocab = vocabs.back()
```

This could be contained here.

XapaJIaMnu: This is bad practise in general, as you could have multiple source vocabularies. We don't technically support it, as we assume only 2 vocabularies, but this is not necessarily true!

src/translator/vocabs.h
XapaJIaMnu: Generator is more appropriate name as it generates the vocab class.

src/translator/vocabs.h
qianqianzhu: Can you be more specific here?

XapaJIaMnu: As it should? It's typical to overload a function when it does the same thing (Eg constructing a Vocabulary either from a bytes or from path)

src/translator/vocabs.h
jerinphilip: I understand what you're saying. But interspersing use of `vocabs.front()`, `vocabs[0]`, `vocabs.back()`, `vocabs[1]` in code is bad practice. If I want source, I should be calling `vocabs.source()`. If I want `target`, I should be calling `vocabs.target()`. 

If I have multiple sources or targets, I attach size_t idx or maybe another method to specify the i-th source-vocabulary or the i-th target vocabulary. Giving me a list of Vocabs having to go look 3 files down to figure out which one's which is bad practice. The `Vocabs` class can vary it's internals while the external API stays (maybe) stable and the rest of the wiring in marian::bergamot works irrespective of how if multiple source-vocabs or target-vocabs come in.

XapaJIaMnu: Vocas source in this case should return a vector and target could return only one vocab. Are you ok with this?

src/translator/vocabs.h
jerinphilip: What I'm suggesting is a change to a multi-Vocabs container with knowledge of which ones are source vocabularies and which ones are target vocabularies. Remember, marian has `Vocab` and `vector<Ptr<Vocab const>>`. I want to replace `vector<Ptr<Vocab const>>` with this class.

src/translator/vocabs.h
jerinphilip: > Vocabs source in this case should return a vector and target could return only one vocab. Are you ok with this?

Ideally not. For text-proc, at places I'm only using the first-vocab. And somewhere in PR review, it's mentioned that all the vocabularies Mozilla encounters will have only two vocabs (source and target, which can be same). I'm not saying the following should be implemented, but to communicate with a loose sketch:

```
Vocabs::source() -> Only one source vocab
Vocabs::multiSource(size_t idx) -> idx-th source vocab
Vocabs::multiSources() -> All source vocabs as a vector (This is what you're suggesting)
Vocabs::target() -> Only one target vocab
Vocabs::multiTargets() -> All target-vocabs as a vector
```

The above can still work with `vector<Ptr<Vocab const>>` under the hood, except the documentations/assumptions baked into code rather than convention like first N being source and last 1 being target.

Far as I understand, the multi-source is a training only setup I'm unsure how it can be used here (bergamot/marian-server is pure inference).

XapaJIaMnu: Multi source can happen when you decode with ensembles. Every model in an ensemble could have a different source vocab, provided they all have the same target vocab. You may only be using the first vocab in text processing, but that means we just don't support ensembles at all. 

You can only ever have one Target  vocabulary during decoding and multiple Source vocabularies (someone correct me if I am wrong) In general MT practitioners are aware of that.

if we say that we never ever support ensembles, then you can make the front() and back() assumptions.

src/translator/vocabs.h
XapaJIaMnu: The reason marian has those two is because you can have multiple source and one target vocabularies as discussed elsewhere.

src/translator/vocabs.h
qianqianzhu: Let's keep `vector<Ptr<Vocab const>>`. What @jerinphilip suggested is good for the readability, but it might be too early to settle. We should allow multiple Source vocabularies for now not to break what Marian supports. (I will add documentation for `vector<Ptr<Vocab const>>`)

jerinphilip: > Vocas source in this case should return a vector

After an offline conversation, I'm ok with this approach, having made aware of https://github.com/marian-nmt/marian-dev/blob/fe74576dc3e957f97b4842951fb6ee7ec52fa324/src/translator/translator.h#L242-L243 and https://github.com/marian-nmt/marian-dev/blob/fe74576dc3e957f97b4842951fb6ee7ec52fa324/src/translator/translator.h#L303-L304

The call-site of `vocabs.front()` at TextProcessor (sentencepiece encodeWithByteRanges) will eventually need to be updated with multiple vocabs if and when we're ensembling. For now it can be `vocabs.source().front()` with a comment describing the situation.

jerinphilip (changes requested):
If we're getting a class (looking at `VocabsGenerator`), we need the methods/ops decoupled from current call-sites in Service/sub components where we operate with `vector<Ptr<Vocab const>>` which doesn't encode information on which one is source or target.

`VocabsGenerator` can easily be a `Vocabs` class wrapping around `vector<Ptr<Vocab const>>` with what is currently `load` being two constructors.

src/translator/vocabs.h
```diff
@@ -0,0 +1,81 @@
...
+      // load vocabs from file
+      auto vocabPaths = options->get<std::vector<std::string>>("vocabs");
+      load(vocabPaths);
+    }
+  }
+
+  /// Get all source vocabularies (as a vector)
+  std::vector<Ptr<Vocab const>>& source(){
+    return srcVocabs_;
+  }
```
jerinphilip, line 26: nit perhaps - `sources()` instead of `source()`. We don't want these (`sources()`, `target()`) to be edited outside this class. Can we get away with making these `const` references?

src/translator/response_builder.cpp
```diff
@@ -65,7 +65,7 @@ void ResponseBuilder::buildTranslatedText(Histories &histories,
 
     Result result = onebest[0]; // Expecting only one result;
     Words words = std::get<0>(result);
-    auto targetVocab = vocabs_->back();
+    auto targetVocab = vocabs_->target();
```
jerinphilip, line 68: Statements of these nature are now redundant, part of the intent of the refactor was to directly pass `vocabs_->target()` at call-site than temporarily assign to variable to describe better.

src/translator/batch_translator.cpp
```diff
@@ -22,17 +22,17 @@ void BatchTranslator::initialize() {
...
+                                                   vocabs_->source().front(), vocabs_->target(),
                                                      srcIdx, trgIdx, shared_vcb, check);
     }
     else {
       // Changed to BinaryShortlistGenerator to enable loading binary shortlist file
       // This class also supports text shortlist file
-      slgen_ = New<data::BinaryShortlistGenerator>(options_, vocabs_->front(),
-                                                    vocabs_->back(), srcIdx,
+      slgen_ = New<data::BinaryShortlistGenerator>(options_, vocabs_->source().front(),
+                                                    vocabs_->target(), srcIdx,
```
jerinphilip, line 35: @XapaJIaMnu Can you shed some light here? `ShortlistGenerator` seem to be operating  with only one source-vocab. What becomes in the ensembling case? Is there an advantage to providing an alias through a member function to only the first source?

qianqianzhu: I added comments for why we invoke `vocabs_->source().front()`(because we currently only use one source vocab).

src/translator/batch_translator.cpp
```diff
@@ -117,7 +117,7 @@ void BatchTranslator::translate(Batch &batch) {
   auto corpus_batch = Ptr<CorpusBatch>(new CorpusBatch(subBatches));
   corpus_batch->setSentenceIds(sentenceIds);
 
-  auto trgVocab = vocabs_->back();
+  auto trgVocab = vocabs_->target();
```
jerinphilip, line 120: redundant variable assignment. conveyed through `target()`.

src/translator/definitions.h
```diff
@@ -34,14 +34,14 @@ struct MemoryBundle {
...
   }
 
   MemoryBundle &operator=(MemoryBundle &&from) {
     model = std::move(from.model);
     shortlist = std::move(from.shortlist);
-    vocabs = std::move(vocabs);
+    vocabs = std::move(from.vocabs);
     ssplitPrefixFile = std::move(from.ssplitPrefixFile);
     return *this;
   }
```
jerinphilip, line 47: Can you check if the compiler assigned move and copy-constructors/assignments are good enough? All members are `AlignedMemory` which are only move-ctor/assign-able. Trivially generated ones should be okay?

This is more room for error.

src/translator/batch_translator.cpp
jerinphilip: I'd want to hide away the `.front()` tail (access skipping `Vocabs` layer) by an alias (`vocabs_.source()` or `vocabs_.source_first()`) which gives just the first vocab (so it's easier to edit later within the `Vocab` class). This skipped access violates the attempt to freely vary internals of `Vocabs` while outside wiring remains functional. Hence asking Nick or someone to confirm how `ShortListGenerator` works with ensembling.

src/translator/service.cpp
```diff
@@ -5,45 +5,12 @@
...
-  return vocabs;
-}
-
 namespace marian {
 namespace bergamot {
 
 Service::Service(Ptr<Options> options, MemoryBundle memoryBundle)
     : requestId_(0), options_(options),
-      vocabs_(std::move(loadVocabularies(options, std::move(memoryBundle.vocabs)))),
+      vocabs_(Vocabs(options, std::move(memoryBundle.vocabs))),
```
jerinphilip, line 13: You can drop the `Vocabs` here, yes and use member initialization directly?

qianqianzhu: What do you mean?

jerinphilip: `vocabs_(options, std::move(memoryBundle.vocabs))` instead of `vocabs_(Vocabs(...))`.

src/translator/batch_translator.cpp
XapaJIaMnu: As you can see from the beamsearch, each scorer could have its own shortlist. https://github.com/marian-nmt/marian-dev/blob/fe74576dc3e957f97b4842951fb6ee7ec52fa324/src/translator/beam_search.cpp#L434

However in practise when decoding with ensembles you don't use a shortlist.

Don't do `source_first()` method. Just don't. Don't use such aliases please, they are not making the code more readable or simpler to understand.

src/translator/response_builder.h
```diff
@@ -81,7 +82,7 @@ class ResponseBuilder {
   // Data members are context/curried args for the functor.
 
   ResponseOptions responseOptions_;
-  std::vector<Ptr<Vocab const>> *vocabs_; // vocabs are required for decoding
+  Vocabs *vocabs_; // vocabs are required for decoding
```
XapaJIaMnu, line 85: Minor && sorry to bother, but... Is there any reason why a `unique_ptr` is not used here?

src/translator/batch_translator.cpp
jerinphilip: That clarifies enough not to keep a first on the shortlist. 

@qianqianzhu we're looking at a vector like init on shortlist similar to the model and zipping both to set respective shortlist on each scorer, looking at batch translator, eventually? That will get rid of the first() in BatchTranslator. 

As for preprocessing, next place with shortlist, I think I'd probably have to rework a lot I guess. 

Marking this as resolved for purposes of this PR.

src/translator/response_builder.h
jerinphilip: Lifetime/ownership is handled at Service. This is a reference alive so long as Service is.

XapaJIaMnu: In this case you need to use `Vocab& vocabs_` or to be more precise `const Vocab& vocabs_`. Using raw pointers in C++ in 2021 is only really justifiable if you are implementing some very low level datastructures. I have refactored the code.

XapaJIaMnu: (Correct me if I am wrong)

src/translator/service.cpp
qianqianzhu: My bad. Fixed.

src/translator/response_builder.h
qianqianzhu: @XapaJIaMnu Thanks for refactoring. We certainly do not want users to change the vocabs members outside this class.

XapaJIaMnu: Ok, seriously, we use Vocab * in a lot of places, I'm really not sure why, This would warrant further refactoring. Can someone explain to me the logic behind it. Is it because we want to move objects a lot and we want full move semantics?

jerinphilip: @XapaJIaMnu Thanks. I didn't know of this syntax (most C++ I learned was pre 2011). That's why there's Vocab* in places. I'll update the rest in a following PR.

qianqianzhu: @jerinphilip @XapaJIaMnu  I will do the rest refactorings by replacing `Vocabs *` to `const Vocabs&`.

kpu: In Google style it's & for const and * for mutable but I tend to just use & for everything.

jerinphilip (approved):

XapaJIaMnu (approved):

PR #148, opened 2021-05-12
Remove packaging step from wasm build commands

abhi-agg:
Model, shortlist and vocabulary files can be passed as bytes. => We can stop packaging these files into wasm during build steps.

This PR:
- Updates READMEs to reflect these changes
- Updates CMakeLists.txt files to remove packaging steps from them
- Updates wasm README to reflect API changes to pass vocab as bytes

### DIFF
--- a/wasm/CMakeLists.txt
+++ b/wasm/CMakeLists.txt
@@ -14,14 +14,7 @@ target_include_directories(bergamot-translator-worker
 target_compile_definitions(bergamot-translator-worker PRIVATE WASM_BINDINGS)
 target_compile_options(bergamot-translator-worker PRIVATE ${WASM_COMPILE_FLAGS})
 
-set(LINKER_FLAGS "-g2 --bind -s ASSERTIONS=0 -s DISABLE_EXCEPTION_CATCHING=1 -s FORCE_FILESYSTEM=1 -s ALLOW_MEMORY_GROWTH=1 -s NO_DYNAMIC_EXECUTION=1 -s EXPORTED_RUNTIME_METHODS=[addOnPreMain]")
-if (NOT PACKAGE_DIR STREQUAL "")
-  get_filename_component(REALPATH_PACKAGE_DIR ${PACKAGE_DIR} REALPATH BASE_DIR ${CMAKE_BINARY_DIR})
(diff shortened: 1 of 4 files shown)


wasm/CMakeLists.txt
```diff
@@ -15,13 +15,6 @@ target_compile_definitions(bergamot-translator-worker PRIVATE WASM_BINDINGS)
 target_compile_options(bergamot-translator-worker PRIVATE ${WASM_COMPILE_FLAGS})
 
 set(LINKER_FLAGS "-g2 --bind -s ASSERTIONS=0 -s DISABLE_EXCEPTION_CATCHING=1 -s FORCE_FILESYSTEM=1 -s ALLOW_MEMORY_GROWTH=1 -s NO_DYNAMIC_EXECUTION=1 -s EXPORTED_RUNTIME_METHODS=[addOnPreMain]")
```
motin, line 17: Do we still need FORCE_FILESYSTEM=1?

wasm/README.md
```diff
@@ -18,10 +18,7 @@ gunzip models/*/*
 ```js
 // The model configuration as YAML formatted string. For available configuration options, please check: https://marian-nmt.github.io/docs/cmd/marian-decoder/
 // This example captures some of the most relevant options
-const modelConfig = `vocabs:
-  - /esen/vocab.esen.spm
-  - /esen/vocab.esen.spm
-beam-size: 1
+const modelConfig = `beam-size: 1
```
jerinphilip, line 21: Can't this be an `example.js` loaded and linked here? Why do you have two copies.. I remember reviewing the same yesterday.

jerinphilip (dismissed):

wasm/README.md
abhi-agg: The goal here is to provide an idea on how to use the JS APIs. The code that you had reviewed yesterday is a fully functional wasm test page :)

wasm/CMakeLists.txt
abhi-agg: Actually no. I will remove this as well 👍

abhi-agg: @motin Done 👍

motin (approved):

jerinphilip (approved):

PR #151, opened 2021-05-14
Adding clang-format and updating existing sources to adhere

jerinphilip:
A workflow and associated files to make the sources here consistent with a defined style-guide. For demonstration, `BasedOnStyle: Google` is put in place, but I have no strong opinions that it should stick.

@XapaJIaMnu, @kpu Let know if there is some authentic "marian" style file which I can use, I couldn't find any. I found an [autoformat.sh](https://github.com/browsermt/marian-dev/blob/master/contrib/autoformat.sh) among source files which I think is using the default LLVM based style.


XapaJIaMnu (changes requested):
Several issues here:

1) The clang-format requests changes to the javascript bindings in a way that is not very intuitive for people used to writing JS: https://github.com/browsermt/bergamot-translator/pull/151/checks?check_run_id=2584634373#step:4:1657 This change is not appropriate.
2) Maximum line length is too low even for 1080p monitor: https://github.com/browsermt/bergamot-translator/pull/151/checks?check_run_id=2584634373#step:4:1619
Google style sheet uses 100, I personally would like 120, but this tool seems to recommend 80, judging by the suggested corrections.

3) Before this PR gets in, you should also apply the recommended changes, but we can leave that for after we agree on a format.


Also, I think that it is a good idea to have a consistent format, so thank you for that!

jerinphilip:
1. I understand this as ignore wasm directory when running style-checker and is implementable. 
2. I understand this to be `ColumnLimit: 120`. Will add it in the file. 
3. I'll run the formatter once the .clang-format + ignores are approved.

Requesting other devs to leave additional comments / debate any of the above choices if they have and strong opinions in a day or two, after which the formatting dictated by the files brought by this PR will stick. 

/cc @kpu @abhi-agg @qianqianzhu

abhi-agg:
>     1. I understand this as ignore wasm directory when running style-checker and is implementable.

I am also fine with running the formatter for `.cpp` and `.h` files in `wasm` folder and ignoring rest of the files there.

Btw, will you be applying the formatter in the same PR?

jerinphilip:
> I am also fine with running the formatter for .cpp and .h files in wasm folder.

I read that people who edit the CPP over there will prefer consistency to inconsistency for javascript appropriate styles, therefore wasm will have clang-format run over. Default extensions are only cpp in the python script, so ignoring non CPP files are already covered.

https://github.com/Sarcasm/run-clang-format/blob/39081c9c42768ab5e8321127a7494ad1647c6a2f/run-clang-format.py#L34

XapaJIaMnu:
This looks good. We need to apply the edits first, which will upset other pending PRs, but oh well.. The other option is to have a simple beautifier to help other people fix their format in the ongoing PRs

qianqianzhu (commented):
Looks nice. Many thanks. So after this, our new PR will automatically apply this code format once submitted, right? What should we do if we want to apply this code format in our ongoing PR (not submitted yet)? Run run-clang-format.py?

jerinphilip:
@qianqianzhu There's no autoapply, although it's possible to do this through actions. The CI is only set to complain of not adhering to format, the authors are supposed to apply the command through clang-format or git-clang-format or the python wrapper provided here to get the test to pass before merge. If you have your editor configured to autoformat on write buffer that should work as well.

qianqianzhu:
It makes sense. Autoapply shouldn't happen after merge.
> The authors are supposed to apply the command through clang-format or git-clang-format or the python wrapper provided here to get the test to pass before merge.

Shall we put this statement somewhere in this repo? Just in case other newcomers know they must use our code style.

jerinphilip:
> Shall we put this statement somewhere in this repo? Just in case other newcomers know they must use our code style.

I think this sorta thing goes into a CONTRIBUTING.md or [docs/marian-integration.md] or maybe a pull request template. Perhaps we can look into this eventually when we gain more critical developer mass which justifies it? Right now, the people who send in PRs are already reminded when CI on pull-request to main fail on coding style checks to update with the tool.

XapaJIaMnu:
Add a note in somewhere in the readme, I'm not really fussed. We should have a standalone tool to run and verify your code that is not a github workflow action, so that we don't depend on them.

doc/marian-integration.md
```diff
@@ -93,3 +93,18 @@ ARGS=(
 </summary>
 </details>
 
+## Coding Style
+
+This repository contains C++ and JS source-files, of which C++ should adhere to
+the clang-format based style guidelines. You may configure your development
+environment to use the `.clang-format` and `.clang-format-ignore` files with
```
qianqianzhu, line 100: Very minor. We shall mention `.clang-format` and `.clang-format-ignore` files are provided in the root directory (or if links if you prefer).

qianqianzhu (approved):

XapaJIaMnu (dismissed):

PR #152, opened 2021-05-16
Use binary lexical shortlist in documentation

kpu:


kpu:
@abarbosa94 Does this work for you?

XapaJIaMnu:
Waiting for a confirmation before it works, then merging.

XapaJIaMnu (dismissed):

abarbosa94:
@kpu yes!

I'd also recommend adding the CPU version-only instructions that are shown in [marian docs](https://marian-nmt.github.io/docs/)

```
CPU version
Marian CPU version requires Intel MKL or OpenBLAS. Both are free, but MKL is not open-sourced. Intel MKL is strongly recommended as it is faster. On Ubuntu 16.04 and newer it can be installed from the APT repositories:

wget -qO- 'https://apt.repos.intel.com/intel-gpg-keys/GPG-PUB-KEY-INTEL-SW-PRODUCTS-2019.PUB' | sudo apt-key add -
sudo sh -c 'echo deb https://apt.repos.intel.com/mkl all main > /etc/apt/sources.list.d/intel-mkl.list'
sudo apt-get update
sudo apt-get install intel-mkl-64bit-2020.0-088
For more details see the official instructions.

A CPU build needs to be enabled by adding -DCOMPILE_CPU=on to the CMake command:

cd /path/to/marian-dev
mkdir -p build
cd build
cmake .. -DCOMPILE_CPU=on
make -j4
```

But it is up to you to decide if this makes sense to include :)

XapaJIaMnu:
@abarbosa94 the CPU build is enabled by default. You do however need to disable the GPU build. I will edit and merge.

Issue #56, opened 2021-03-18
QualityEstimation Implementation [RFC]


jerinphilip:
**Input**: translated text, Source text, model scores for tokens, tokenization information to make sense of model scores.

**Output** is expected to be containing for each sentence the following:

 * sentence level quality score: `float`
* Word level quality score `vector<float>` corresponding to each of the Word

where Word is space separated words of a sentence  (mozilla prefers word, not subword level scores). Continuous values preferred for more experimentation capabilities.

Let output be a struct called `QualityEstimate`. Implementation which can start from the below skeleton is tentatively going to be used by `Service` to make `QualityEstimate` a member in `Response`. (The layer above in UnifiedAPI doesn't have access to logprobs, so).

```cpp
class QualityEstimator {
public:
  QualityEstimator(Args…) {
       // Use constructor to load an initialize any trained models
       // This is where I expect you to load any neural nets into a graph 
       // or prepare the model parameters (logistic regression) or something if you're using simpler.
  }
  QualityEstimate quality(Histories &histories, Response &response) {
      // AnnotatedText has the blob of text and sentence/word-token information which should be extracted.
      // modelScores are logprobs, they're accessed and ready.
      … your calculation code here
  }
};
```

This is to be built native first, and when readied exported to WASM.

@ugermann @abhi-agg @fredblain @mfomicheva /cc @kpu

jerinphilip:
Now adding the implementation details, to conveniently prototype (as @ugermann mentioned in the meeting) the class called `QualityEstimator`, you can already see that `Response` [already contains](https://github.com/browsermt/bergamot-translator/blob/a8dbb00a984030ebd039cfa9ffb9e3533ce00f31/src/translator/response.h#L75-L78) most of the information you suggested as inputs to quality estimation. Catch is you will have to work on top a pending alignments PR (this is the only place with the logprobs and target sentencepiece annotations available which is required per input specifications).

There are some ~qualityscores~ modelscores already being printed at https://github.com/browsermt/bergamot-translator/blob/a8dbb00a984030ebd039cfa9ffb9e3533ce00f31/app/main-mts.cpp#L59-L70

The output looks like this, for a sample text:
```
Quality: whole(547.219), tokens below:
[...]

Issue #113, opened 2021-04-27
Validating AlignedMemory passed via Service constructor


abhi-agg:
Please correct me if I am wrong but currently the [Service constructor that uses bytes for files](https://github.com/browsermt/bergamot-translator/blob/main/src/translator/service.h#L54-L55) doesn't perform:
1. Validation of the shortlist bytes
2. Validation of the model bytes

The [validation code](https://github.com/browsermt/bergamot-translator/blob/main/src/translator/byte_array_util.cpp#L30) is already present in the repo but is being invoked only in the [apps](https://github.com/browsermt/bergamot-translator/blob/main/app/service-cli-bytearray.cpp#L19-L20).

@kpu @XapaJIaMnu @jerinphilip In my opinion, this validation should also be invoked from within the [Service constructor that uses bytes for files](https://github.com/browsermt/bergamot-translator/blob/main/src/translator/service.h#L54-L55).
cc @andrenatal

jerinphilip:
> check increases shortlist loading time because it has to go over all indices and check they are in range. I specifically asked for it to be optional. And honestly it should be. Mozilla can later go in and check once if they want to, and care about loading time, which they seem to.
>
> I think of the check here as more for untrusted input.


https://github.com/browsermt/bergamot-translator/pull/69#discussion_r600544180

XapaJIaMnu:
The binary model validation code is cheaper than the shortlist one. Can you measure its footprint in WASM just in case, before we commit to running it every time?

abhi-agg:
The comment https://github.com/browsermt/bergamot-translator/pull/69#discussion_r600544180 and what @XapaJIaMnu wrote

> The binary model validation code is cheaper than the shortlist one.

explains to me why we are not doing validation for shortlist file.

But, I am confused regarding the default behavior as per https://github.com/browsermt/bergamot-translator/pull/69#discussion_r600544180 and https://github.com/browsermt/bergamot-translator/pull/69#discussion_r600550911.

1. What I understood is that **by default**, the validation of both shortlist and model bytes is **NOT** happening when [instantiating Service class with byte loading constructor](https://github.com/browsermt/bergamot-translator/blob/main/src/translator/service.h#L84-L86). Right?

> Mozilla can later go in and check once if they want to

2. If `1` is correct, then the how to perform this check once?

Looping in @qianqianzhu as she is the implementer 👍

jerinphilip:
[...]

PR #154, opened 2021-05-17
Fix failures when loading text shortlist

qianqianzhu:
This PR provides a quick patch to fix failures when loading text shortlist as mentioned in https://github.com/browsermt/bergamot-translator/issues/56#issuecomment-841855415.

`service-cli` now depends on `--check-bytearray` option to decide whether to construct `Service` from file-based loading or preloaded bytearrays. The bytearray of shortlist can only be loaded from binary shortlist. 
`--check-bytearray` option is set to true by default (assuming construct `Service` from bytearrays), thus causes failure when users provide text shortlist without setting `--check-bytearray` to false.

The `BinaryShortlistGenerator` itself did not break loading from text shortlist files. But this class can be further improved by allowing construct bytearrays from text shortlist (it is already in the code but I didn't provide the API).

Let me know what works best to fix this issue, @kpu @jerinphilip @XapaJIaMnu


XapaJIaMnu:
@qianqianzhu I think we had the same discussion on another PR and we decided to keep it on by default due to wanting users to verify it. Is it possible to somehow make it conditional based on the format of the shortlist file (eg if you have a shortlist file ending on .bin, verify, else don't.)

jerinphilip:
> Is it possible to somehow make it conditional based on the format of the shortlist file? 

Does marian's configparser allow this? CMake does which kpu did recently for a PR, I was looking for an equivalent heree. The alternative was the hard-code where I set these off when I don't have modelBytes, which you weren't particularly in favour of. I think this PR and adopting new convention is the cleanest. We'll update behaviour in the respective conversation.

https://github.com/browsermt/bergamot-translator/issues/113. The defaults everyone wants now are off, so.

jerinphilip:
@qianqianzhu Is there anything waiting for this to be ready for review?

qianqianzhu:
> we decided to keep it on by default due to wanting users to verify it.

I agree that. But the question is that do we want to load from bytearrays by default? (as `--check-bytearray` will also set that to true).

jerinphilip (approved):

qianqianzhu:
@jerinphilip This PR is just a quick patch. I am more for fixing this issue with a more concrete solution. If you think we can further improve it, I am happy to work on it. 

> Is it possible to somehow make it conditional based on the format of the shortlist file?

We can add this condition in `BinaryShortlistGenerator` class. Let me know if you would like to change this class.

kpu:
I don't think the name of the file should determine if it's validated!  In general, using filenames to determine behavior is bad IMO.  

The goal here is that other people can provide models with some assurance that they won't crash / reveal memory they shouldn't / otherwise pwn your computer.  And that's really a download-then-validate thing.  I don't think changing the file name makes sense.

qianqianzhu:
The binary shortlist has a magic signature in the header, we do not need to check the filename to determine the format. The magic tells. 

It's my fault that I did not explain clearly why I want to turn off `--check-bytearray`. 
`--check-bytearray` option has two effects:
  - `true`: loading model, shortlist and vocabs from bytearrays and validate the bytearrays
  - `false`:  loading model, shortlist and vocabs from files and not validate the files

The key problem might be that `--check-bytearray` should only be responsible to determine whether to validate the bytearrays (not loading from bytearrays or files).

kpu:
Maybe we should kill `--check-bytearray`.  The binaries provided by the user should be used if provided.  If somebody wants to check the binary data is in bounds that should happen regardless of bytearray or file name.

XapaJIaMnu:
So in other words there should be a code that validates the non-binary lexical shortlist?

kpu:
What I meant is `--check-bytearray` seems to refer to the user providing `bergamot-translator` with a bytearray as opposed to a file name that then contains a binary.  Whether something is checked or not has little to do with being loaded by the library user or from a file.

qianqianzhu:
Mark this PR as a draft as more things to be done.

qianqianzhu:
Sorry this issus takes forever. The main thing makes me struggle is that we use `check-bytearray` to decide whether we construct Service from bytearray or file names in `service-cli`.
```cpp
  if (options->get<bool>("check-bytearray")) {
    // Load legit values into bytearrays.
    memoryBundle = marian::bergamot::getMemoryBundleFromConfig(options);
  }

  marian::bergamot::Service service(options, std::move(memoryBundle));
```
As the default value of `check-bytearray` is true, then it will always load the files into memory first and construct Service from bytearray. When non-binary files loaded into memory, the crash happens.

As bytearray loading is only a use case inclusive to Mozilla, we should construct Service from file names by default, right? So we can potentially allow more file types (like model.npz, lex.s2t, vocab.json), not limited to certain binary files.

> Maybe we should kill --check-bytearray.

If we kill `--check-bytearray`, we still need a switch/option to tell `service-cli` when to construct service from bytearray (mainly for testing purpose) and when not.

So I adopted previous @jerinphilip 's approach by introducing `--bytearray` option for testing purpose only.

@XapaJIaMnu @kpu @jerinphilip let me know if you are happy with this solution.

jerinphilip (approved):

jerinphilip (approved):

jerinphilip:
@qianqianzhu can you update the coding style and let's get this in?

jerinphilip (approved):

PR #155, opened 2021-05-17
Rewriting batching for threadsafety

kpu:
This does make the batcher a critical section across job submission and
cleaving though.  If that becomes a problem, we should go back to
incoming and outgoing queues with a batcher thread.

Also removes blocking mode from native compiles.

Note that translateMultiple no longer guarantees great batching.  Guess
we could lease the mutex from ThreadsafeBatcher and create a session.

There is the risk that one sentence comes in at a time and each thread
grabs one sentence at a time instead of better batching.  Not sure what
to do about that other than some sort of Nagle algorithm.

Due to non-deterministic batching, even with one thread, the regression
tests may go haywire if they depend on batching.


src/translator/service.h
```diff
@@ -99,9 +93,9 @@ class Service {
   /// some member in the Response, also specify any additional configurable
   /// parameters.
   std::future<Response> translate(std::string &&source,
-                                  ResponseOptions options);
+                                  ResponseOptions options = ResponseOptions());
```
XapaJIaMnu, line 96: Did something change to require the explicit default constructor?

XapaJIaMnu (approved):
So, basically you're adding a synchronisation point in the batch construction so that we don't crash when having multiple concurrent batch submissions.

src/translator/service.h
kpu: I probably should have done this in a separate PR but it was too painful to see two separate functions:

`std::future<Response> translate(std::string &&source)` that just creates a default ResponseOptions

`std::future<Response> translate(std::string &&source, ResponseOptions opions)`

XapaJIaMnu: I'm pretty sure i had complained about that somewhere...

qianqianzhu (dismissed):
Looks good to me. Maybe we shall replace Batcher with ThreadsafeBatcher (not for this PR).

kpu:
@qianqianzhu I thought about doing that but then the batching logic and threading code were getting intertwined and I was finding it hard to prove the number of entries was correct.  So I did it was a wrapper.  Possibly the next version will get bigger because it would have its own thread and queues if the batcher takes too much time.

qianqianzhu (approved):

PR #159, opened 2021-05-18
Merge histories across bergamot-translator forks

motin:
Merge histories across bergamot-translator forks. The goal is to get rid of unnecessary admin for manually syncing contents across forks while retaining both forks' histories. mozilla/main will be force-updated to be the same as browsermt/main after this PR is merged.


XapaJIaMnu (approved):

abhi-agg (approved):

motin:
Remember that we need to merge this PR, not rebase/squash, to retain both forks' histories :)

jerinphilip (approved):

PR #160, opened 2021-05-18
GitHub action to push browsermt/main branch to mozilla/bergamot-translator every hour

motin:
Designed to run in mozilla/bergamot-translator

### DIFF
--- a/.github/workflows/push-browsermt-main-to-mozilla-main.yml
+++ b/.github/workflows/push-browsermt-main-to-mozilla-main.yml
@@ -1,0 +1,23 @@
+name: Push browsermt/main branch to mozilla/bergamot-translator
+
+on:
+  schedule:
+  # Run every hour
+  - cron: "0 * * * *"
+
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Push browsermt/main branch to mozilla/bergamot-translator
+        if: github.repository == 'mozilla/bergamot-translator'
+        run: |
+          git clone -b main https://github.com/browsermt/bergamot-translator.git
+          cd bergamot-translator
+          git remote add mozilla https://x-access-token:${{ secrets.GITHUB_TOKEN }}@github.com/mozilla/bergamot-translator.git
+          git push mozilla main
--- a/.github/workflows/push-browsermt-main-to-mozilla-main.yml
+++ b/.github/workflows/push-browsermt-main-to-mozilla-main.yml
@@ -9,7 +9,7 @@ on:
 
 jobs:
   build:
-  if: github.repository == 'mozilla/bergamot-translator'
+    if: github.repository == 'mozilla/bergamot-translator'
     runs-on: ubuntu-latest
     name: Push browsermt/main branch to mozilla/bergamot-translator
     steps:
--- a/.github/workflows/push-browsermt-main-to-mozilla-main.yml
+++ b/.github/workflows/push-browsermt-main-to-mozilla-main.yml
@@ -10,7 +10,7 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
-
+name: Mirror
     steps:
       - uses: actions/checkout@v2
 
--- a/.github/workflows/push-browsermt-main-to-mozilla-main.yml
+++ b/.github/workflows/push-browsermt-main-to-mozilla-main.yml
@@ -9,13 +9,13 @@ on:
 
 jobs:
   build:
+  if: github.repository == 'mozilla/bergamot-translator'
     runs-on: ubuntu-latest
-name: Mirror
+    name: Push browsermt/main branch to mozilla/bergamot-translator
     steps:
       - uses: actions/checkout@v2
 
       - name: Push browsermt/main branch to mozilla/bergamot-translator
-        if: github.repository == 'mozilla/bergamot-translator'
         run: |
           git clone -b main https://github.com/browsermt/bergamot-translator.git
           cd bergamot-translator
--- a/.github/workflows/push-browsermt-main-to-mozilla-main.yml
+++ b/.github/workflows/push-browsermt-main-to-mozilla-main.yml
@@ -9,15 +9,19 @@ on:
 
 jobs:
   build:
-    if: github.repository == 'mozilla/bergamot-translator'
     runs-on: ubuntu-latest
-    name: Push browsermt/main branch to mozilla/bergamot-translator
+    name: Mirror a branch from a remote repo into this one
+    env:
+      source_repository: 'browsermt/bergamot-translator'
+      this_repository: 'mozilla/bergamot-translator'
+      branch: 'main'
     steps:
       - uses: actions/checkout@v2
 
-      - name: Push browsermt/main branch to mozilla/bergamot-translator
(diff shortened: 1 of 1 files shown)


motin:
I am not sure how to verify that this will actually work before merging it. @XapaJIaMnu @jerinphilip you are more familiar with GitHub actions, does this look like the right approach?

kpu:
@graemenail is our most qualified person for GitHub actions.

jerinphilip:
@motin Is there a way to make this push on main triggered here? For testing purposes use two branches instead of main to verify at both mozilla/browsermt. If it works we'll merge this in. Otherwise I guess you bring this in at Mozilla (keys/push rights) and see if it runs every hour testing on a branch over a day and bring the working file here.

abhi-agg:
Also, adding capability to automatically push tag info as a part of this action would be great 👍

motin:
> @motin Is there a way to make this push on main triggered here? 

If I understand Github actions access model correctly, the action needs to run in the mozilla/bergamot-translator repo to be able to have write access to the mozilla/bergamot-translator repo. Not sure if triggering on main branch push would help. 

> For testing purposes use two branches instead of main to verify at both mozilla/browsermt. If it works we'll merge this in. Otherwise I guess you bring this in at Mozilla (keys/push rights) and see if it runs every hour testing on a branch over a day and bring the working file here.

I pushed this branch to the mozilla repo as well, but that in itself doesn't make the action available for a test-run, and I don't want to merge it to or push to mozilla's main branch only since that would surely make the action fail. It is expected to work only when the branches share a common history cleanly. 

Are you suggesting that we first merge a PR for similar action that pushes something like browsermt/bergamot-translator#pr-160-sandbox to mozilla/bergamot-translator#pr-160-sandbox? Or would you be ok with merging this as is since this action only runs on mozilla/bergamot-translator by design?

.github/workflows/push-browsermt-main-to-mozilla-main.yml
```diff
@@ -0,0 +1,23 @@
...
+  schedule:
+  # Run every hour
+  - cron: "0 * * * *"
+
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
```
graemenail, line 13:
```suggestion
name: Mirror
```

or similar

graemenail (dismissed):
This approach works.

graemenail:
> > @motin Is there a way to make this push on main triggered here?
> 
> If I understand Github actions access model correctly, the action needs to run in the mozilla/bergamot-translator repo to be able to have write access to the mozilla/bergamot-translator repo. Not sure if triggering on main branch push would help.

You're right on that - you'd have to generate a person access token with permissions in the mozilla/bergamot-translator if you wanted trigger on pushes from here.

jerinphilip:
> Are you suggesting that we first merge a PR for similar action that pushes something like browsermt/bergamot-translator#pr-160-sandbox to mozilla/bergamot-translator#pr-160-sandbox? 

Just use this to demonstrate.

> Or would you be ok with merging this as is since this action only runs on mozilla/bergamot-translator by design?

There's no objection to begin with. Just saying it'd be easier to approve if you can show the .yml file runs on GitHub CI and works as intended, except with different parameters. In my experience, GitHub actions can fail due to type/indentations and testing it in main can make things messy.

motin:
> Just use this to demonstrate. No need of a merge.

@jerinphilip @graemenail This sounds good. How do I get the demonstration-workflow available for manual triggering without merging it to main?

motin:
Rebased this PR on the current main branch

jerinphilip:
> How do I get the demonstration-workflow available for manual triggering without merging it to main?

I dry ran this with a minimal example following @graemenail

* https://github.com/jerinphilip/gh-workflow-multi-repo-sync-mirror
* https://github.com/jerinphilip/gh-workflow-multi-repo-sync

There's a problem if we edit the workflow files, where GitHub rejects the push: [Reference](https://github.com/jerinphilip/gh-workflow-multi-repo-sync-mirror/runs/2612062097?check_suite_focus=true). The above example is parameterized enough with source, target repository and a branch which you can use to take this forward if necessary (use 'test-branch' instead of 'main'), finally update to main before merge.

@graemenail says you can have repo-secrets and store a new SECRET_TOKEN to use in the workflow. I'm not going that far since I don't have a use-case, feel free to try with the above boilerplate.

motin:
> > How do I get the demonstration-workflow available for manual triggering without merging it to main?
> 
> I dry ran this with a minimal example following @graemenail
> 
>     * https://github.com/jerinphilip/gh-workflow-multi-repo-sync-mirror
> 
>     * https://github.com/jerinphilip/gh-workflow-multi-repo-sync
> 
> 

Thanks for this!

> There's a problem if we edit the workflow files, where GitHub rejects the push: [Reference](https://github.com/jerinphilip/gh-workflow-multi-repo-sync-mirror/runs/2612062097?check_suite_focus=true).

This is an excellent safeguard which we want to have in place. The workaround is for an admin to manually push when workflow files are updated.

.github/workflows/push-browsermt-main-to-mozilla-main.yml
```diff
@@ -0,0 +1,27 @@
...
+    name: Mirror a branch from a remote repo into this one
+    env:
+      source_repository: 'browsermt/bergamot-translator'
+      this_repository: 'mozilla/bergamot-translator'
+      branch: 'main'
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Mirror a branch from a remote repo into this one
+        if: github.repository == '${{ env.this_repository }}'
```
jerinphilip, line 22: I'm hoping this works. I just got bit by an issue in my source repository with failures because

> When you use expressions in an if conditional, you may omit the expression syntax (${{ }}) because GitHub automatically evaluates the if conditional as an expression, unless the expression contains any operators. If the expression contains any operators, the expression must be contained within ${{ }} to explicitly mark it for evaluation. For more information, see "Context and expression syntax for GitHub Actions."

I'd strongly recommend testing `test-branch` -> `test-branch` updating it with master and ensuring it works before a merge to avoid debugging later.

motin: > I'd strongly recommend testing test-branch -> test-branch updating it later with main -> main and ensuring it works before a merge to avoid debugging later.

I'd like to do that, but can it be tested without merging? Or are you suggesting that I also set up two separate repos and apply it there first?

graemenail: I think what is being suggested is to have `source/temp_branch` pushed to `target/temp_branch` such that the main branch isn't altered in the event it doesn't work as expected.

However, in order to use `schedule`, the workflow must be on the default branch of the repository. I believe the same* is true of `workflow_dispatch`. In either case, it would need this file merged to the default branch.

*the workflow has to exist on the default branch, but can be run against other branches

jerinphilip (approved):
I think the YAML has been scrutinized enough now, go have fun with main @motin.

motin:
Thanks everybody for scrutinizing and testing this!

motin:
I'll wait for an approval @andrenatal before merging.

andrenatal (approved):

jerinphilip:
@motin The if is probably not working in this repository. https://github.com/browsermt/bergamot-translator/actions/runs/867430846/workflow;

PR #163, opened 2021-05-19
Adding a lot of warnings

jerinphilip:
Currently importing marian's warnings. 

[WIP] Also trying to have compiler warn about redundant `std::move(...)` which is used plenty in the source. https://developers.redhat.com/blog/2019/04/12/understanding-when-not-to-stdmove-in-c.


CMakeLists.txt
```diff
@@ -34,6 +37,18 @@ if(MSVC)
   set(CMAKE_EXE_LINKER_FLAGS_RELEASE "${CMAKE_EXE_LINKER_FLAGS} /NODEFAULTLIB:MSVCRT")
   set(CMAKE_EXE_LINKER_FLAGS_DEBUG   "${CMAKE_EXE_LINKER_FLAGS} /NODEFAULTLIB:MSVCRTD")
   set(CMAKE_STATIC_LINKER_FLAGS      "${CMAKE_STATIC_LINKER_FLAGS} /LTCG:incremental")
+else(MSVC)
+
+  # Borrowed from marian.
+  list(APPEND ALL_WARNINGS -Wall; -Werror; -Wextra; -Wno-unused-result; -Wno-deprecated;
+    -Wno-pragmas; -Wno-unused-parameter; -Wno-unused-function;
+    -Wno-unused-value; -Wno-unknown-pragmas; -Wno-sign-compare;
```
XapaJIaMnu, line 45: Unused value is not really useful. It creates more harm than good. At least don't promote it to `Werror`as it is done in the marian codebase. Leave it as a warning and let people fix it as they code. 

Same goes for unused parameter and unused functions and unused result. Let's not go to the marian codebase style of writing a lot of pointless stuff to make a compiler happy (eg https://github.com/marian-nmt/marian-dev/blob/fe74576dc3e957f97b4842951fb6ee7ec52fa324/src/training/communicator.cpp#L53 )

Not sure

CMakeLists.txt
```diff
@@ -40,15 +40,22 @@ if(MSVC)
...
-    -Wno-pragmas; -Wno-unused-parameter; -Wno-unused-function;
-    -Wno-unused-value; -Wno-unknown-pragmas; -Wno-sign-compare;
-    -Wno-missing-field-initializers;)
+  list(APPEND BERGAMOT_WARNINGS -Wall; -Wextra; -Werror
+      -Wunused-result; -Wdeprecated;
+      -Wno-pragmas; -Wno-unknown-pragmas; -Wno-sign-compare; -Wno-missing-field-initializers;)
   # This warning does not exist prior to gcc 5.0
   if(CMAKE_COMPILER_IS_GNUCC AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 5.0)
-    list(APPEND ALL_WARNINGS -Wsuggest-override -Wno-int-in-bool-context)
+    list(APPEND BERGAMOT_WARNINGS -Wno-suggest-override -Wno-int-in-bool-context)
```
jerinphilip, line 48: set back to suggest-override

jerinphilip:
clang-tidy is softly operational which in about 10 minutes throws a lot of warnings about source code-smells. This PR has been stale for a while. I don't expect to pursue this via compiler warnings, closing.

Thank you @XapaJIaMnu for your initial input, I should be able to put to use the learning eventually. 😄

XapaJIaMnu:
I tried myself to get this to work per source file, so we can have a more fine grained choice regarding the warnings, but I couldn't get it to work with cmake. https://stackoverflow.com/questions/15133332/ignore-only-show-errors-warnings-from-certain-directory-using-cmake Maybe our includes cascaded too much? Not sure.

Issue #161, opened 2021-05-18
Version file generation is done in the src tree


XapaJIaMnu:
@abhi-agg it is bad practise to have cmake generate files in the source directory. If a file is generated after configure, it should be in `${CMAKE_BINARY_DIR}`, not in `src/translator/project_version.h`.

At the very least, you can add a `.gitignore` entry for it so that this file is not accidentally picked up by someone working on the project.

I have not fixed this myself, as other projects also read this file, and I am not sure where exactly things need to be updated.

kpu:
Agreed, generated files go in the binary directory.

PR #164, opened 2021-05-19
Git-ignore generated project version files

motin:
Git-ignore generated project version files from https://github.com/browsermt/bergamot-translator/pull/153

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,10 @@
 *.swp
 *.swo
 
+# Config files from CMake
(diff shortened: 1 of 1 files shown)


kpu:
We shouldn't do this.  They should be built properly in the build directory.

kpu (changes requested):
Don't generate the version files in the source tree #161

motin:
> Don't generate the version files in the source tree #161

Agree, that is a much better approach. Closing.

PR #165, opened 2021-05-19
Pin emsdk version to the same one used in Circle CI

motin:
This makes the WASM artifacts built locally and those generated by Circle CI more similar.


build-wasm.sh
```diff
@@ -30,8 +30,8 @@ if [ "$EMSDK" == "" ]; then
   fi
   if [ "$EMSDK_UPDATE_REQUIRED" == "1" ]; then
     cd emsdk
-    ./emsdk install latest
-    ./emsdk activate latest
+    ./emsdk install 2.0.9
```
jerinphilip, line 33: I'm no expert here, but $EMSDK_VERSION and also some overridability outside if need be is sensible? Default to 2.0.9?

jerinphilip (approved):

andrenatal (approved):

build-wasm.sh
motin: There are already some env var similar to that set up emsdk, so I don't want to confuse things. Having this build parameter parametrized would be good, but pinning it is at least a step in that direction. Incremental improvement :)

PR #168, opened 2021-05-19
Adding documentation action

jerinphilip:
Uploads latest documentation to repository's gh-pages branch.

There ~~can be~~ is some trickery to do this per version and also latest, like how docs are normally structured. Have latest `v<major>.<minor>.<suffix>` corresponding to releases as well.

Here's demonstration of a successful run on my fork:

* https://github.com/jerinphilip/bergamot-translator/runs/2625000572?check_suite_focus=true
* https://jerinphilip.github.io/bergamot-translator/ (sample-render)

Also could have doxygen warn about undocumented members, but maybe in a later PR.

The yaml is borrowed from @graemenail, upload-artifact ~~cut~~ only on pull request, gh-pages publish moved in.

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -2,7 +2,7 @@ name: Documentation
 
 on:
   push:
-    branches: [ $default-branch ]
+    branches: [ $default-branch, ci-sandbox ]
 
 jobs:
   api-documentation:
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -16,7 +16,12 @@ jobs:
         with:
           submodules: recursive
 
-      # Be careful if you have a tag named main, bad idea
+      # Runs javascript to extract push events from both tags and branch (only main, due to workflow trigger)
+      # converts refs/<>/<name> -> <name>
+      # eg:
+      #     refs/head/main   -> main
+      #     refs/tags/v0.1.0 -> v0.1.0
+      #
       - name: Extract tag name
         id: tag
         uses: actions/github-script@0.2.0
@@ -28,13 +33,15 @@ jobs:
             [refs, category, ...rest] = args;
             return rest.join("/");
 
-      - name: Deploy-time patch version
-        run: |
-            echo ${{ steps.tag.outputs.result }} > BERGAMOT_VERSION
+      # Patches the BERGAMOT_VERSION file used by sphinx-docs at run time to
+      # obtain names like 'main' or 'ci-sandbox' to not confuse with version
+      # based documentation built separately.
+      - name: Deploy-time patch version 
+        run: | 
+            echo ${{steps.tag.outputs.result }} > BERGAMOT_VERSION
 
       - name: Set up Doxygen
         run: sudo apt-get install -y doxygen
-
 
       - name: Set up Python
         uses: actions/setup-python@v2
@@ -70,6 +77,8 @@ jobs:
 
       # This artifact contains the HTML output of Sphinx only.
       # With index.html at the root of the produced zip file.
+      # For use for maintainers to download the zip and check render of
+      # documentation while generated at pull-request. 
       - name: Upload documentation
         uses: actions/upload-artifact@v2
         if: ${{ github.event_name == 'pull_request'}}
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -2,7 +2,7 @@ name: Documentation
 
 on:
   push:
-    branches: [ master ]
+    branches: [ main ]
 
 jobs:
   api-documentation:
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -42,3 +42,13 @@ jobs:
         with:
           branch: gh-pages # The branch the action should deploy to.
           folder: ./doc/build/ # The folder the action should deploy.
+
+      # This artifact contains the HTML output of Sphinx only.
+      # With index.html at the root of the produced zip file.
+      - name: Upload documentation
+        uses: actions/upload-artifact@v2
+        with:
+          name: api-docs
+          path: ./doc/build/
+          if-no-files-found: error
+
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -57,9 +57,11 @@ jobs:
         uses: JamesIves/github-pages-deploy-action@4.1.3
         if: ${{ github.event_name == 'push' }}
         with:
+          repository-name: 'browsermt/docs' 
           branch: gh-pages # The branch the action should deploy to.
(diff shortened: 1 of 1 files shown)


qianqianzhu:
There are two things:
 -  Can we only include *.h in generating Library API. There is a bug in Exhale if we both include *.h and *.cpp, the global functions/variables declared could appear in *.cpp.  You can see that this bug also happens in [byte_array_util.h](https://jerinphilip.github.io/bergamot-translator/api/file_src_translator_byte_array_util.h.html#file-src-translator-byte-array-util-h) and [byte_array_util.cpp](https://jerinphilip.github.io/bergamot-translator/api/file_src_translator_byte_array_util.cpp.html#file-src-translator-byte-array-util.cpp). This is super annoying and I haven't found a way to solve it for Marian docs (we have to include several .cpp files there). If you don't need *.cpp files, let's only include *.h in generating Library API. 
 - Can we host docs webpages in a more proper place? Like `bergamot-translator.github.io`. I don't think users will build the docs themselves locally. They just click the link.

jerinphilip:
> Can we only include *.h in generating Library API. There is a bug in Exhale if we both include *.h and *.cpp, the global functions/variables declared could appear in *.cpp.

This is okay, right? From a user perspective? Are there any negatives to this? If and when marian has a workaround, we'll import it here? I can investigate this in a later PR if need be. I don't want to block this one on what could be a harder task (I don't know the impact of hiding *.cpp from Doxyfile, so..)

>  - Can we host docs webpages in a more proper place? Like `bergamot-translator.github.io`. I don't think users will build the docs themselves locally. They just click the link.

Is it not obvious that's where this is going? It pushes to gh-pages in this repository, which once this PR is in and gh-pages ready I will configure in settings to publish. The render is for demonstration on my repo, while I have push rights here I'm not overdoing it that's all.

graemenail:
> Is it not obvious that's where this is going? It pushes to gh-pages in this repository, which once this PR is in and gh-pages ready I will configure in settings to publish. The render is for demonstration on my repo, while I have push rights here I'm not overdoing it that's all.

It should end up at https://browsermt.github.io/bergamot-translator

.github/workflows/doc.yml
```diff
@@ -2,7 +2,7 @@ name: Documentation
 
 on:
   push:
-    branches: [ master ]
+    branches: [ main ]
```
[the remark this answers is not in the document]
jerinphilip: Thanks, didn't know this! :D

graemenail:
You may want to keep an artifact build for PRs as a check before merging

jerinphilip:
Note that to not have dead links later and some clean structuring, we could attempt the version thing as follows. 

* https://browsermt.github.io/bergamot-translator/latest (built from main)
* https://browsermt.github.io/bergamot-translator/vM.m.p (built from tags)

Do we only care for the current documentation like https://marian-nmt.github.io/docs/api/?

jerinphilip:
> You may want to keep an artifact build for PRs as a check before merging

I'll have to download and launch a server right? Does it render on the web? In this case won't pulling the PR/branch and running locally be more convenient?

qianqianzhu:
> This is okay, right? From a user perspective? Are there any negatives to this?
Personally, I would expect API reference to only appears in header files. But anyway, it is not quite critical. You may leave it for a later PR.

graemenail:
> > You may want to keep an artifact build for PRs as a check before merging
> 
> I'll have to download and launch a server right? Does it render on the web? If not won't pulling the PR/branch and running locally be more convenient?

Should be just HTML, so you can download the artifact and just open it in the browser.

.github/workflows/doc.yml
```diff
@@ -0,0 +1,54 @@
...
+        working-directory: ./doc
+        run: python3 -m pip install -r requirements.txt
+
+      - name: Build documentation
+        working-directory: ./doc
+        run: sphinx-build -b html ./ build/
+
+      - name: Deploy 🚀
+        uses: JamesIves/github-pages-deploy-action@4.1.3
+        with:
```
graemenail, line 42:
```suggestion
        if: ${{ github.event_name == 'push'}}
        with:
```

.github/workflows/doc.yml
```diff
@@ -0,0 +1,54 @@
...
+          branch: gh-pages # The branch the action should deploy to.
+          folder: ./doc/build/ # The folder the action should deploy.
+
+      # This artifact contains the HTML output of Sphinx only.
+      # With index.html at the root of the produced zip file.
+      - name: Upload documentation
+        uses: actions/upload-artifact@v2
+        with:
```
graemenail, line 50:
```suggestion
        if: ${{ github.event_name == 'pull_request'}}
        with:
```

.github/workflows/doc.yml
```diff
@@ -0,0 +1,54 @@
+name: Documentation
+
+on:
+  push:
+    branches: [ $default-branch, ci-sandbox ]
```
graemenail, line 5:
```suggestion
    branches: [ $default-branch, ci-sandbox ]
  pull_request:
    branches: [ $default-branch ]
```

graemenail (commented):
This is more what I had in mind for artifacts and PRs

.github/workflows/doc.yml
jerinphilip: I can remove this following a final approve, right now I demonstrate yaml works using ci-sandbox. Hope you understand.

What I'm trying for is to catch github.ref and push to target folder now attempting to strip `refs/tags`. It'll solve the issue with ci-sandbox.

https://github.community/t/how-to-get-just-the-tag-name/16241

graemenail: This adds a PR trigger; I left the `ci-sandbox` as I assumed you were testing something.

jerinphilip:
The behaviour is updated as follows. 

* https://browsermt.github.io/bergamot-translator/<branch> (built from branch)
* https://browsermt.github.io/bergamot-translator/vM.m.p (built from tags that map to semantic versions)

jerinphilip:
After some offline inputs from @graemenail and @kpu, the strategy is now to keep the generated files in a docs repository (https://github.com/browsermt/docs). There's an SSH public-key (browsermt/docs) and private-key (here, bergamot-translator) to enable the same, stored in secrets.

ci-sandbox documentation is now live for a demo at: 
* https://browser.mt/docs/ci-sandbox/

There's a runtime patch to update `BERGAMOT_VERSION` file which sphinx uses to point to the branch in place.

graemenail (approved):
Looks great.

I suggest you add a few lines in the README on `main` over at [browsermt/docs](https://github.com/browsermt/docs) to explain that the content is located at `https://browser.mt/docs/` and is built from this repo.

qianqianzhu (approved):

PR #170, opened 2021-05-21
Pin emsdk version to the same one used in Circle CI (#165)

motin:
Not meant to be merged, only to clarify the differences of this branch and the v0.3.1 tag. This change was necessary to build the wasm artifacts from a fresh clone. Artifacts built with the most recent emsdk are not compatible with the extension currently.


abhi-agg:
@motin Should we close this PR?

motin:
> @motin Should we close this PR?

Yes, it is time.

PR #172, opened 2021-05-23
Load sentence-splitter (non-breaking prefixes) from ByteArray

jerinphilip:
1. There was previously a `SentenceSplitter` wrapper around ssplit-cpp's `SentenceSplitter`, we're getting rid of it. This moves the logic one-level up into `TextProcessor`. We'll slowly keep code one-place here to expose consistent sentence-splitting API for the client to know where to cut sentences.
2. These areas have not been visited since we added documentation, so the documentation is updated.
3. Update BRT with a bunch of ssplit cases, to confirm behaviour is as intended after the mechanisms here: This is covered in https://github.com/browsermt/bergamot-translator/pull/191, which requires #184. ssplit-cpp's sample.en is copied over and used. You can check the inputs and expected outputs at https://github.com/browsermt/bergamot-translator-tests/pull/37, which are succeeding at #191. 

Tests here are failing because I forbid the absence of ssplit-prefix-file.


XapaJIaMnu (approved):
I don't see anything wrong here. Are there supposed to be any changes at the wasm side?

Also, what happens when ssiplit-prefix-file is not provided? We get a warning?

src/translator/text_processor.cpp
```diff
@@ -3,27 +3,85 @@
...
+
+ug::ssplit::SentenceSplitter loadSplitter(const std::string &ssplitPrefixFile) {
+  // Temporarily supports empty, will be removed when mozilla passes ssplitPrefixFile
+  ug::ssplit::SentenceSplitter splitter;
+  if (ssplitPrefixFile.size()) {
+    std::string interpSsplitPrefixFile = marian::cli::interpolateEnvVars(ssplitPrefixFile);
+    LOG(info, "Loading protected prefixes for sentence splitting from {}", interpSsplitPrefixFile);
+    splitter.load(interpSsplitPrefixFile);
+  } else {
+    LOG(warn,
```
XapaJIaMnu, line 36: What behaviour do we have in this case? What default is used?

src/translator/text_processor.cpp
```diff
@@ -3,27 +3,85 @@
...
+        "Missing list of protected prefixes for sentence splitting. "
+        "Set with --ssplit-prefix-file.");
+  }
+  return splitter;
+}
+
+ug::ssplit::SentenceSplitter loadSplitter(const AlignedMemory &memory) {
+  // Temporarily supports empty, will be removed when mozilla passes memory
+  ug::ssplit::SentenceSplitter splitter;
```
XapaJIaMnu, line 45: What happens if the memory is nullptr? Is the `splitter` object valid? Does it default to the old behaviour?

src/translator/text_processor.cpp
```diff
@@ -3,27 +3,85 @@
...
 Segment TextProcessor::tokenize(const string_view &segment, std::vector<string_view> &wordRanges) {
   // vocabs_->sources().front() is invoked as we currently only support one source vocab
   return vocabs_.sources().front()->encodeWithByteRanges(segment, wordRanges, /*addEOS=*/false, /*inference=*/true);
 }
 
-TextProcessor::TextProcessor(Vocabs &vocabs, Ptr<Options> options) : vocabs_(vocabs), sentence_splitter_(options) {
-  max_length_break_ = options->get<int>("max-length-break");
-  max_length_break_ = max_length_break_ - 1;
-  ABORT_IF(max_length_break_ < 0, "max-length-break cannot be < 0");
+TextProcessor::TextProcessor(Ptr<Options> options, const Vocabs &vocabs, const std::string &ssplit_prefix_file)
```
XapaJIaMnu, line 60: Do we preserver an option for no ssplit file provided?

src/translator/text_processor.h
```diff
@@ -3,43 +3,70 @@
...
-  // TextProcessor handles loading the sentencepiece vocabulary and also
-  // contains an instance of sentence-splitter based on ssplit.
-  //
-  // Used in Service to convert an incoming blog of text to a vector of
-  // sentences (vector of words). In addition, the ByteRanges of the
-  // source-tokens in unnormalized text are provided as string_views.
+  /// TextProcessor handles loading the sentencepiece vocabulary and also
+  /// contains an instance of sentence-splitter based on ssplit.
+  ///
+  /// Used in Service to convert an incoming blog of text to a vector of
```
XapaJIaMnu, line 21: Blog -> Blob

XapaJIaMnu:
Not entirely sure what happens when no ssplit-prefix file is provided. Other than that, I approve, but want some clarifications before I merge.

src/translator/text_processor.cpp
jerinphilip: Yeah, I put it back in, although it's not tested because common args are modified to include ssplit-prefix-file, effecting across all tests (https://github.com/browsermt/bergamot-translator-tests/pull/37). Should I forbid this behaviour? I'm inclined towards (1) bifurcating ByteArray based construction and FilePath based construction into two explicit paths and additionally constraining that (2) all these payloads *should* exist for bergamot-translator (i.e ssplit file should always be provided). 

Checking if ByteArray Empty to then switch-over to filepath based load to then a case of no-prefix file seems rather unpleasant.

src/translator/text_processor.cpp
jerinphilip: Regular expression based splitting without non-breaking prefixes accounted for. @ugermann can confirm.

src/translator/text_processor.cpp
jerinphilip: `TextProcessor` construction switches over to filepath based and this function is not called. Should I place an assert here just in case?

src/translator/text_processor.cpp
XapaJIaMnu: My concern is that the models we have currently don't include ssplit prefix files. Also, mozilla have their own fork of the models. We would need to negotiate ssplit prefix being mandatory with everyone before merging this.

kpu: Well, let's add the ssplit files to the models.

src/translator/text_processor.cpp
kpu: It breaks sentences where it shouldn't.

kpu (approved):

PR #173, opened 2021-05-24
[WIP] Quality estimator

abarbosa94:
Hi there,

This is the first attempt to design the overview of the Quality Estimator class. 

What I had in mind was the following:
- It should be better if the class is agnostic w.r.t. the model that is used to estimate translation quality. Then, if we decide to use a distilled neural net instead of a glm this should be ok to port :)
- Therefore, I remember that this is the idea behind [ONNX](https://github.com/onnx/onnx). Then, the ideal behavior should be that the QualityEstimator model receives a `.onnx` file, as well as the response object, returned from the `service.translate` method. If we go for it, then it should be easy to generate onnx files from sklearn pipelines from the instance.

Let's take an example:
Suppose that you have the following pipeline:

```
pipe = Pipeline([('std',StandardScaler()), ('lr',LogisticRegression())])
```

You could convert it to ONNX format with the following code:

```
# Convert into ONNX format
from skl2onnx import convert_sklearn
from skl2onnx.common.data_types import FloatTensorType
initial_type = [('float_input', FloatTensorType([None, 30]))]
onx = convert_sklearn(pipe, initial_types=initial_type)
with open("tmp.onnx", "wb") as f:
    f.write(onx.SerializeToString())

```

And then, read it through our C++ app with:
```
onnx::ModelProto model;
std::ifstream in(filename, std::ios_base::binary);
model.ParseFromIstream(&in);
in.close();

```

The pipeline methods would be accessible as well. Suppose that you want to access the third coefficient of our logistic model. Then, as it is the second step in our pipeline:
```
model.graph().node(1).attribute(1).floats(3)*(-1)
```

- there is a small bug that I still did not understand to get why it is necessary to multiply by -1 to get the original value

I just created an integration with ONNX only having in mind if we would like to show to the user some confidence "importance" based on the model coefficients, but if we just want the predictions themselves, it should be better to use [ONNX Runtime](https://www.onnxruntime.ai/), which also has an API for C++

## Help wanted

Please notice that my C++ is not so fluent ): So I'd like to ask for some help considering:

- In  app/service-cli.cpp, I hardcoded the file path with onnx location, how should I fix this?
- src/translator/quality_estimator.h I was only managed to import ONNX correctly if I use the path `3rd_party/onnx/onnx/onnx/onnx.pb.h` How should I fix this for the correct size?

- Finally, I've noticed that files inse 3rd_party repo are linked to other github repos. This is necessary to do with [ONNX](https://github.com/onnx/onnx) since to get it run I basically cloned the repo and compiled it with cmake + make. How is the best approach in this scenario?

Feel free to criticize and make suggestions since there have been ~8years since the last time I've coded in C++ and therefore I might be far away from best practices so this is a good opportunity from me to learn from my mistakes :)

### DIFF
--- a/src/translator/quality_estimator.h
+++ b/src/translator/quality_estimator.h
@@ -1,0 +1,31 @@
+#ifndef SRC_BERGAMOT_QUALITY_ESTIMATOR_H_
+#define SRC_BERGAMOT_QUALITY_ESTIMATOR_H_
+
+#include <iostream>
+#include <fstream>
+
+#define ONNX_API
+#include "response.h"
+#include "quality_estimator.h"
+#include "3rd_party/onnx/onnx/onnx/onnx.pb.h"
+
+namespace marian {
+namespace bergamot {
+
+class QualityEstimator {
+  public:
+    onnx::ModelProto model;
+    Response response_;
+    QualityEstimator(Response &response, std::string onnx_path) {
+      std::ifstream in(onnx_path, std::ios_base::binary);
+      model.ParseFromIstream(&in);
+      in.close();
+      response_ = response;
+    }
+
+    Response updateQualityScores();
+};
+} // namespace bergamot
+} // namespace marian
+
+#endif //  SRC_BERGAMOT_QUALITY_ESTIMATOR_H_
--- a/src/translator/response_builder.cpp
+++ b/src/translator/response_builder.cpp
@@ -5,11 +5,11 @@ namespace marian {
 namespace bergamot {
 
 void ResponseBuilder::buildQualityScores(Histories &histories,
-                                         Response &response) {
+                                         Response &response, int beamSize) {
   std::vector<Quality> qualityScores;
   for (auto &history : histories) {
-    // TODO(jerin): Change hardcode of nBest = 1
-    NBestList onebest = history->nBest(1);
+    //Due to inference performance, this should always be 1
+    NBestList onebest = history->nBest(beamSize);
 
     Result result = onebest[0]; // Expecting only one result;
     Words words = std::get<0>(result);
@@ -19,7 +19,7 @@ void ResponseBuilder::buildQualityScores(Histories &histories,
     // logprobs.
     auto normalizedPathScore = std::get<2>(result);
     auto wordQualities = hyp->tracebackWordScores();
-    wordQualities.pop_back();
+    wordQualities.pop_back(); //remove EOS token
     response.qualityScores.push_back(
         Quality{normalizedPathScore, wordQualities});
   }
--- a/src/translator/response_builder.h
+++ b/src/translator/response_builder.h
@@ -66,7 +66,7 @@ private:
   /// subword information.
   /// @param histories [in]
   /// @param response [out]
-  void buildQualityScores(Histories &histories, Response &response);
+  void buildQualityScores(Histories &histories, Response &response, int beanSize=1);
 
   /// Builds alignments from histories and writes onto response.
   /// @param histories [in]
--- a/3rd_party/CMakeLists.txt
+++ b/3rd_party/CMakeLists.txt
@@ -7,6 +7,8 @@ endif(COMPILE_WASM)
 
 add_subdirectory(ssplit-cpp)
 
+add_subdirectory(onnx)
+
 # Add include directories for 3rd party targets to be able to use it anywhere in the
 # project without explicitly specifying their include directories. Once they
 # fixe this problem, it can be removed.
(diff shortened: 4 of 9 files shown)


src/translator/response_builder.h
```diff
@@ -66,7 +66,7 @@ class ResponseBuilder {
   /// subword information.
   /// @param histories [in]
   /// @param response [out]
-  void buildQualityScores(Histories &histories, Response &response);
+  void buildQualityScores(Histories &histories, Response &response, int beanSize=1);
```
XapaJIaMnu, line 69: bean -> beam

kpu:
ONNX has poor support for sequential models due to its preference for a static graph.  Have you talked to Marina about whether her model is even supported by ONNX?  Last I heard there were n-gram language models somewhere.  

Sadly Marian doesn't have ONNX import.  For demo purposes, I guess one could cobble on another neural network toolkit, but it doesn't sound great from a binary size or production perspective (which is where we are aiming at this stage of the project).  I'm not sure how complicated the compute graph is but https://marian-nmt.github.io/docs/api/ .  

Welcome to WASM or even native containers.  Instead of opening files, you get handed memory.  
https://github.com/browsermt/bergamot-translator/blob/9dcf6ab665511fbe62d7ef327a367d13693c0330/src/translator/definitions.h#L21
https://github.com/browsermt/bergamot-translator/blob/9dcf6ab665511fbe62d7ef327a367d13693c0330/src/translator/service.h#L78
Put your model files in https://github.com/browsermt/students then somewhere in the magic of Mozilla and content distribution networks, it makes its way to loaded memory.  

> Please notice that my C++ is not so fluent
> have been ~8years since the last time I've coded in C++

Sheffield is responsible for integrating quality estimation in C++.  I understand Sheffield hired you as a C++ consultant.

src/translator/quality_estimator.h
```diff
@@ -0,0 +1,31 @@
...
+#include "3rd_party/onnx/onnx/onnx/onnx.pb.h"
+
+namespace marian {
+namespace bergamot {
+
+class QualityEstimator {
+  public:
+    onnx::ModelProto model;
+    Response response_;
+    QualityEstimator(Response &response, std::string onnx_path) {
```
jerinphilip, line 19: Your model (be it a neural network graph, logistic regression weights) will be loaded and kept in memory longer than one `Response`. The ideal plan is to instantiate `QualityEstimator` in `Service` and a reference to it given to `ResponseBuilder`. 

On each call to update `Response`, `ResponseBuilder` will call the same instance.

src/translator/response_builder.cpp
```diff
@@ -19,7 +19,7 @@ void ResponseBuilder::buildQualityScores(Histories &histories,
     // logprobs.
     auto normalizedPathScore = std::get<2>(result);
     auto wordQualities = hyp->tracebackWordScores();
-    wordQualities.pop_back();
+    wordQualities.pop_back(); //remove EOS token
```
jerinphilip, line 22: Try to not pop EOS and also return a quality score for EOS in your pipeline. It should be possible to provide quality score to EOS token.

The `pop_back()` needs to happen after your transformation for now, not before it goes into your pipeline.

src/translator/response_builder.cpp
```diff
@@ -5,11 +5,11 @@ namespace marian {
...
 
 void ResponseBuilder::buildQualityScores(Histories &histories,
-                                         Response &response) {
+                                         Response &response, int beamSize) {
   std::vector<Quality> qualityScores;
   for (auto &history : histories) {
-    // TODO(jerin): Change hardcode of nBest = 1
-    NBestList onebest = history->nBest(1);
+    //Due to inference performance, this should always be 1
+    NBestList onebest = history->nBest(beamSize);
```
jerinphilip, line 12: Restore to old state. I'm not sure what we gain by this parameterization of `beamSize` - could you elaborate why this was done?

kpu: I think Marina was doing something with comparing the outputs of beam search.  Should really document what the intent is here.

kpu:
If I understand correctly from @mfomicheva this is mostly a logistic regression model or other simple mathematical expression of the features.  Seems a bit overkill to bring in ONNX for a few matrix multiplies.  Can you write down the operations?

src/translator/response_builder.cpp
abarbosa94: @jerinphilip indeed. Thanks for the notice.

abarbosa94:
@kpu thanks for your feedback. I'll have a meeting with Marina on Wednesday so we can discuss further details about the model itself.
As I understood from the first meeting, the only information that I had was that it was trained on sklearn. Therefore, this is why I thought onnx should be a good deal.
The sklearn pipeline could be loaded as a computation graph at once and we could access its components sequentially. 
At a first sight, it should seem a better approach than handling two different data types (`normalization` coefficient from standard scaling + coefficients of the model). 

Another reason that I thought about this solution was that if we decide to go for different models, this should be easier to port since onnx provides well-defined abstractions. 

However, if ONNX seems to be an overkill decision. I could write down the operations as well. After the meeting with Marina, I intend to have a clear notion of the quality estimation model and thus I could update this PR.

src/translator/response_builder.cpp
abarbosa94: I'll have more info after Wednesday, but the primary reason to set this a parameter was just to use variable names instead of hard-coded code. I've set it as with a default value of 1 in `response_builder.h` in order to not change the behavior of the function. 

wdyt?

abarbosa94:
I'm closing this PR in favor of: https://github.com/browsermt/bergamot-translator/pull/197

PR #175, opened 2021-05-25
Single executable

### LINKED ISSUE #37: Single source file as an Application
Currently, there are 3 different files in [app folder](https://github.com/browsermt/bergamot-translator/tree/jp/absorb-batch-translator/app).

The app folder should contain only one file that demonstrates the use of unified APIs (which is done in [main.cpp](https://github.com/browsermt/bergamot-translator/blob/jp/absorb-batch-translator/app/main.cpp))

The other files should be removed once everything gets stabilized.

jerinphilip:
Almost there. Review ready.

Fixes #37. Each of the executable is about 400 MB due to static linking + debugging symbols, 3x is simply wasteful and hurts cache/artifact upload. This is now collapsing into a single executable with a `--bergamot-mode {service-cli,decoder,wasm}` switch. There's a runtime type inference and template code duplication. Let know if there are other clean ways to do RTTI.

TODO:
* ~~Update documentation which mentions three command line applications~~ Example render available at https://jerinphilip.github.io/bergamot-translator/marian-integration.html
* This breaks bergamot-translator-app and its use elsewhere, so negotiate.


kpu:
Related: we should override the -g in Marian for Release.

app/cli.h
```diff
@@ -0,0 +1,206 @@
...
+
+void execute(const std::string &mode, Ptr<Options> options) {
+  if (mode == "wasm") {
+    _execute<app::WASM>(options);
+  } else if (mode == "native") {
+    _execute<app::Native>(options);
+  } else if (mode == "decoder") {
+    _execute<app::Decoder>(options);
+  } else {
+    ABORT("Unknown --mode {}. Use one of {wasm,service-cli,decoder}", mode);
```
jerinphilip, line 199: note to self: service-cli -> native

app/cli.h
```diff
@@ -0,0 +1,206 @@
...
+/// @param [cmdline]: Options to translate passed down to marian through Options.
+/// @param [stdin] sentences as lines of text.
+/// @param [stdout] translations for the sentences supplied in corresponding lines
+class WASM : public CLIAppInterface {
+ public:
+  void execute(Ptr<Options> options) {
+    std::string config = options->asYamlString();
+
+    // Route the config string to construct marian model through TranslationModel
+    Service model(config);
```
XapaJIaMnu, line 45: https://github.com/browsermt/bergamot-translator/blob/8bec1b7b6ba30c522c102621e55b8cbe4bd051da/src/translator/service.h#L70

Service takes options as well, why do we need to convert back and forth?

app/cli.h
```diff
@@ -0,0 +1,206 @@
...
+    std::cout << "--------------------------\n";
+    std::cout << '\n';
+  }
+};
+
+}  // namespace app
+
+/// Template based duplication to call the respective classes based on mode string at run-time.
+template <class CLIAppType>
+void _execute(Ptr<Options> options) {
```
XapaJIaMnu, line 186: why not `operator()`

app/cli.h
```diff
@@ -0,0 +1,206 @@
...
+  CLIAppType app;
+  app.execute(options);
+}
+
+void execute(const std::string &mode, Ptr<Options> options) {
+  if (mode == "wasm") {
+    _execute<app::WASM>(options);
+  } else if (mode == "native") {
+    _execute<app::Native>(options);
+  } else if (mode == "decoder") {
```
XapaJIaMnu, line 196: Naming here is a bit confusing. Is decoder supposed to be "benchmark-only"

app/cli.h
```diff
@@ -0,0 +1,206 @@
...
+/// Template based duplication to call the respective classes based on mode string at run-time.
+template <class CLIAppType>
+void _execute(Ptr<Options> options) {
+  CLIAppType app;
+  app.execute(options);
+}
+
+void execute(const std::string &mode, Ptr<Options> options) {
+  if (mode == "wasm") {
+    _execute<app::WASM>(options);
```
XapaJIaMnu, line 193: As far as I can tell this template is not really used anywhere but here and it doesn't really achieve much. You can Avoid those 4 lines of code and and do something like. No need to do the fancy inheritance either.
```c++
if (mode == 'wasm") {
    WASM app;
    app.execute(options)
    }
```

doc/marian-integration.md
```diff
@@ -15,32 +27,47 @@ sudo apt-get install intel-mkl-64bit-2020.0-088
...
+`-DUSE_WASM_COMPATIBLE_SOURCE=on`.
+
 ```
 $ git clone https://github.com/browsermt/bergamot-translator
 $ cd bergamot-translator
 $ mkdir build
 $ cd build
 $ cmake .. -DUSE_WASM_COMPATIBLE_SOURCE=off -DCMAKE_BUILD_TYPE=Release
-$ make -j
+$ make -j4
```
XapaJIaMnu, line 45: This requires at least 16GB of RAM, or it would fail, beware.

XapaJIaMnu (changes requested):
I don't think the inheritance + templates are necessary, We don't make use of them. EG we don't pass the initiated object to another function to make use of the base class.

app/cli.h
jerinphilip: Objective is to test mozilla's path. 
cmdline -> Options -> std::string -> invoke std::string constructor.

XapaJIaMnu: Oh god... Can we, like, not?!! At the very least comment the shit out of it.

jerinphilip: Yes, I understand the horror of converting structured data back into string and then reconverting it either, but you gotta test that string based constructor somewhere. :'D

You see what I have to deal with on a daily basis?  I'll go with comment the hell out option, thanks.

jerinphilip: > Is there really no other usage of this codepath?

Nope. I like `Ptr<Options>` and will go with it any day in my dev. Would also not mind a struct equivalent that's not marian-dependent, but I don't see it a justified investment of time to write a new one. Only Mozilla uses it. I also use it in the cmdapp which tests Mozilla's codepath, coz, obvious reasons.

doc/marian-integration.md
jerinphilip: Set to 2

app/cli.h
jerinphilip: decoder is supposed to be same as `marian-decoder`, except using a different translation code-path (the request response one). This is yeah, for our purposes, benchmark only at the moment. 

I can see this collapsing into native somehow. I'll probably turn this into hijacking options and calling service directly, but later.

app/cli.h
jerinphilip:
```cpp
void execute(const std::string &mode, Ptr<Options> options) {
  if (mode == "wasm") {
    app::WASM(options).run();
  } else if (mode == "native") {
    app::Native(options).run();
  } else if (mode == "decoder") {
    app::Decoder(options).run();
  }
}
```

Changed to the above syntax, is this ok?

XapaJIaMnu: Sure. I don't see why not `app::Decoder(options)` though, but not too fussed.

jerinphilip: https://github.com/psalvaggio/cppregpattern This is what I'm angling at, pulling all that looks like an overkill. I want to runtime switch cli interfaces, all of which adhere to a common interface (CLIAppInterface) by the mode string.

app/cli.h
XapaJIaMnu: So, decoder = native - extra info besides translation?

app/cli.h
```diff
@@ -0,0 +1,206 @@
...
+namespace bergamot {
+
+// marian::bergamot:: makes life easier, won't need to prefix it everywhere and these classes plenty use constructs.
+
+namespace app {
+
+/// Interface for command-line applications. All applications are expected to use the Options based parsing until
+/// someone builds a suitable non-marian based yet complete replacement.
+
+class CLIAppInterface {
```
XapaJIaMnu, line 28: This is completely unnecessary.

app/cli.h
XapaJIaMnu: But... Why. This is done when you want to pass that object down the line to a number of functions or classes, so that you can take advantage of the polymorphic structure. In our case this is essentially a main, except it takes options, not argc and argv

app/cli.h
```diff
@@ -0,0 +1,206 @@
...
+  virtual void run(Ptr<Options> options) = 0;  // Make pure virtual, attempt to compile at MSVC
+};
+
+/// Previously bergamot-translator-app. Provides a command-line app on native which executes the code-path used by Web
+/// Assembly. Expected to be maintained consistent with how the browser (Mozilla through WebAssembly) dictates its API
+/// and tests be intact. Also used in [bergamot-evaluation](https://github.com/mozilla/bergamot-evaluation).
+/// @param [cmdline]: Options to translate passed down to marian through Options.
+/// @param [stdin] sentences as lines of text.
+/// @param [stdout] translations for the sentences supplied in corresponding lines
+class WASM : public CLIAppInterface {
```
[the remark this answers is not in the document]
jerinphilip: Compiler complains when subclasses don't implement `run(Ptr<Options>)` is the intention. As I mentioned, I want to turn the mode string to dynamic dispatch.

app/cli.h
```diff
@@ -0,0 +1,185 @@
+#ifndef BERGAMOT_APP_CLI_H
+#define BERGAMOT_APP_CLI_H
+#include <cstdlib>
+#include <future>
+#include <iostream>
+#include <sstream>
+#include <unordered_map>
```
XapaJIaMnu, line 7: Unnecessary include

XapaJIaMnu (approved):
With the understanding that we will eventually support proper command line options for bergamot, I approve this.

qianqianzhu (approved):

XapaJIaMnu (approved):

XapaJIaMnu:
@jerinphilip what's with the ci/circleci ? Can it be ignored?

jerinphilip:
@XapaJIaMnu For purposes of this repository, yes. Mozilla uses CircleCI to generate artifacts for extension. I don't think there are any tests per say, only that source builds for WASM.

I'd ideally prefer wait till someone from Mozilla notices this PR since I'm breaking the bergamot-translator-app and it has effects in places like bergamot-evaluation, if you're looking at merging this by the way. Since they're non responsive for a while, maybe a timeout on the preference for a week.

kpu:
circleci is borken for unrelated reasons, merging anyway

PR #176, opened 2021-05-25
Including WASM documentation in sphinx build toc

jerinphilip:
Includes wasm/README.md in sphinx generation toc. Removes hardcoded links to ensure sphinx not broken. 

See render demo at:
* https://browser.mt/docs/ci-sandbox/


qianqianzhu (commented):
Looks great! I only added one minor concern. Feel free to ignore it if you don't think it is important.

wasm/README.md
```diff
@@ -1,9 +1,11 @@
-## Using Bergamot Translator in JavaScript
+# Using Bergamot Translator in JavaScript
 The example file `bergamot.html` in the folder `test_page` demonstrates how to use the bergamot translator in JavaScript via a `<script>` tag.
```
qianqianzhu, line 2: Here is one minor concern: if the user reads `wasm-example` from sphinx docs, it is not clear where to find `test_page`. Maybe it is better change `test_page` to `wasm/test_page`.

jerinphilip: I have left a note in the beginning. Linking to this folder in the repository.

qianqianzhu (approved):

PR #179, opened 2021-05-26
Deploy generated documentation only if browsermt

jerinphilip:
The documentation workflow seems to attempt to push on forks as well, which is undesirable. This PR adds an additional guard to do this only if the main browsermt/bergamot-translator repository, not any fork.


qianqianzhu (approved):
I agree that we do not need docs for forks.

graemenail (approved):
> Secrets are not passed to workflows that are triggered by a pull request from a fork.

Just noting that the purpose of this PR is to stop this workflow from failing in forks.

PR #181, opened 2021-05-30
Update native (ubuntu, mac) workflows with ccache

jerinphilip:
* Matrix is now more organized, Ubuntu 20.04-gcc9.3, Ubuntu-18.04-gcc7.5 is added.
* ccache is extended to MacOS, and brings down CI run times to <5m when
  ccache works.
* The compiler hash scripts are gone, ccache already covers most ground
  by default. The shell script is unnecessary. Cache works by preprocessor
  mode output of running the compiler with -E, which includes the
  necessary information. [ccache-docs:How the cache works](https://ccache.dev/manual/4.3.html#_how_ccache_works).
* BRT if failed prints the final 100 lines of the test*.log to inspect 
  what's going wrong without having to artifact download.
* The rerun without cache workflow generates a bigger yml file, which inhibits
  editing by hand. The strategy is to bring in a `workflow_dispatch`, manually by
  the user input of the git-commit sha if failure is suspected due to ccache. 
  (This will come in a later PR).


.github/workflows/native.yml
```diff
@@ -0,0 +1,223 @@
...
+    - main
+  pull_request:
+    branches:
+    - main
+env:
+  ccache_base_dir: ${{ github.workspace }}
+  ccache_dir: "${{ github.workspace }}/.ccache"
+  ccache_compilercheck: content
+  cache_compress: 'true'
+  ccache_maxsize: 50M
```
graemenail, line 14: `50M` is sufficient?

.github/workflows/native.yml
```diff
@@ -0,0 +1,223 @@
...
+      with:
+        path: ${{ env.ccache_dir }}
+        key: ccache-${{ matrix.identifier }}-${{ steps.ccache_vars.outputs.hash }}-${{ github.ref }}-${{ steps.ccache_vars.outputs.timestamp }}
+        restore-keys: |-
+          ccache-${{ matrix.identifier }}-${{ steps.ccache_vars.outputs.hash }}-${{ github.ref }}
+          ccache-${{ matrix.identifier }}-${{ steps.ccache_vars.outputs.hash }}
+          ccache-${{ matrix.identifier }}
+    - name: ccache environment setup
+      run: |-
+        echo "CCACHE_COMPILER_CHECK=${{ env.ccache_compilercheck }}" >> $GITHUB_ENV
```
graemenail, line 82: Is there any reason we can't use the proper names for these when defining `env:` and skip this step?

.github/workflows/native.yml
```diff
@@ -0,0 +1,223 @@
...
+    - name: Upload regression-tests artifacts
+      uses: actions/upload-artifact@v2
+      if: ${{ always() && steps.brt_run.outcome != 'skipped' }}
+      with:
+        name: brt-${{ matrix.identifier }}
+        path: |-
+          bergamot-translator-tests/**/*.expected
+          bergamot-translator-tests/**/*.log
+          bergamot-translator-tests/**/*.out
+  mac:
```
graemenail, line 128: Do we prefer the Ubuntu and MacOS jobs to be separate? 
 
If we wanted to reduce some of the duplication it seems we could use either: 
```yaml
if: runner.os == 'Linux'
if: runner.os == 'macOS' 
```
to break run Linux
 - Linux dependencies
 - MKL
 
and MacOS
 - MacOS dependencies
 - BLAS paths

specific portions.

graemenail (changes requested):

.github/workflows/native.yml
jerinphilip: Recent workflow runs: [Ubuntu 18.04](https://github.com/browsermt/bergamot-translator/pull/181/checks?check_run_id=2714017754), [MacOS 10.15](https://github.com/browsermt/bergamot-translator/pull/181/checks?check_run_id=2714017807) seem to be doing okay, look at the ccache epilog. 


<details>
<summary>Ubuntu 18.04 full</summary>

```
Run ccache -s # Print current cache stats
  ccache -s # Print current cache stats
  shell: /bin/bash -e {0}
  env:
    ccache_base_dir: /home/runner/work/bergamot-translator/bergamot-translator
    ccache_dir: /home/runner/work/bergamot-translator/bergamot-translator/.ccache
    ccache_compilercheck: content
    cache_compress: true
    ccache_maxsize: 50M
    ccache_cmake: -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_C_COMPILER_LAUNCHER=ccache
    CCACHE_COMPILER_CHECK: content
    CCACHE_BASE_DIR: /home/runner/work/bergamot-translator/bergamot-translator
    CCACHE_COMPRESS: 
    CCACHE_DIR: /home/runner/work/bergamot-translator/bergamot-translator/.ccache
    CCACHE_MAXSIZE: 50M
cache directory                     /home/runner/work/bergamot-translator/bergamot-translator/.ccache
primary config                      /home/runner/work/bergamot-translator/bergamot-translator/.ccache/ccache.conf
secondary config      (readonly)    /etc/ccache.conf
stats zero time                     Mon May 31 22:52:01 2021
cache hit (direct)                   221
cache hit (preprocessed)               8
cache miss                            14
cache hit rate                     94.24 %
cleanups performed                     2
files in cache                       700
cache size                          30.1 MB
max cache size                      50.0 MB
```

</details>


<details>
<summary>MacOS 10.15 full</summary>

```
Run ccache -s # Print current cache stats
  ccache -s # Print current cache stats
  shell: /bin/bash -e {0}
  env:
    ccache_base_dir: /Users/runner/work/bergamot-translator/bergamot-translator
    ccache_dir: /Users/runner/work/bergamot-translator/bergamot-translator/.ccache
    ccache_compilercheck: content
    cache_compress: true
    ccache_maxsize: 50M
    ccache_cmake: -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_C_COMPILER_LAUNCHER=ccache
    LDFLAGS: -L/usr/local/opt/openblas/lib
    CPPFLAGS: -I/usr/local/opt/openblas/include
    CCACHE_COMPILER_CHECK: content
    CCACHE_BASE_DIR: /Users/runner/work/bergamot-translator/bergamot-translator
    CCACHE_COMPRESS: 
    CCACHE_DIR: /Users/runner/work/bergamot-translator/bergamot-translator/.ccache
    CCACHE_MAXSIZE: 50M
cache directory                     /Users/runner/work/bergamot-translator/bergamot-translator/.ccache
primary config                      /Users/runner/work/bergamot-translator/bergamot-translator/.ccache/ccache.conf
secondary config (readonly)         /usr/local/Cellar/ccache/4.3/etc/ccache.conf
stats updated                       Mon May 31 23:16:32 2021
stats zeroed                        Mon May 31 23:13:30 2021
cache hit (direct)                   211
cache hit (preprocessed)              14
cache miss                            18
cache hit rate                     92.59 %
cleanups performed                     2
files in cache                       477
cache size                          20.9 MB
max cache size                      50.0 MB

```

</details>

Should I keep more? I'm assuming I have 5GB across all workflows for ccache. Can raise this number potentially depending on how much we blow up matrix.

.github/workflows/native.yml
jerinphilip: Are both `env.{xx}` and environment variable  (bash / PS / etc) same? I'm treating `env` (global, job), `matrix` more as variable store than environment variables updated behind the scenes in YAML generation, and this notation is more explicit.

> You can use most contexts at any point in your workflow, including when default environment variables would be unavailable.  

https://docs.github.com/en/actions/reference/environment-variables#determining-when-to-use-default-environment-variables-or-contexts

.github/workflows/native.yml
graemenail: Lets keep it at `50M`

.github/workflows/native.yml
graemenail: My experience with Ubuntu/MacOS is that the name has been the same. The `env` context contains only the items defined in the YAML entry, while for `run: env` it contains items in the context as well as the default runner environment.

If I want to add a new CCACHE option, it's clearer and easier to do it once.

jerinphilip: Discussed and agreed upon offline that we're keeping the additional indirection, to maintain the ops to be more explicit and consistent.

.github/workflows/native.yml
jerinphilip: 1. I don't feel positive about the if in the job steps code and a runner-os switch. We put such an `if` in the workflow, it makes the editing workflow a bit more complex than it needs to be (For example, I'm now thinking of tweaking the matrix combos vs adding another if to achieve the same thing). 

2. An ideal solution that contains things within the YAML IMO will be to do `${{setup-deps}}` at each matrix include-entry, and have it evaluated in a setup-deps step. However, this will create duplication in the matrix. I'm trying to add as less programming to the YAML as possible keeping it simple and stupid.

3. Another nice way I see people doing this is replace this is with a `bash setup.sh <build-type>` and run a `bash setup.sh ${build-type}}` loading build-type from matrix, where `build-type=(['ubuntu', 'mac'] x ['minimal', 'full'])`, which will internally makes the switch. This way we only use the matrix (less duplicity because build-type is a string) and no ifs to do the same thing as the matrix within job steps. The downside here is that things are no longer contained within the YAML.

Among the possibilities, I think the current state is the sweet spot (Process is contained in YAML, and the workflow is simple and stupid at the cost of some duplication).

graemenail: I'm happy for Ubuntu and MacOS to exist in distinct matrix runs.

3 is a desirable solution if we imagine people are going to run `setup.sh` locally, but I don't imagine they will as it's purely setting up the environment. Let's just stay as is?

jerinphilip:
I need this for #184, we're fixing any further things in main. Failure paths for ifs here will have to be tested when this goes live.

PR #177, opened 2021-05-26
translateMultiple only for WASM

jerinphilip:
This was formerly a second half of #142, which cannot go in without making a mess in the BRT repo until #175 gets in. Therefore, we'll queue it behind.

PR #184, opened 2021-05-31
Partial test-apps and tolerance in evaluations

jerinphilip:
Previously, service-cli was testing [all-or-nothing](https://github.com/browsermt/bergamot-translator-tests/blob/main/tests/basic/service-cli.intgemm_8bit.avx2.expected) (quality, alignments, sentences, etc).

Instead of making BRTs worse to hasten #142, #177 this PR brings makes the area cleaner first (BLEU(?) for text, precision, allow diffs based on [diff-nums.py](https://github.com/browsermt/bergamot-translator-tests/blob/main/tools/diff-nums.py) for numeric values). This (tentatively) allows for room for variations in output, effectively mitigating the issues due to disparities in outputs stemming from compiler / BLAS / operating-system / instruction differences.

The general-idea is to output text or numbers (not interleaved). Text is often printed mapped to source-sentences inorder to achieve some level of matching when alignment matrix values or quality-scores vary while the target-sentence/structure is constructed after floating point operations.


XapaJIaMnu:
What does this do on the approx-diff.py `   parser.add_argument('--greater-than', '-gt', default=40, type=float)`. It is used in a few tests.

jerinphilip:
It's an `assetGreaterThan(BLEU between generated and expected, 40)` in effect. BLEU is very fluctuating especially for small phrases. So the script also allows an error rate (`--allow-error-rate`) in the latest source.

I'm looking for feedback on the CPP here which can be reviewed now, I'll open a (bigger) separate PR on brt once the test apps here are sorted.

jerinphilip:
In plain English, rather than dictating the entire text should match 100% we can now:

1. each sentences have greater than B bleu with an allowed error rate of e%. (`--sentence-only --greater-than B --allow-error rate e`. 
2. Entire text has BLEU greater than B.

This is kindof a text analog to diff-nums.py. I haven't touched diff-nums.py as of now.

qianqianzhu (commented):
Is it a good idea that we mix functions and testing in one cli app? Ideally, `app/bergamat.cpp` is only used for demo and testing-purpose code should be in `src/tests`.

XapaJIaMnu:
The cpp part is dirty, but if we don't have cmd parsing than I guess it's OK. I +1 for moving it to separate test directory, at least we'd know it's not supposed to be run independently.

As far as tolerance is involved, I have had a bug in the past that gave 0.5 bleu loss and it took forever to point out. I'm more or less OK with some fp threshold tolerance for the exact scores, but we shouldn't use just small 2 sentences to test bleu. We have datasets with 1000 sentences that take less than a minute to complete. We can discuss this in another pr.

jerinphilip:
@XapaJIaMnu, @qianqianzhu tests are moved, and CI is green now. It's kept similar to marian, I don't know if this is for better or worse.

@XapaJIaMnu This tolerance + per experiment adaptation is extremely ad-hoc. Even I'm not happy with parameterizing it with BLEU etc. What I've actually done instead is have exact/approx switches at BRT. This can be adjusted per test-script, so I can apply specific tolerances to specific inputs. The following is my rationale.

1. Separating text and numbers now allow different modes of comparisons for both. 
2. The all-in-one test had to go away (I'm unable to independently experiment with ssplit-cpp bytearray loads). @qianqianzhu can now with lesser work add a test-app for alignments coming in, perhaps a unit test for the inside-outside algorithm.

> The cpp part is dirty, but if we don't have cmd parsing than I guess it's OK

There are two parts to this PR - (1) https://github.com/browsermt/bergamot-translator/pull/184, (2) https://github.com/browsermt/bergamot-translator-tests/pull/36 , let's use this thread on getting the CPP code reviewed. Can you clarify what cmd parsing you're looking for? I understand BRT#36 might be too big to review, but it'd help to use the discussion there to incorporate the BRT changes. Let me know if I can break it down into a few (I'm experimenting first time with a bunch of things - I don't know what's the "right" way to break it). 

> We have datasets with 1000 sentences that take less than a minute to complete. We can discuss this in another pr.

The marian-decoder part is not tested in BRT on GitHub CI as of now. I want to bring in wmt test sets (a few) with sacrebleu and have --bergamot-mode decoder path also incorporated. This will indeed be another PR.

src/tests/apps.cpp
```diff
@@ -0,0 +1,169 @@
...
+void annotated_text_sentences(Ptr<Options> options, bool source) {
+  ResponseOptions responseOptions;
+  Response response = translate_from_stdin(options, responseOptions);
+  AnnotatedText &annotatedText = source ? response.source : response.target;
+  for (size_t s = 0; s < annotatedText.numSentences(); s++) {
+    std::cout << annotatedText.sentence(s) << "\n";
+  }
+}
+
+void legacy_service_cli(Ptr<Options> options) {
```
qianqianzhu, line 114: I guess we can remove the `legacy_service_cli` method here. As each feature is tested separately in the previous test mode/method, this part becomes redundant. (Maybe it's better to keep this part in `app/cli.h` to show the usage of `Response`.)

qianqianzhu (changes requested):
@jerinphilip Many thanks for your efforts to separate tests to make it easier to test new features. Now a new test-app case can be simply added as a method. That is really helpful. I addressed a few comments. Feel free to discuss.

src/tests/apps.cpp
```diff
@@ -0,0 +1,169 @@
...
+namespace marian {
+namespace bergamot {
+namespace testapp {
+
+// Utility function, common for all testapps.
+Response translate_from_stdin(Ptr<Options> options, ResponseOptions responseOptions) {
+  // Prepare memories for bytearrays (including model, shortlist and vocabs)
+  MemoryBundle memoryBundle;
+
+  if (options->get<bool>("bytearray")) {
```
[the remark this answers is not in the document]
jerinphilip: I think this is preferable? My next use-case is where I will demonstrate ssplit's bytearray loads work, by printing source sentences from annotation (that are constructed based on ssplit) and compare against what I expect when I have non-breaking prefixes. I will activate both paths during the same `--bytearray` switch is easier here.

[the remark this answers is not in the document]
qianqianzhu: Then let's keep it for now.

src/tests/apps.h
```diff
@@ -0,0 +1,52 @@
...
+// empty-lines. The first line contains whole quality scores and the second line word quality scores, for each entry.
+void quality_scores(Ptr<Options> options);
+
+// Reads from stdin and translates. Alignments are printed aligned to the source-tokens, following format src-token:
+// [possible-target-alignments], if numeric is false. If numeric is true, only alignment probabilities are printed
+// instead of the tokens.
+void alignment_aggregated_to_source(Ptr<Options> options, bool numeric = false);
+
+// Reads from stdin and translates.  Prints the tokens separated by space for each sentence.
+void annotated_text_words(Ptr<Options> options, bool source = true);
```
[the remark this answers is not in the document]
jerinphilip: Comments are added. There is both source and target sentences / words which need to be printed where I have to switch between source.annotation or target.annotation, and I'm DRY-ing the shared bit making use of a boolean.

CMakeLists.txt
```diff
@@ -41,6 +41,9 @@ include(CMakeDependentOption)
 # Project specific cmake options
 option(COMPILE_WASM "Compile for WASM" OFF)
 cmake_dependent_option(USE_WASM_COMPATIBLE_SOURCE "Use wasm compatible sources" OFF "NOT COMPILE_WASM" ON)
+
+# WASM disables a million libraries, which also includes the unit test-library.
+cmake_dependent_option(COMPILE_UNIT_TESTS "Compile unit tests" OFF "USE_WASM_COMPATIBLE_SOURCE" ON)
```
kpu, line 46: @abhi-agg Opinion on this?  I'm going to let it in for now, but seems we should run tests...

src/tests/cli.cpp
```diff
@@ -0,0 +1,27 @@
...
+  } else if (mode == "test-response-source-sentences") {
+    testapp::annotatedTextSentences(options, /*source=*/true);
+  } else if (mode == "test-response-target-sentences") {
+    testapp::annotatedTextSentences(options, /*source=*/false);
+  } else if (mode == "test-response-source-words") {
+    testapp::annotatedTextWords(options, /*source=*/true);
+  } else if (mode == "test-response-target-words") {
+    testapp::annotatedTextWords(options, /*source=*/false);
+  } else {
+    ABORT("Unknown --mode {}. Please run a valid test", mode);
```
kpu, line 24: Would be more useful if it said what the valid tests are or at least pointed to the source code.

qianqianzhu (approved):

CMakeLists.txt
abhi-agg: The whole idea of creating `USE_WASM_COMPATIBLE_SOURCE` was to run tests. So, yeah tests should run for "minimal" marian. It's worth checking which libraries are missing that stop us from running the test suite (that runs for "full" marian) on "minimal" marian.

@jerinphilip Do you know which libraries are missing for this config? I can try enabling them for "minimal" marian.

PR #186, opened 2021-06-03
Remove addSentenceWithPriority

kpu:
This removes the unused `addSentenceWithPriority` function and inlines the one caller it had.


XapaJIaMnu (approved):
We are abandoning this?

jerinphilip (approved):

kpu:
We're not abandoning priority.  The function really means that it adds one sentence instead of a whole request, but nobody calls that.

Issue #149, opened 2021-05-12
Deciding on interface class names


abhi-agg:
Currently the extension is using the names `TranslationModel`, `TranslationRequest` and `TranslationResult` as these were the interface classes for the API consumers before.

However, as a part of [Collapse draft API and actual implementation](https://github.com/browsermt/bergamot-translator/issues/77), these names have been changed to `Service`, `ResponseOptions` and `Response` respectively.
 
There has been some concerns already (in this [PR](https://github.com/browsermt/bergamot-translator/pull/139/files#r629570571)) regarding doing this refactoring cleanly.

I see 3 options:
1. We finalize on the names `Service`,  `ResponseOptions` and `Response` => Refactoring will be done on Extension side (@motin can provide an estimate on the effort to do it)
2. We finalize on the names `TranslationModel`, `TranslationRequest` and `TranslationResult` => Refactoring will be done in this repo (I believe the effort is not too much?)
3. We adopt completely different names for these structs => Refactoring happens on both sides

Personally, the name `Service` for the main interface class doesn't sound intuitive to me. However, all of this is a matter of taste. All I am saying is, we should agree on names now so that no one has to do significant refactoring later.

cc @kpu @andrenatal @motin @XapaJIaMnu @jerinphilip

abhi-agg:
Ping @kpu @jerinphilip @XapaJIaMnu @andrenatal @motin

We need a resolution here. I proposed 3 options above and, personally, I am for the option `2`. Merging of https://github.com/browsermt/bergamot-translator/pull/139 is currently blocked on this.

kpu:
I don't particularly care either way about naming.  

There is a pending issue that we'll need to support multiple translation models at once in order to do back translation for outbound translation.  Also to support e.g. fr->de translation if we do pivoting via en.  So it will happen that TranslationModel / Service becomes more explicitly separated.

motin:
> Refactoring will be done on Extension side (@motin can provide an estimate on the effort to do it)

Refactoring effort is minimal regardless of what names are chosen. < 5 minutes.

XapaJIaMnu:
I don't particularly care about the name, as long as it's consistent.

kpu:
I'm going to interpret this as @abhi-agg is free to make a pull request on naming.

PR #195, opened 2021-06-14
Refactor wasm bindings to use consistent interface names as in native

abhi-agg:
Fixes https://github.com/browsermt/bergamot-translator/issues/149

Additionally, renamed bindings files following the standard that is being used in this repository.


wasm/test_page/bergamot.html
```diff
@@ -206,10 +206,10 @@
     const from = lang.substring(0, 2);
     const to = lang.substring(2, 4);
     let start = Date.now();
-    await constructTranslationModel(from, to);
-    log(`translation model ${from}${to} construction took ${(Date.now() - start) / 1000} secs`);
+    await constructTranslationService(from, to);
+    log(`translation service ${from}${to} construction took ${(Date.now() - start) / 1000} secs`);
     document.querySelector("#load").disabled = false;
-    //log('Model Alignment:', translationModel.isAlignmentSupported());
+    //log('Model Alignment:', translationService.isAlignmentSupported());
```
jerinphilip, line 212: The commenting implies `translationService.isAlignmentSupported()` is not covered by the test-page? 

Do we have any use for this information? Can we agree here to get rid of `isAlignmentSupported()` making it always true, removing the commented line, given the translation model is Marian, and from the looks of it I'm always setting alignment on since from the looks of it tag matching is a "requirement" to properly match links and from the methods chosen, internally we are always supporting (word) alignments. (cc @qianqianzhu, @kpu).

abhi-agg: It is not used anywhere at the moment. I have no issue in removing `isAlignmentSupported()` if the trained marian models can always return alignment information. The extension will use the returned alignments soon to re-insert tags.

However, this refactoring PR shouldn't be blocked on this. Removing this API can be done as a separate PR.

jerinphilip (approved):

PR #199, opened 2021-06-24
Implementation of Inside-outside algorithm for tag translation

qianqianzhu:
This PR implements an inside-outside algorithm for tag translation.
The basic idea of this feature is to reserve HTML tag positions in the translated sentence using the alignment and token information. 

The current implementation following the suggestions by @kpu:
- `TagProcessor` is built as a layer on top translate using the alignment and token information.
- The browser provides tag information as `ByteRange` in the source sentence and we return tag information as `ByteRange` in the target sentence in the same order.

There are still some things to be done:
(1) optimise the no-solution case. (some randomisations in the greedy algorithm)
(2) some corner cases are not optimal.
(3) empty tags need to be optimised.


src/translator/tag_processor.h
```diff
@@ -0,0 +1,130 @@
...
+  std::vector<std::vector<std::vector<double>>> inside_;
+
+public:
+  std::vector<TagNode> tagTreeTarget; // store query results
+
+  TagProcessor(const marian::data::SoftAlignment &align,
+               std::vector<TagNode> tagTree,
+               size_t srcLength,
+               size_t tgtLength) : tagTree_(std::move(tagTree)), srcLength_(srcLength), tgtLength_(tgtLength) {
+    inside_.resize(srcLength, std::vector<std::vector<double>>(srcLength, std::vector<double>(tgtLength, 0)));
```
jerinphilip, line 37: Allocate `srcLength * srcLength * tgtLength` on a flat array and use [address-calculation](https://en.wikipedia.org/wiki/Row-_and_column-major_order#Address_calculation_in_general)?

src/translator/tag_processor.h
```diff
@@ -0,0 +1,130 @@
...
+          double product = 1;
+          for (size_t t = 0; t < tgtLength_; t++)
+          {
+            if (t >= l && t < r)
+            {
+              product *= inside_[query.begin][query.end - 1][t];
+            }
+            else
+            {
+              product *= 1 - inside_[query.begin][query.end - 1][t];
```
jerinphilip, line 74: not sure if the right place but `argmax (sum(log(...)) = argmax (prod(...))` with better numeric stability. `maxi` is effectively `argmax`?

src/translator/tag_processor.h
qianqianzhu: That optimisation is good to be done later. Let me first make the core algorithm correct.

src/translator/tag_processor.h
qianqianzhu: I have the same thought with you since turning prod() to sum(log()) can also potentially reduce the overall time complexity using Kadane's subarray algorithm. However, I cannot prove the correctness of this strategy for this maximisation problem:
argmax_{i,j} (prod(t in [i,j)} p_{k} * (1-prod(k not in [i,j))
Do you have a sketch proof, or I should switch to sum(log()) if it empirically works?

kpu: Sounds like all you need is a f(x) = log(1-exp(x)) ?  There should be numerically stable implementations of this; log1p is a library function but dunno about integrated exp.

qianqianzhu: Let me try `log1p`.

motin:
If work on this PR or follow-up PRs eventually leads to a capability for bergamot-translator to translate tags and in turn allow for the extension not having to maintain and finish the detagAndProject methods, then maybe these test cases are relevant: https://github.com/mozilla-extensions/firefox-translations/blob/7cb7faffd9847ff87f587092078234d71ac2b820/src/core/ts/content-scripts/dom-translation-content-script.js/dom-translators/detagAndProject.spec.ts#L36-L188
Note that these test cases passed in v0.2.0 of bergamot-translator, but are failing since v0.3.0 and @abhi-agg is maintaining the detagAndProject code nowadays.

qianqianzhu:
> If work on this PR or follow-up PRs eventually leads to a capability for bergamot-translator to translate tags and in turn allow for the extension not having to maintain and finish the detagAndProject methods, then maybe these test cases are relevant: https://github.com/mozilla-extensions/firefox-translations/blob/7cb7faffd9847ff87f587092078234d71ac2b820/src/core/ts/content-scripts/dom-translation-content-script.js/dom-translators/detagAndProject.spec.ts#L36-L188
> Note that these test cases passed in v0.2.0 of bergamot-translator, but are failing since v0.3.0 and @abhi-agg is maintaining the detagAndProject code nowadays.

@motin many thanks for pointing out the relevant tests. That's exactly what I am looking for. I will try to integrate those test cases into this PR.

src/tests/units/tag_tests.cpp
```diff
@@ -0,0 +1,259 @@
...
+  //            TagNode({0, 0}, {})
+  //    };
+  //
+  //    testCase(softAlign, srcLength, tgtLength, tagTree, tagTreeTargetExpected);
+  //  }
+
+  SECTION("Nested tags with one child") {
+    // [original]: <span><span>A republican strategy</span> to counteract the re-election of Obama.</span>
+    // [translated]:<span><span>Eine republikanische Strategie</span>, um der Wiederwahl Obamas entgegenzuwirken.</span>
+    std::vector<TagNode> tagTree{TagNode({0, 14}, {1}), TagNode({0, 4}, {})};
```
jerinphilip, line 142: The way of constructing this should change, this is very unnatural. This should be something like: 

```cpp

class TagTree {
public:
	TagTree(TagNode root){
		vertices_.push_back(root);
	}
	void addChild(TagNode parent, TagNode child){
		vertices.push_back(child);
		parent.children.push_back(vertices.size()-1);
		// I should be able to place asserts here that illegal nesting does not happen.
		// assert (parent.l <= child.l and parent.r >= child.r); 
	} 
	
	const std::vector<TagNode> &vertices(){ return vertices_; }
	
private:
	std::vector<TagNode> vertices_;
};

TagTree tree;
TagNode root = tree.root({0, 14});
TagNode child = tree.addChild(root, {0, 4}}; 
```

The above structure is not ideal (I wouldn't mind pointers for parent and children if there are no complaints of allocations), but externally, something like this provides a more semantically meaningful API to construct the structure from DOM. `TagTree` (even if it's `vector<TagNode>`), `TagNode` needs to be documented heavily.

jerinphilip: In essence:
1. Any TagTree I construct through the API should be valid.
2. Any nodes I add are valid (children only to existing parents, with range constraints possibly checked). 
3. I should be able to programmatically build this using methods that are natural iterating through DOM elements. Currently I have to do one pass DOM, get the values and then create the `vector<TagNode>` which is not ideal.

qianqianzhu: My current version is only some experimental code which helps to get inside-outside algorithm work. The important part I would like to get from the browser (from Mozilla side) is a tree structure that contains tag information. Because it helps to preserve the tag nesting constraints. For the detailed design of API, let's get inputs from Mozilla side to see what works best.

qianqianzhu: Just two things to confirm. (i) Each validation check could take significant time. It can be as long as O(n) where n is the number of existing nodes in the tree (or O(log n) with an error-prone algorithm and additional space). The total time usage to build the whole TagTree would rise to O(n^2) (or O(n log n) with an error-prone algorithm). I think the user's DOM->tree conversion algorithm should naturally ensure tree validness with zero additional complexity. Is it a good trade to ensure the validness by ourselves at the price of higher complexity or lower maintainability? (ii) I don't mind using pointers to support the physical structure, but in this case, the addChild function would have a pointer-type argument. Is it fine to expose pointers to users? There are several ways to conceal the pointer, but I don't find a particularly good one.

abhi-agg: > The important part I would like to get from the browser (from Mozilla side) is a tree structure that contains tag information. Because it helps to preserve the tag nesting constraints.

@qianqianzhu I am thinking purely from the perspective of the changes required in the main interface class ([Service](https://github.com/browsermt/bergamot-translator/blob/main/src/translator/service.h#L42)) for this feature to work. I think, providing tag information as a list of `ByteRange` (e.g. `vector<ByteRange>`) should suffice. Right?

Of course, `vector<ByteRange>` can then easily be converted to any representation (e.g. tree structure) internally on which your algorithm can run efficiently. It would be great to keep the [Service](https://github.com/browsermt/bergamot-translator/blob/main/src/translator/service.h#L42) interface as simple as possible + passing a `vector<ByteRange>` instead of a tree structure from the browser would be more efficient.

qianqianzhu: @abhi-agg `vector<ByteRange>` would be OK.  I will do my best to convert this (using the begin and end info from `ByteRange`) into a tree structure.

abhi-agg: That's great. Thanks. If you get stuck somewhere then please feel free to drop a message 👍

qianqianzhu: @abhi-agg There is one potential problem when using `vector<ByteRange>`: if there are more than one tag around a word(s), the tag nesting order is not clear. E.g., `<p><em>hello world</em></p>`, the corresponding `vector<ByteRange>` is `std::vector<ByteRange> tagVector = {{0,2},{0,2}};`. The nesting order is not clear: we don't know if `tagVector[0]` is wrapped by `tagVector[1]` or `tagVector[1]` is wrapped by `tagVector[0]`. I was assuming the vector of byterange don't hold any tag sequence information. 

To deal with this problem, we could agree on some rule such as **when two byterange are the same in the vector, the first appearing byterange is the outermost tag**. (No expert in HTML. Maybe this nesting order for the same text is not a real problem in HTML.)

kpu: Can we go slightly stronger?  
The ranges shall always be sorted by:
1. (Primary) Increasing order of begin
2. Increasing order of end
3. Outer nesting

This way a depth first left-first traversal of the tag tree is equivalent to iterating through the vector in order.

abhi-agg: > The ranges shall always be sorted by:
> 
>     1. (Primary) Increasing order of begin
> 
>     2. Increasing order of end
> 
>     3. Outer nesting
> 
> This way a depth first left-first traversal of the tag tree is equivalent to iterating through the vector in order.

For clarity, what would be the expected `std::vector<ByteRange>` for the following html text:

Original HTML text:         `<div><br><b>abc def<i>ghi jkl</i></b> <u>stu vw</u></div>.`
Text after stripping tags:  `abc defghi jkl stu vw.`
Assuming:
ByteRange of tag `<i>` = [7, 14)
ByteRange of tag `<b>` = [0, 14)
ByteRange of tag `<u>` = [15, 21)
ByteRange of tag `<div>` = [0, 21)
ByteRange of tag `<br>` = [0, )  => please note that the tag `<br>` is an unclosed html tag

kpu: `<br>` and `` are more of a self-closing HTML tag.  
So I would do:
`<div><br><b>abc def<i>ghi jkl</i></b> <u>stu vw</u></div>`
maps to 
`[0, 21) [0,0) [0,14) [7,14) [15, 21)`
A simpler way of saying things is: the byte ranges should appear in the same order as the open tags in HTML.  

If you have the case `A <span>sentence. Second one.</span>` then after sentence splitting it acts like the tags close at the end of sentence and open at the beginning of the next sentence `A <span>sentence.</span>` and `<span>Second one.</span>`

qianqianzhu: @kpu thanks for your clarification.

@abhi-agg just one additional note about the unclosed/empty html tag: `begin = end` always holds and the tag is placed before the character. Based on your example,

`<br>abc`, the byterange for `<br>` is [0,0)

`a<br>bc`, the byterange for `<br>` is [1,1)

> The byte ranges should appear in the same order as the open tags in HTML.

I will follow this rule to construct TagTree. Since Marian’s alignment works on the token-level rather than character-level (which is passed from the browser), I need to find a way to bridge them.

abhi-agg:
@qianqianzhu That is great 👍  Couple of queries:

>     2. current tag processor only works on one sentence (after Sentence splitter). I need to solve the cross-sentence tag case.

1. Do you mean you need to solve the case where a tag spans across multiple sentences? e.g. `A <b>sentence. Second one.</b>`

2. For the above example, I am assuming that you are expecting a single entry in `std::vector<ByteRange>`. Is it correct? It would be `std::vector<ByteRange> = [2, 23)`.

Additional question:
For the entire text that comes your way (irrespective of whether it contains a phrase, single sentence or multiple sentences), you always expect just one `std::vector<ByteRange>` (i.e. you don't expect one`std::vector<ByteRange>` for each sentence in the text). Right?

kpu:
Abandoned in favor of @jelmervdl implementation.

Issue #104, opened 2021-04-19
Sentence splitter non-breaking prefixes file


kpu:
The non-breaking prefixes file for the sentence splitter depends on the source language.  We should bind this to the model somehow (i.e. by knowing what language it is translating).  Otherwise the model will be confused when it sees the wrong sentence split and has a mismatch with training.  

I'm beginning to think we should have a unified binary file like @XapaJIaMnu was suggesting.

kpu:
This is a problem.  We're not consistent between training and test.  We're also creating the impression to Mozilla that this file doesn't exist when it needs to, which will bite us later.

kpu:
Cleanest solution is probably to ship the file with the MT models.

Issue #102, opened 2021-04-17
Need API to find reasonable split for large blob of text


kpu:
If we take, say, the entire text of Crime and Punishment https://www.gutenberg.org/files/2554/2554-0.txt then it shouldn't be sent to the app all at once as one big string.  So ideally we'd impose some sort of size limit (like a maxi-batch) and break it into pieces.  Currently the API provides no way for the consumer to know where a good split point is without translating everything.  We should provide a facility to find a sentence split point in the vicinity of an offset in the text, so that the client can limit incoming blob size.

ugermann:
That's more or less why I implemented sentence splitting as a stream. You submit a blob of text and  get back one sentence one at a time. You can batch them off at your leisure once you have enough. If you use splitting mode one_paragraph_per_line or wrapped_text, you'll get an empty string_view at each paragraph boundary (the difference being that in one_paragraph_per_line, EOL counts as a paragraph boundary, whereas in wrapped_text, an empty line counts as a paragraph boundary, and non-empty lines are concatenated into a paragraph). So you can either send off the text one paragraph at a time, and/or count the sentences as they come in and make your own decisions as to when enough is enough (in the case of giant blobs of text without paragraph boundaries). 

We could push that logic into the translator by way of returning a multi-part response, sending back the response in pieces if necessary. In terms of sentence splitting the mechanics are there; you'll just have to let go of the idea of one giant string in, one giant string out; my original suggestion of returning a sequence of text chunks (sentences, paragraphs, whatever) for a single input was based on that idea but was immediately shot down at the time.

ugermann:
See also: ssplit_stream() here: https://github.com/ugermann/ssplit-cpp/blob/master/src/command/ssplit_main.cpp. Replace ssplit_chunk() by translate_chunk().

kpu:
Every case I see in https://github.com/ugermann/ssplit-cpp/blob/master/src/command/ssplit_main.cpp presumes the entire text to split is already in RAM or memory mapped.  With that assumption, we've already lost the battle for bounded memory consumption.  The browser shouldn't even be shipping us the entire document.  

So what I'm looking for is a means to say "here's the first 1 MB of text, tell me where the last sentence split boundary is, don't bother translating after that boundary, and the next request will start from there."  This could be as simple as an API to efficiently find the LAST sentence boundary in the chunk of text.  (And since we're capping sentence length, there will be one in a guaranteed range.)

PR #202, opened 2021-06-30
Caching translations implementation

### LINKED ISSUE #201: Caching translations
For potential use in outbound translation (#78) and https://github.com/XapaJIaMnu/translateLocally, we're implementing a translation caching mechanism. 

The plan is to implement a simple LRU cache (using a map + linked list) with

* Key : `std::string` or equivalent `marian::Words`
* Value : `History`

It is agreed upon offline that the value is `History`. Since `marian::Words` are what generates history, I propose we just implement a hash for `marian::Words` and use this as key. With this cache is only marian primitives + the cache code.

The easiest way to get this into the existing source is during the preparation of a `Request`. `Segments = marian::Words` and we can have a stage during construction where we pre-fill the histories. 

https://github.com/browsermt/bergamot-translator/blob/cb855be1a7f3d362995b9528119549517a81a5eb/src/translator/request.cpp#L19-L29

The iteration mechanism provided to convert to a batch can now skip the entries where `Histories` are already computed (not `nullptr`). The rest of the pipeline (`Response` construction matching the `Request` etc) will follow.

This means an addition of `isCacheCompleted()` on `RequestSentence` relaying back to `isCacheCompleted(size_t index)` on `Request` for use in the batch preparation below.

https://github.com/browsermt/bergamot-translator/blob/cb855be1a7f3d362995b9528119549517a81a5eb/src/translator/batcher.cpp#L47-L54

The cache will have to be thread-safe since multiple async calls can potentially read simultaneously. A cache fetch will update recently used (write), so slapping a mutex for access is reasonable(?).

The cache will be populated on `processHistory` at `Request`:

https://github.com/browsermt/bergamot-translator/blob/cb855be1a7f3d362995b9528119549517a81a5eb/src/translator/request.cpp#L37-L47

cc @kpu @XapaJIaMnu

jerinphilip:
Fixes #201. WIP.

### DIFF
--- a/src/tests/apps.cpp
+++ b/src/tests/apps.cpp
@@ -50,6 +50,44 @@ void annotatedTextSentences(Ptr<Options> options, bool source) {
   }
 }
 
+void translationCache(Ptr<Options> options) {
+  // Prepare memories for bytearrays (including model, shortlist and vocabs)
+  MemoryBundle memoryBundle;
+  ResponseOptions responseOptions;
+
+  if (options->get<bool>("bytearray")) {
+    // Load legit values into bytearrays.
+    memoryBundle = getMemoryBundleFromConfig(options);
+  }
+
+  Service service(options, std::move(memoryBundle));
+
+  // Read a large input text blob from stdin
+  std::ostringstream inputStream;
+  inputStream << std::cin.rdbuf();
+  std::string input = inputStream.str();
+
+  auto translateForResponse = [&service, &responseOptions](std::string input) {
+    std::future<Response> responseFuture = service.translate(std::move(input), responseOptions);
+    responseFuture.wait();
+    Response response = responseFuture.get();
+    return response;
+  };
+
+  Response response;
+  auto &stats = service.cacheStats();
+
+  // Round 1
+  response = translateForResponse(input);
+  LOG(info, "Cache Hits/Misses = {}/{}", stats.hits, stats.misses);
+  ABORT_IF(stats.hits != 0, "Expecting no cache hits, but hits found.");
+
+  // Round 2; There should be cache hits
+  response = translateForResponse(input);
+  LOG(info, "Cache Hits/Misses = {}/{}", stats.hits, stats.misses);
+  ABORT_IF(stats.hits == 0, "No cache hits while expected non-zero");
+}
+
 }  // namespace testapp
 }  // namespace bergamot
 }  // namespace marian
--- a/src/translator/batcher.cpp
+++ b/src/translator/batcher.cpp
@@ -46,10 +46,12 @@ bool Batcher::cleaveBatch(Batch &batch) {
 
 void Batcher::addWholeRequest(Ptr<Request> request) {
   for (size_t i = 0; i < request->numSegments(); i++) {
-    RequestSentence sentence(i, request);
-    size_t bucket_id = sentence.numTokens();
-    assert(bucket_id < bucket_.size());
-    bucket_[bucket_id].insert(sentence);
+    if (!request->isCachePrefilled(i)) {
+      RequestSentence sentence(i, request);
+      size_t bucket_id = sentence.numTokens();
+      assert(bucket_id < bucket_.size());
+      bucket_[bucket_id].insert(sentence);
+    }
   }
 }
 
--- a/src/translator/cache.h
+++ b/src/translator/cache.h
@@ -1,3 +1,6 @@
+#ifndef SRC_BERGAMOT_CACHE_H_
+#define SRC_BERGAMOT_CACHE_H_
+
 #include <list>
 #include <mutex>
 #include <unordered_map>
@@ -57,3 +60,5 @@ typedef LRUCache<std::string, History> TranslatorLRUCache;
 
 }  // namespace bergamot
 }  // namespace marian
+
+#endif  // SRC_BERGAMOT_CACHE_H_
--- a/src/translator/cache.h
+++ b/src/translator/cache.h
@@ -12,7 +12,7 @@ namespace marian {
 namespace bergamot {
 
 /// Thread-safe LRUCache
-template <typename Key, typename CacheItem, typename Hash = std::hash<Key>, typename KeyEqual = std::equal_to<Key>>
+template <typename Key, typename CacheItem, typename Hash = std::hash<Key>>
 class LRUCache {
  public:
   // Storage includes Key, so when LRU is evicted corresponding hashmap entry can be deleted.
@@ -52,7 +52,7 @@ class LRUCache {
  private:
   size_t sizeCap_;
   std::list<StorageItem> storage_;
-  std::unordered_map<Key, StorageItr, Hash, KeyEqual> map_;
+  std::unordered_map<Key, StorageItr, Hash> map_;
   std::mutex rwMutex_;
 };
 
--- a/src/translator/cache.h
+++ b/src/translator/cache.h
@@ -17,8 +17,9 @@ class LRUCache {
   typedef typename std::list<StorageItem>::iterator StorageItr;
 
   LRUCache(size_t sizeCap) : sizeCap_(sizeCap) {}
+
   void insert(const Key key, const CacheItem item) {
-    std::lock_guard<std::recursive_mutex> guard(rwMutex_);
+    std::lock_guard<std::mutex> guard(rwMutex_);
     if (storage_.size() + 1 > sizeCap_) {
       // Evict LRU
       auto p = storage_.rbegin();
@@ -31,7 +32,7 @@ class LRUCache {
   }
 
   bool fetch(const Key key, CacheItem &item) {
-    std::lock_guard<std::recursive_mutex> guard(rwMutex_);
+    std::lock_guard<std::mutex> guard(rwMutex_);
     auto mapItr = map_.find(key);
     if (mapItr == map_.end()) {
       return false;
@@ -39,7 +40,8 @@ class LRUCache {
       auto storageItr = mapItr->second;
       item = storageItr->second;
       storage_.erase(storageItr);
-      insert(key, item);
+      storage_.push_front({key, item});
+      map_.insert({key, storage_.begin()});
       return true;
     }
   }
@@ -48,7 +50,7 @@ class LRUCache {
   size_t sizeCap_;
   std::list<StorageItem> storage_;
   std::unordered_map<Key, StorageItr> map_;
-  std::recursive_mutex rwMutex_;
+  std::mutex rwMutex_;
 };
 
 typedef LRUCache<std::string, History> TranslatorLRUCache;
--- a/src/translator/cache.h
+++ b/src/translator/cache.h
@@ -11,6 +11,11 @@
 namespace marian {
 namespace bergamot {
 
+struct LRUCacheStats {
+  size_t hits;
+  size_t misses;
+};
+
 /// Thread-safe LRUCache
 template <typename Key, typename CacheItem, typename Hash = std::hash<Key>>
 class LRUCache {
@@ -38,8 +43,10 @@ class LRUCache {
     std::lock_guard<std::mutex> guard(rwMutex_);
     auto mapItr = map_.find(key);
     if (mapItr == map_.end()) {
+      ++stats_.misses;
       return false;
     } else {
+      ++stats_.hits;
       auto storageItr = mapItr->second;
       item = storageItr->second;
       storage_.erase(storageItr);
@@ -49,11 +56,14 @@ class LRUCache {
     }
   }
 
+  const LRUCacheStats &stats() const { return stats_; }
+
  private:
   size_t sizeCap_;
   std::list<StorageItem> storage_;
   std::unordered_map<Key, StorageItr, Hash> map_;
   std::mutex rwMutex_;
+  LRUCacheStats stats_;
 };
 
 // Specialize for marian
--- a/src/translator/cache.h
+++ b/src/translator/cache.h
@@ -12,8 +12,8 @@ namespace marian {
 namespace bergamot {
 
 struct LRUCacheStats {
-  size_t hits;
-  size_t misses;
+  size_t hits{0};
+  size_t misses{0};
 };
 
 /// Thread-safe LRUCache
--- a/src/translator/cache.h
+++ b/src/translator/cache.h
@@ -1,0 +1,57 @@
+#include <list>
+#include <mutex>
+#include <unordered_map>
+#include <utility>
+
+#include "translator/beam_search.h"
+
+namespace marian {
+namespace bergamot {
+
+/// Thread-safe LRUCache
+template <typename CacheItem, typename Key>
+class LRUCache {
+ public:
+  // Storage includes Key, so when LRU is evicted corresponding hashmap entry can be deleted.
+  typedef std::pair<Key, CacheItem> StorageItem;
+  typedef typename std::list<StorageItem>::iterator StorageItr;
+
+  LRUCache(size_t sizeCap) : sizeCap_(sizeCap) {}
+  void insert(const Key key, const CacheItem item) {
+    std::lock_guard<std::recursive_mutex> guard(rwMutex_);
+    if (storage_.size() + 1 > sizeCap_) {
+      // Evict LRU
+      auto p = storage_.rbegin();
+      map_.erase(/*key=*/p->first);
+      storage_.pop_back();
+    }
+
+    storage_.push_front({key, item});
+    map_.insert({key, storage_.begin()});
+  }
+
+  bool fetch(const Key key, CacheItem &item) {
+    std::lock_guard<std::recursive_mutex> guard(rwMutex_);
+    auto mapItr = map_.find(key);
+    if (mapItr == map_.end()) {
+      return false;
+    } else {
+      auto storageItr = mapItr->second;
+      item = storageItr->second;
+      storage_.erase(storageItr);
+      insert(key, item);
+      return true;
+    }
+  }
+
+ private:
+  size_t sizeCap_;
+  std::list<StorageItem> storage_;
+  std::unordered_map<Key, StorageItr> map_;
+  std::recursive_mutex rwMutex_;
+};
+
+typedef LRUCache<std::string, History> TranslatorLRUCache;
+
+}  // namespace bergamot
+}  // namespace marian
--- a/src/translator/cache.h
+++ b/src/translator/cache.h
@@ -12,7 +12,7 @@ namespace marian {
 namespace bergamot {
 
 /// Thread-safe LRUCache
-template <typename CacheItem, typename Key>
+template <typename Key, typename CacheItem, typename Hash = std::hash<Key>, typename KeyEqual = std::equal_to<Key>>
 class LRUCache {
  public:
   // Storage includes Key, so when LRU is evicted corresponding hashmap entry can be deleted.
@@ -52,11 +52,26 @@ class LRUCache {
  private:
   size_t sizeCap_;
   std::list<StorageItem> storage_;
-  std::unordered_map<Key, StorageItr> map_;
+  std::unordered_map<Key, StorageItr, Hash, KeyEqual> map_;
   std::mutex rwMutex_;
 };
 
-typedef LRUCache<std::string, History> TranslatorLRUCache;
+// Specialize for marian
+// This is a lazy hash, we'll fix this with something better later.
+struct WordsHashFn {
+  size_t operator()(const Words &words) const {
+    std::string repr("");
+    for (size_t idx = 0; idx < words.size(); idx++) {
+      if (idx != 0) {
+        repr += " ";
+      }
+      repr += words[idx].toString();
+    }
+    return std::hash<std::string>{}(repr);
+  }
+};
+
+typedef LRUCache<Words, History, WordsHashFn> TranslatorLRUCache;
 
 }  // namespace bergamot
 }  // namespace marian
--- a/src/translator/request.cpp
+++ b/src/translator/request.cpp
@@ -14,17 +14,18 @@ namespace bergamot {
 Request::Request(size_t Id, Segments &&segments, ResponseBuilder &&responseBuilder, TranslatorLRUCache &cache)
     : Id_(Id),
       segments_(std::move(segments)),
-      responseBuilder_(std::move(responseBuilder)) cache_(cache)
+      responseBuilder_(std::move(responseBuilder)),
+      cache_(cache)
 
 {
   counter_ = segments_.size();
   histories_.resize(segments_.size(), nullptr);
 
   for (size_t idx = 0; idx < segments_.size(); idx++) {
-    Ptr<History> history = new History();
-    if (cache_.fetch(getSegment(idx), *history)) {
-      histories_[idx] = history;
-      --counter;
+    History history(/*lineNo=*/idx);
+    if (cache_.fetch(getSegment(idx), history)) {
+      histories_[idx] = New<History>(history);
+      --counter_;
     }
   }
 
--- a/src/translator/request.cpp
+++ b/src/translator/request.cpp
@@ -50,6 +50,9 @@ void Request::processHistory(size_t index, Ptr<History> history) {
   // ready. The container storing histories is set with the value obtained.
   histories_[index] = history;
 
+  // Update cache
+  cache_.insert(getSegment(index), *history);
+
   // In case this is last request in, completeRequest is called, which sets the
   // value of the promise.
   if (--counter_ == 0) {
--- a/src/translator/request.cpp
+++ b/src/translator/request.cpp
@@ -11,14 +11,22 @@ namespace marian {
 namespace bergamot {
 
 // -----------------------------------------------------------------
-Request::Request(size_t Id, Segments &&segments, ResponseBuilder &&responseBuilder)
+Request::Request(size_t Id, Segments &&segments, ResponseBuilder &&responseBuilder, TranslatorLRUCache &cache)
     : Id_(Id),
       segments_(std::move(segments)),
-      responseBuilder_(std::move(responseBuilder))
+      responseBuilder_(std::move(responseBuilder)) cache_(cache)
 
 {
   counter_ = segments_.size();
   histories_.resize(segments_.size(), nullptr);
+
+  for (size_t idx = 0; idx < segments_.size(); idx++) {
+    Ptr<History> history = new History();
+    if (cache_.fetch(getSegment(idx), *history)) {
+      histories_[idx] = history;
+      --counter;
+    }
+  }
 
   // If there are no segments_, we are never able to trigger the responseBuilder
   // calls from a different thread. However, in this case we want an empty valid
@@ -27,6 +35,8 @@ Request::Request(size_t Id, Segments &&segments, ResponseBuilder &&responseBuild
     responseBuilder_(std::move(histories_));
   }
 }
+
+bool Request::isCachePrefilled(size_t index) { return (histories_[index] != nullptr); }
 
 size_t Request::numSegments() const { return segments_.size(); }
 
@@ -72,6 +82,8 @@ bool operator<(const RequestSentence &a, const RequestSentence &b) {
   }
   return a.request_ < b.request_;
 }
+
+bool RequestSentence::isCachePrefilled() { return request_->isCachePrefilled(index_); }
 
 // ----------------------------------------------------------------------
 
--- a/src/translator/request.cpp
+++ b/src/translator/request.cpp
@@ -29,10 +29,11 @@ Request::Request(size_t Id, Segments &&segments, ResponseBuilder &&responseBuild
     }
   }
 
-  // If there are no segments_, we are never able to trigger the responseBuilder
-  // calls from a different thread. However, in this case we want an empty valid
-  // response.
-  if (segments_.size() == 0) {
+  // If there are no segments_, we are never able to trigger the responseBuilder calls from a different thread. However,
+  // in this case we want an empty valid response. Also, if cache somehow manages to decrease all counter prefilling
+  // histories, then we'd have to trigger ResponseBuilder as well. No segments go into batching and therefore no
+  // processHistory triggers.
+  if (segments_.size() == 0 || counter_.load() == 0) {
     responseBuilder_(std::move(histories_));
   }
 }
--- a/src/translator/request.h
+++ b/src/translator/request.h
@@ -46,7 +46,7 @@ class Request {
   /// @param [in] responseBuilder: Callback function (of ResponseBuilder type)
   /// to be triggered upon the completion of translation of all units in a
   /// Request.
-  Request(size_t Id, Segments &&segments, ResponseBuilder &&responseBuilder);
+  Request(size_t Id, Segments &&segments, ResponseBuilder &&responseBuilder, TranslatorLRUCache &cache);
 
   /// Obtain the count of tokens in the segment correponding to index. Used to
   /// insert sentence from multiple requests into the corresponding size bucket.
@@ -67,6 +67,9 @@ class Request {
   /// compiled from requests.
   void processHistory(size_t index, Ptr<History> history);
 
+  /// Check if History has been prefilled from cache_
+  bool isCachePrefilled(size_t index);
+
  private:
   size_t Id_;
 
@@ -86,6 +89,9 @@ class Request {
   /// Constructing Response requires the vocabs_ used to generate Request.
   /// std::vector<Ptr<Vocab const>> *vocabs_;
   ResponseBuilder responseBuilder_;
+
+  /// Reference to cache to prefill / populate
+  TranslatorLRUCache &cache;
 };
 
 /// A RequestSentence provides a view to a sentence within a Request. Existence
@@ -106,6 +112,9 @@ class RequestSentence {
   /// Forwards history to Request to set history corresponding to this
   /// RequestSentence.
   void completeSentence(Ptr<History> history);
+
+  /// Check if this RequestSentence is already prefilled with a History from cache.
+  bool isCachePrefilled();
 
   friend bool operator<(const RequestSentence &a, const RequestSentence &b);
 
--- a/src/translator/request.h
+++ b/src/translator/request.h
@@ -6,6 +6,7 @@
 #include <vector>
 
 #include "annotation.h"
+#include "cache.h"
 #include "common/logging.h"
 #include "data/types.h"
 #include "definitions.h"
@@ -91,7 +92,7 @@ class Request {
   ResponseBuilder responseBuilder_;
 
   /// Reference to cache to prefill / populate
-  TranslatorLRUCache &cache;
+  TranslatorLRUCache &cache_;
 };
 
 /// A RequestSentence provides a view to a sentence within a Request. Existence
--- a/src/tests/apps.h
+++ b/src/tests/apps.h
@@ -33,6 +33,8 @@ void annotatedTextWords(Ptr<Options> options, bool source = true);
 // in each line, depending on source = true or false respectively.
 void annotatedTextSentences(Ptr<Options> options, bool source = true);
 
+void translationCache(Ptr<Options> options);
+
 }  // namespace testapp
 }  // namespace bergamot
 }  // namespace marian
--- a/src/tests/cli.cpp
+++ b/src/tests/cli.cpp
@@ -14,6 +14,8 @@ int main(int argc, char *argv[]) {
     testapp::annotatedTextWords(options, /*source=*/true);
   } else if (mode == "test-response-target-words") {
     testapp::annotatedTextWords(options, /*source=*/false);
+  } else if (mode == "test-translation-cache") {
+    testapp::translationCache(options);
   } else {
     ABORT("Unknown --mode {}. Please run a valid test", mode);
   }
--- a/src/tests/units/CMakeLists.txt
+++ b/src/tests/units/CMakeLists.txt
@@ -1,6 +1,7 @@
 # Unit tests
 set(UNIT_TESTS
     annotation_tests
+    cache_tests
 )
 
 foreach(test ${UNIT_TESTS})
--- a/src/tests/units/cache_tests.cpp
+++ b/src/tests/units/cache_tests.cpp
@@ -1,0 +1,90 @@
+#include <random>
+#include <vector>
+
+#include "catch.hpp"
+#include "translator/cache.h"
+
+using marian::bergamot::LRUCache;
+
+const int SIZE = 1000;
+
+TEST_CASE("LRU Cache single-thread") {
+  typedef LRUCache<int, int> Cache;
+
+  int size = SIZE;
+  Cache cache(size);
+
+  auto identity = [](int key) { return key; };
+  int val = -1;
+
+  // First pass
+  for (int key = 0; key < size; key++) {
+    CHECK(!cache.fetch(key, val));  // cache-miss
+    cache.insert(key, identity(key));
+  }
+
+  // Second pass
+  for (int key = 0; key < size; key++) {
+    CHECK(cache.fetch(key, val));  // cache-hit
+    CHECK(val == identity(key));
+  }
+
+  // Add more
+  for (int key = size; key < 2 * size; key++) {
+    CHECK(!cache.fetch(key, val));  // cache-miss
+    cache.insert(key, identity(key));
+  }
+
+  for (int key = 0; key < size; key++) {
+    CHECK(!cache.fetch(key, val));  // cache-miss due to eviction
+  }
+}
+
+TEST_CASE("LRU Cache multi-thread") {
+  typedef LRUCache<int, int> Cache;
+
+  int size = SIZE;
+  Cache cache(size);
+
+  // cache.debug();
+  std::vector<std::thread> threads;
+
+  auto f = [](int key) { return key; };
+  int val = -1;
+  auto fn = [&cache, &f](int key) { cache.insert(key, f(key)); };
+
+  std::cout << threads.size() << std::endl;
+
+  // First pass
+  for (int key = 0; key < size; key++) {
+    CHECK(!cache.fetch(key, val));  // cache-miss
+    threads.emplace_back(fn, key);
+  }
+
+  std::cout << threads.size() << std::endl;
+
+  // Barrier
+  for (auto &worker : threads) {
+    worker.join();
+  }
+
+  threads.clear();
+  // cache.debug();
+
+  // Second pass
+  for (int key = 0; key < size; key++) {
+    if (cache.fetch(key, val)) {  // cache-hit, but threads mean ordering can lead to weird evictions.
+      CHECK(val == f(key));
+    }
+  }
+
+  // Add more
+  for (int key = size; key < 2 * size; key++) {
+    CHECK(!cache.fetch(key, val));  // cache-miss
+    cache.insert(key, f(key));
+  }
+
+  for (int key = 0; key < size; key++) {
+    CHECK(!cache.fetch(key, val));  // cache-miss due to eviction
+  }
+}
--- a/src/translator/parser.h
+++ b/src/translator/parser.h
@@ -29,6 +29,8 @@ inline marian::ConfigParser createConfigParser() {
   cp.addOption<std::string>("--bergamot-mode", "Bergamot Options",
                             "Operating mode for bergamot: [wasm, native, decoder]", "native");
 
+  cp.addOption<size_t>("--cache-size", "Bergamot Options", "Number of histories to keep in translation cache", 20);
+
   return cp;
 }
 
--- a/src/translator/service.cpp
+++ b/src/translator/service.cpp
@@ -18,7 +18,7 @@ Service::Service(Ptr<Options> options, MemoryBundle memoryBundle)
       numWorkers_(std::max<int>(1, options->get<int>("cpu-threads"))),
       modelMemory_(std::move(memoryBundle.model)),
       shortlistMemory_(std::move(memoryBundle.shortlist)),
-      cache_(options->get<size_t>("cache-size")
+      cache_(options->get<size_t>("cache-size"))
 #ifdef WASM_COMPATIBLE_SOURCE
       ,
       blocking_translator_(DeviceId(0, DeviceType::cpu), vocabs_, options_, &modelMemory_, &shortlistMemory_)
@@ -87,7 +87,7 @@ std::future<Response> Service::queueRequest(std::string &&input, ResponseOptions
   auto future = responsePromise.get_future();
 
   ResponseBuilder responseBuilder(responseOptions, std::move(source), vocabs_, std::move(responsePromise));
-  Ptr<Request> request = New<Request>(requestId_++, std::move(segments), std::move(responseBuilder));
+  Ptr<Request> request = New<Request>(requestId_++, std::move(segments), std::move(responseBuilder), cache_);
 
   batcher_.addWholeRequest(request);
   return future;
--- a/src/translator/service.cpp
+++ b/src/translator/service.cpp
@@ -17,7 +17,7 @@ Service::Service(Ptr<Options> options, MemoryBundle memoryBundle)
       batcher_(options),
       numWorkers_(std::max<int>(1, options->get<int>("cpu-threads"))),
       modelMemory_(std::move(memoryBundle.model)),
-      shortlistMemory_(std::move(memoryBundle.shortlist))
+      shortlistMemory_(std::move(memoryBundle.shortlist)),
       cache_(options->get<size_t>("cache-size")
 #ifdef WASM_COMPATIBLE_SOURCE
       ,
--- a/src/translator/service.cpp
+++ b/src/translator/service.cpp
@@ -18,6 +18,7 @@ Service::Service(Ptr<Options> options, MemoryBundle memoryBundle)
       numWorkers_(std::max<int>(1, options->get<int>("cpu-threads"))),
       modelMemory_(std::move(memoryBundle.model)),
       shortlistMemory_(std::move(memoryBundle.shortlist))
+      cache_(options->get<size_t>("cache-size")
 #ifdef WASM_COMPATIBLE_SOURCE
       ,
       blocking_translator_(DeviceId(0, DeviceType::cpu), vocabs_, options_, &modelMemory_, &shortlistMemory_)
--- a/src/translator/service.h
+++ b/src/translator/service.h
@@ -115,7 +115,7 @@ class Service {
   bool isAlignmentSupported() const { return options_->hasAndNotEmpty("alignment"); }
 
   /// Returns cache stats
-  const LRUCacheStats &stats() const { return cache_.stats(); }
+  const LRUCacheStats &cacheStats() const { return cache_.stats(); }
 
  private:
   /// Queue an input for translation.
--- a/src/translator/service.h
+++ b/src/translator/service.h
@@ -2,6 +2,7 @@
 #define SRC_BERGAMOT_SERVICE_H_
 
 #include "batch_translator.h"
+#include "cache.h"
 #include "data/types.h"
 #include "response.h"
 #include "response_builder.h"
@@ -148,6 +149,9 @@ class Service {
   /// Batcher handles generation of batches from a request, subject to
   /// packing-efficiency and priority optimization heuristics.
   ThreadsafeBatcher batcher_;
+
+  /// LRUCache, threadsafe.
+  TranslatorLRUCache cache_;
 
   // The following constructs are available providing full capabilities on a non
   // WASM platform, where one does not have to hide threads.
--- a/src/translator/service.h
+++ b/src/translator/service.h
@@ -114,6 +114,9 @@ class Service {
   /// Returns if model is alignment capable or not.
   bool isAlignmentSupported() const { return options_->hasAndNotEmpty("alignment"); }
 
+  /// Returns cache stats
+  const LRUCacheStats &stats() const { return cache_.stats(); }
+
  private:
   /// Queue an input for translation.
   std::future<Response> queueRequest(std::string &&input, ResponseOptions responseOptions);


src/translator/request.h
```diff
@@ -107,6 +114,9 @@ class RequestSentence {
   /// RequestSentence.
   void completeSentence(Ptr<History> history);
 
+  /// Check if this RequestSentence is already prefilled with a History from cache.
+  bool isCachePrefilled();
```
jerinphilip, line 118: note to self - unused, remove.

kpu:
If we don't care too much about redoing some work, why not a direct mapped cache with atomic pointers?

jerinphilip:
So... I spent a lot of time digging up theory on lock-free hashmaps, primarily because after stripping boost L4 was only giving 1/4 cache hits, while I expected 4/4, and reading source turned out to be not very pleasant.

L4 is lock-free on read. On write, it takes locks, but there are shards from what I gather. There's an epoch-based reclamation strategy for unreachable value-types with a dedicated thread running in the background every specified processing interval to collect garbage.

Storage in L4 is for arbitrary bytes of Keys and Values `Bytes = (data*, size)` requires serialization, so `marian::History` which is more "lazy" is converted into something "eager" to keep a lid on storage and to keep serialization code minimum. L4, unlike other alternatives seem to offer config to keep a lid on memory usage as well. 

WASM builds seem to be failing with int64/size_t issues which can eventually be fixed(?). There are bunch of L4 constructs I stripped down trying to remove boost which I don't fully understand yet. The changes can be found at https://github.com/browsermt/L4/pull/1. 

I'm not very happy with L4 from a readability/maintenence perspective. Alternatives I find nicer to at least read are [folly/ConcurrentHashMap](https://github.com/facebook/folly/blob/master/folly/concurrency/ConcurrentHashMap.h) (has a lot of dependencies, might be a bit hard to pull a minimum required hashmap; uses hazard pointers), [xenium](https://github.com/mpoeter/xenium) (this looks more academic; offers many variants of reclamation, minimal dependencies).

> why not a direct mapped cache with atomic pointers?

@kpu CAS only on pointers? Is there something that allows freeing the memory being pointed to while access being atomic when multiple threads are trying to access? If this was trivial there would be implementations out there would there not be?

jerinphilip:
The edit workflow [benchmarked with/without cache](https://github.com/jerinphilip/bergamot-translator/pull/13) indicates with cache it takes the following time to edit-type through [c10k.html](http://www.kegel.com/c10k.html). There are about 600 lines in the HTML. A simulated experiment measures the time taken to type in the whole document making a few mistakes along the way. The constructed string so far is sent to be translated on a simulated "onKeyUp". 

| Cache     |  Hits | Misses | WallTime |
| --------- | ------| ------ | --------:|
| without   | -     |   -    |  1226.80s|
| with (L4) | 1764424|  5363 |  633.46s |

This is without any sentence-split modifications, #102 should improve on these numbers even further when the client takes up some partitioning. 

* L4 cannot be used with Web Assembly unless threads become active. The garbage-collector for each epoch [runs in a background thread ](https://github.com/microsoft/L4/blob/9968fd0f07dc106c46b603ec867d229bb425861e/inc/L4/LocalMemory/EpochManager.h#L35)and is difficult to workaround.
* I'm using streams for writing bytes, which I expect will have to be replaced with something more suitable (I'm not sure what's the best here). We could use some common binary-io primitive utils across cache, load-from-byte-array (right now these are anonymous and in respective CPP files). 
* I potentially see use-cases for iterators in C++ API built from `Annotation` (Iterate through word/tokens: `WordsIterator`, iterate through sentences, `SentenceIterator`)

jerinphilip:
Few things to note:

* https://github.com/browsermt/L4/pull/1: Removing boost from L4 (potentially) is also associated with this PR and could use some review attention.
* https://github.com/jerinphilip/bergamot-translator/pull/13: Benchmarking with/without cache uses a workflow of heavier workload and could be useful like speed-tests (WNGT20) which run on _var_ to catch anything that only appears at times and complex thread-ops. This can potentially be merged into this PR and integrated into BRT.
* Note that `WASM_COMPATIBLE_SOURCE` path is **not** tested under heavy workload. Primarily because I can't be bothered to put large input data into single-thread and wait for results until the next day. However the simpler tests which run on GitHub CI run for both at this time. 
* There are possibly **privacy implications**, say if extension is enabled in private browsing cache might have data saved to disk (while the intention is to not retain any information like browsing history etc locally).

jerinphilip:
Based on an offline discussion today, cache dev (this PR) is on pause until review attention is available. The following however needs eventual looking into:

- Concern is lot of copies, hula-hoops and recomputations going around which needs to be minimized. We are at least looking at reconstructing the complete output target sentence for every request which can be avoided with careful design. We will need the ability to save and reuse work somehow (somehow) on target sentence, annotation at least which are wastefully recomputed as a whole at the moment.
- Some discussions about optimizing Key storage and introducing hash-optimizations (will have to look into this). I believe the gist was work with collision probabilities to simplify Key storage (which is currently `marian::Words` moving to `(model-identifier, marian::Words)` to something occupying lesser space but still avoiding collisions).

XapaJIaMnu:
I have started looking at this, but it will take me some days to digest as it's a lot. Several preliminary questions:

First and foremost, thank you for the bulk of work. It's a lot to do and getting a functional boostless implementation is a great milestone!

Why is hashing done as it is? Instead of producing one big string and then computing its hash for no reason, you should instead use an implementation of `hash_combine` https://stackoverflow.com/questions/2590677/how-do-i-combine-hash-values-in-c0x to   get the hash of a sequence of words, without generating a useless string. Marian also has a variant of this implemented somewhere (I think it is just a copy from the boost code).

Could you produce a benchmark of translating 1M non-repeating sentences (sources.shuf) with and without cache enabled so that we can see what is the overhead of caching. Could you do that with 1 thread (reduce num sentences to 100k for example) and 16 threads so that we can see impact of lock contention. Please report sys time alongside realtime.

src/translator/cache.cpp
```diff
@@ -0,0 +1,171 @@
...
+  }
+};
+
+#endif
+
+ThreadUnsafeLRUCache::ThreadUnsafeLRUCache(Ptr<Options> options)
+    : storageSizeLimit_(options->get<size_t>("cache-size") * 1024 * 1024), storageSize_(0) {}
+
+bool ThreadUnsafeLRUCache::fetch(const marian::Words &words, ProcessedRequestSentence &processedRequestSentence) {
+  auto p = cache_.find(words);
```
XapaJIaMnu, line 126: Where is this defined, sorry (I don't have it open in an editor yet)

src/translator/request.h
XapaJIaMnu: You have multiple `PreFilled` methods

src/translator/request.h
```diff
@@ -3,12 +3,18 @@
 
 #include <cassert>
 #include <future>
+#include <iostream>
```
XapaJIaMnu, line 6: Are those used or are they leftover from previous implementation?

XapaJIaMnu (changes requested):
Sorry for the late reply. I have more parts to review, but I'll do that after I get feedback from you on those.

src/translator/cache.cpp
```diff
@@ -0,0 +1,171 @@
+#include "translator/cache.h"
+
+#include <cstdlib>
+
+namespace marian {
+namespace bergamot {
+
+namespace cache_util {
+
+std::string wordsToString(const marian::Words &words) {
```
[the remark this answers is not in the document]
jerinphilip: Yes. L4 is a black box with Bytes for keys. We need `(*data, size)`. Could alternatively work with `(words.data(), words.size()*sizeof(marian::Word))`

Do we try that?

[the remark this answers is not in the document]
XapaJIaMnu: How about the byte array that is the hash? It's uint64_t so that 8 int8_t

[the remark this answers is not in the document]
XapaJIaMnu: I don't want to copy huge strings

jerinphilip:
I'll put some English explanation.

We deal with `Request`s, each of which correspond to a blob of text which is to be translated. A `Response` with the blob of translated text and additional information is sent back, for each `Request`. The blob is sentence-split, and wrapped to a max-length and the unit obtained corresponds to a `RequestSentence` constituted by `marian::Words`, and has a counterpart of a `marian::History` in marian terms.

**Where are we placing Cache?** We have (or had) plans to compile multiple requests into the same batch. How this is accomplished is by creating `RequestSentence`, which has a backref to a `Request`. The batches complete whatever `RequestSentence`s that are in them, and each completion of a `RequestSentence` also checks for whether the `Request` is complete (by means of an atomic counter). If it is complete, it fires the `Response` to be constructed corresponding to the request. Cache is set at a `RequestSentence` level.

When a new `Request` comes in - this could be incremental typing for `translateLocally` or outbound translation for extension. This could also be back navigation from a page as well. This is broken down into `RequestSentence`s, while the unique key is the constituent `marian::Words`. For the `marian::Words`s we can find from the cache, we fill the corresponding translated unit (`marian::History`, which we convert to a `ProcessedRequestSentence` being thinner), decrement the counter accordingly, and claim it's translated. The remaining `RequestSentence`s only go in for "actual" translation. 

**What do we cache?** A `(Key, Value)` pair for our use-case is `(marian::Words, History)`. However, History is a scary bloated thing, so we have a thinner cache-friendly equivalent, which stores only the parts we need to construct a `Response`. Until I can find a better name, this is called `ProcessedRequestSentence`.

Thus, `(Key, Value) = (marian::Words, ProcessedRequestSentence)`.  Now, a `Record` is always `(Key, Value)` unless there is perfect hashing. My understanding of hashing is that in the event of collisions, I have to check if the key put in matches with the actual key (hence its place in the record) because hashes can collide. If I don't match the key, I will get wrong translations - there is a risk. @XapaJIaMnu seems to want to replace this with hashes from what I gather, @kpu wanted in an offline discussion to look into collision probabilities and reduce the size of the Key potentially to optimize storage. I think for the time being at least in comparison to `ProcessedRequestSentence`, `marian::Words` isn't as costly to warrant extreme-optimization here. 

**Existing Implementations** There are two caches, primarily because the recommended cache - L4 has a background thread for epoch-based reclamation (EBR) - this cannot run in single-threaded blocking WASM.

* L4 offers as the README says a cache for arbitrary streams of bytes for keys and bytes for values. It internally hashes the keys, if I hash it before, I'm double hashing. Also collisions. L4 stores a Record type as well to check collisions ([Record.h](https://github.com/microsoft/L4/blob/9968fd0f07dc106c46b603ec867d229bb425861e/inc/L4/HashTable/Common/Record.h))
* To get code working for WASM, there's an excuse for an LRU Cache put in. Implementation is linked-list for Records to handle order, unordered_map to do O(1) access. Since extension folks have no plan to have multithreading, all the while having plans for 2x slowdowns, I'm starting to think I'll end up going back and optimizing this bit as well. 

I would not mind spending time if I have it to replace both with hazard-pointer-based Cache, but that's a lot of work and intensive checks. Folly has a [reference implementation](https://github.com/facebook/folly/blob/master/folly/synchronization/Hazptr.h) which if isolated, could be used. It supports [blocking as well](https://github.com/facebook/folly/blob/a782676bbc784f5cfaae653f5cc11e657258bf9b/folly/synchronization/Hazptr.h#L43). 

**Other Known Performance Concerns**: Currently we process sentences again as a whole, and construct response as a whole to get to client. So things get worse as the entire blob is sent and translation back say when "CRIME and PUNISHMENT" is typed in. There may be ways to improve here by delaying construction of Response until absolutely needed. However, I'm skeptical about this, as this would generate a lot of additional spaghetti here, and is better off at client solving sentence-splits when the splitting library becomes exposed. 

I will respond to the queries individually, but I hope the above help have a mental map going in.

src/translator/request.h
jerinphilip: I think the serializing for `Value = ProcessedRequestSentence` happens with ostringstream. We could change this though because the size of a `ProcessedRequestSentence` is known at start.

src/translator/cache.h
```diff
@@ -0,0 +1,171 @@
...
+///
+/// Keys are marian::Words, values are ProcessedRequestSentences.
+///
+/// There is a serialization (converting to binary (data*, size)) of structs overhead with this class as well, which
+/// helps keep tight lid on memory usage. This should however be cheaper in comparison to recomputing through the graph.
+
+#ifndef WASM_COMPATIBLE_SOURCE
+class ThreadSafeL4Cache {
+ public:
+  using KeyBytes = L4::IWritableHashTable::Key;
```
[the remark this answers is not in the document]
jerinphilip: Sensible thing to do, will add this soon.

src/translator/cache.cpp
```diff
@@ -0,0 +1,171 @@
...
+  auto removeCandidatePtr = storage_.begin();
+
+  // Loop until not end and we have not found space to insert candidate adhering to configured storage limits.
+  while (storageSize_ + candidate.size() > storageSizeLimit_ && removeCandidatePtr != storage_.end()) {
+    storageSize_ -= removeCandidatePtr->size();
+    storage_.erase(removeCandidatePtr);
+    ++removeCandidatePtr;
+  }
+
+  // Only insert new candidate if we have space after adhering to storage-limit
```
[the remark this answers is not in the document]
jerinphilip: I see these metrics in terms of sentences are useful. I will look into checking in code to log these.

src/translator/request.h
XapaJIaMnu: Headers should be included in the file that needs them

src/tests/apps.cpp
```diff
@@ -55,6 +55,66 @@ void annotatedTextSentences(Ptr<Options> options, bool source) {
...
+    // Load legit values into bytearrays.
+    memoryBundle = getMemoryBundleFromConfig(options);
+  }
+
+  Service service(options, std::move(memoryBundle));
+
+  // Read a large input text blob from stdin
+  std::ostringstream inputStream;
+  inputStream << std::cin.rdbuf();
+  std::string input = inputStream.str();
```
kpu, line 73: What is this and why would I want it?

src/translator/processed_request_sentence.cpp
```diff
@@ -0,0 +1,102 @@
...
+namespace marian {
+namespace bergamot {
+
+namespace {
+
+// Generic write/read from pointers
+
+// write uses ostream, to get bytes/blob use with ostringstream(::binary)
+template <class T>
+void write(std::ostream &out, const T *data, size_t num = 1) {
```
kpu, line 15: I think you're using ostream in performance-critical code?  ostream is crazy slow.

kpu: Also this method name is bad, since there is a C write function.

src/translator/processed_request_sentence.cpp
```diff
@@ -0,0 +1,102 @@
...
+const char *copyInAndAdvance(const char *src, T *dest, size_t num = 1) {
+  const void *vsrc = reinterpret_cast<const void *>(src);
+  void *vdest = reinterpret_cast<void *>(dest);
+  std::memcpy(vdest, vsrc, num * sizeof(T));
+  return reinterpret_cast<const char *>(src + num * sizeof(T));
+}
+
+// Specializations to read and write vectors. The format stored is [size, v_0, v_1 ... v_size]
+template <class T>
+void writeVector(std::ostream &out, std::vector<T> v) {
```
kpu, line 32: These format conversions make me sad.  
If you want/need a flat format, then just make it a flat format.  And wrap a class around it that accesses the flat format.  Then remove the non-flat format.

src/translator/processed_request_sentence.h
```diff
@@ -0,0 +1,68 @@
...
+/// cache, due to copies being shallow rather than deep. We therefore process marian's lazy History management changing
+/// to something eager, keeping only what we want and units accounted for. Each of these corresponds to a
+/// `RequestSentence` and retains the minimum information required to reconstruct a full `Response` if cached.
+///
+/// L4 requires byte-streams, which allows for it to keep tight control over "memory" we need to have accounting for
+/// cache-size based on memory/bytes as well for any arbitrary serializable into bytestream data-types.
+///
+/// An empty() method is provided which relays if no words are present. This empty type is also treated as  an
+/// incomplete ProcessedRequestSentence during cache operations.
+class ProcessedRequestSentence {
```
kpu, line 23: This whole class can be replaced with just always having the byte representation and accessing it directly, rather than serialization and deserialization.

src/translator/request.h
```diff
@@ -3,12 +3,18 @@
...
+#include <istream>
+#include <sstream>
```
kpu, line 8: These kill performance.

src/translator/request.cpp
```diff
@@ -11,38 +11,68 @@ namespace marian {
...
+    processedRequestSentences_.resize(segments_.size());
+
+    if (cache_ != nullptr) {
+      // Iterate through segments, see if any can be prefilled from cache. If prefilled, mark the particular segments as
+      // complete (non-empty ProcessedRequestSentence). Also update accounting used elsewhere (counter_) to reflect one
+      // less segment to translate.
+      for (size_t idx = 0; idx < segments_.size(); idx++) {
+        ProcessedRequestSentence processedRequestSentence;
+        if (cache_->fetch(getSegment(idx), processedRequestSentence)) {
+          processedRequestSentences_[idx] = processedRequestSentence;
```
kpu, line 38: Why write somewhere else then copy when you could have written directly?

src/translator/processed_request_sentence.h
kpu:
```
class ProcessedRequestSentence {
  public:
    float sentenceScore() const { return reinterpret_cast<const float*>(base_); }
    // Could have an object with begin and end methods that looks like a vector but is actually just a range.  
    const marian::Word *words() const;
    const float *wordScores() const;
    const float *alignmentMatrix() const;
    std::size_t length() const;
  private:
    void *base_;
};
```

src/translator/request.cpp
jerinphilip: This started off as a `Ptr<...>` when it was cheap, I see the issue and the merits of the recommendation. Will fix.

src/translator/processed_request_sentence.cpp
```diff
@@ -0,0 +1,102 @@
...
+// responsibility of the caller to prepare the container with the correct contiguous size so memcpy works correctly.
+template <class T>
+const char *copyInAndAdvance(const char *src, T *dest, size_t num = 1) {
+  const void *vsrc = reinterpret_cast<const void *>(src);
+  void *vdest = reinterpret_cast<void *>(dest);
+  std::memcpy(vdest, vsrc, num * sizeof(T));
+  return reinterpret_cast<const char *>(src + num * sizeof(T));
+}
+
+// Specializations to read and write vectors. The format stored is [size, v_0, v_1 ... v_size]
```
kpu, line 30:
```
template <class T> class FlatVector {
  public:
    // Restore from existing
    explicit FlatVector(void *mem) {
     std::size_t size = *reinterpret_cast<const std::size_t*>(mem);
      begin_ = reinterpret_cast<T*>(reinterpret_cast<char*>(mem) + sizeof(std::size_t));
      end_ = begin_ + size;
    }
    // Initialize from new
    FlatVector(void *mem, std::size_t size) {
     *reinterpret_cast<std::size_t*>(mem) = size;
      begin_ = reinterpret_cast<T*>(reinterpret_cast<char*>(mem) + sizeof(std::size_t));
      end_ = begin_ + size;
    }

    const T *begin() const { return begin_; }
    const T *end() const { return end_; }
    T *begin() { return begin_; }
    T *end() { return end_; }

    std::size_t MemSize() { return sizeof(std::size_t) + (end_ - begin_) * sizeof(T); }

  private:
    T* begin_, end_;
};
```

src/translator/processed_request_sentence.h
jerinphilip: Since the general concern here is conversions and copies...

I'm on board with the idea here. There are, however, a bunch of ownership concerns. L4 owns `Key` and `Value`. I have to `memcpy` `Value = (data*, size)` (as a binary-sequence and have `ProcessRequestSentence` own the copied entity) because the cache (value) entry can be evicted potentially in a background thread once fetch is complete and the element is RAII Object is out of scope. I don't want to leak L4 all the way up. 

There are still (unavoidable) copies when we chop down `marian::History` (freshly computed, not cached) to be thinner and then written onto something flat and sequential (as L4 requires continuous byte-sequence for storing Value). 

With words put on to this flat-storage:


```cpp 
    const marian::Word *words() const;
	const marian::Word *words_begin_;
```

Marian decoding API (https://github.com/browsermt/marian-dev/blob/62bac858bfd37060beb707d12eb9711649ea4cf6/src/data/sentencepiece_vocab.cpp#L259) works with `marian::Words & == vector<marian::Word> &`. The standard vector has to own the underlying storage I can't find a way to override and provide a pseudo const vector that uses the sequential binary flat thing as storage.

I'm looking at adding the `(words*, size)` decoding API (which involve edits to marian-dev) to avoid a copy here. An alternative is a copy onto vector at decoding, which I will do to check things stay intact before spilling over to more repositories (for now at least).

Please let know if I'm missing something or some understanding is wrong.

For now, I have created binary storage within `ProcessedRequestSentence` which is the flat thing asked for and forbid copy on storage, which in-turn forbids copy on `ProcessedRequestSentence` which should eliminate unnecessary copies. Conversions however still exist.

src/translator/request.h
jerinphilip: Headers are removed.

src/translator/request.cpp
jerinphilip: Copy forbidden, now direct move happening.

src/translator/cache.h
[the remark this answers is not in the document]
jerinphilip: I've added a description in comments.

src/translator/cache.h
```diff
@@ -0,0 +1,171 @@
...
+  ///
+  /// @param [in] words: marian::Words processed from a sentence
+  /// @param [in] processedRequestSentence: minimum translated information corresponding to words
+  void insert(const marian::Words &words, const ProcessedRequestSentence &processedRequestSentence);
+
+  CacheStats stats() const;
+
+ private:
+  struct Record {
+    marian::Words key;
```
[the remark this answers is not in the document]
jerinphilip: There's no comparison because of hash_fn, assuming we "probably" can and want to optimize for storage, will replace these with `uint64_t`. 

I will throw this (LRU) cache out for a hazard-pointer-based implementation eventually when I can prioritize this.

src/translator/batcher.cpp
```diff
@@ -46,10 +46,12 @@ bool Batcher::cleaveBatch(Batch &batch) {
...
   for (size_t i = 0; i < request->numSegments(); i++) {
-    RequestSentence sentence(i, request);
-    size_t bucket_id = sentence.numTokens();
-    assert(bucket_id < bucket_.size());
-    bucket_[bucket_id].insert(sentence);
+    if (!request->isCachePrefilled(i)) {
+      RequestSentence sentence(i, request);
+      size_t bucket_id = sentence.numTokens();
+      assert(bucket_id < bucket_.size());
+      bucket_[bucket_id].insert(sentence);
```
jerinphilip, line 53: > Are sentences bucketed based on of the number of tokens they have. Some comments (:

I'm confused. Is the above not self-explanatory?

src/translator/request.h
jerinphilip: Removed.

src/translator/cache.h
```diff
@@ -0,0 +1,183 @@
...
+/// structures are deleted, which runs in the background using an epoch based memory reclamation (EBR) scheme. Eviction
+/// follows the "Clock" algorithm, which is a practical approximation for LRU. Records live either until they're used at
+/// least once, or if there is no space left without removing one that's not been used yet, with removal triggered to
+/// make space for new insertions.
+///
+/// This implementation creates one additional thread which manages the garbage collection for the class.  While it may
+/// be possible to get this class to compile in WASM architecture, the additional thread makes using this class
+/// unsuitable for WASM's current blocking workflow.
+///
+/// Keys are marian::Words, values are ProcessedRequestSentences.
```
jerinphilip, line 52: note-to-self: Update. keys are hash values of `uint64_t` now.

src/translator/processed_request_sentence.h
kpu: Regarding the marian API with Words, it's already a bergamot special API anyway.  The best practice here is to replace it with accepting an iterator range.  Or at least a pointer range.

src/translator/cache.cpp
```diff
@@ -0,0 +1,168 @@
...
+bool ThreadSafeL4Cache::fetch(const marian::Words &words, ProcessedRequestSentence &processedRequestSentence) {
+  auto &hashTable = context_[hashTableIndex_];
+
+  /// We take marian's default hash function, which is templated for hash-type. We treat marian::Word (which are indices
+  /// represented by size_t (unverified) and convert it into a uint64_t for hashing with more bits)
+  ///
+  /// With the use of a 64-bit hash, the probability of collision is as follows:
+  /// 100K sentences: 2.710478508305414e-10
+  ///   1M sentences: 2.7105026889628903e-8
+  ///   1B sentences: 0.026741008873954697
```
XapaJIaMnu, line 36: Those probabilities are wrong/misleading since our cache is not infinite and gets reset often (when the user closes the application).

src/tests/apps.cpp
kpu: So reading the entire file to stdin.  I didn't even know one could do that with a buffer.

kpu:
Open an issue about the range modification to Marian Words API.  But the slower copying path can stick in from the time being.  

Fix the misleading probabilities.

src/translator/cache.cpp
```diff
@@ -0,0 +1,168 @@
...
+                          std::chrono::milliseconds(options->get<size_t>("cache-ebr-interval")),
+                          options->get<size_t>("cache-ebr-num-queues")),
+      cacheConfig_(options->get<size_t>("cache-size") * 1024 * 1024,
+                   std::chrono::seconds(options->get<size_t>("cache-time-to-live")),
+                   options->get<bool>("cache-remove-expired")),
+      service_(epochManagerConfig_),
+      context_(service_.GetContext()) {
+  // There is only a single cache we use. However, L4 construction API given by example demands we give it a string
+  // identifier.
+  const std::string cacheIdentifier = "global-cache";
```
kpu, line 21: Does this mean there can only be one instance of this object?

src/translator/cache.h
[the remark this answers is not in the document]
jerinphilip: Replaced with uint64_t

src/translator/cache.cpp
jerinphilip: Ended up removing.

src/translator/processed_request_sentence.h
jerinphilip: https://github.com/browsermt/marian-dev/issues/54

src/translator/cache.cpp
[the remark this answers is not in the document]
jerinphilip: Cache seems to hit max capacity (configured to be 20MB) at 4000 sentences? This is L4, the other cache is in use in a single-threaded operation, if you need the metrics it'll take running overnight. Will try to polish and add soon though. 

```
[{
"lines" : 1000,
 "hits" : 0,
 "misses" : 1000,
 "evictedRecords": 0,
"activeRecords": 1000,
"totalSize": 7051816
}
,
{
"lines" : 2000,
 "hits" : 2,
 "misses" : 1998,
 "evictedRecords": 0,
"activeRecords": 1996,
"totalSize": 11481684
}
,
{
"lines" : 3000,
 "hits" : 3,
 "misses" : 2997,
 "evictedRecords": 0,
"activeRecords": 2995,
"totalSize": 16489254
}
,
{
"lines" : 4000,
 "hits" : 4,
 "misses" : 3996,
 "evictedRecords": 47,
"activeRecords": 3947,
"totalSize": 20960258
}
,
{
"lines" : 5000,
 "hits" : 7,
 "misses" : 4993,
 "evictedRecords": 1173,
"activeRecords": 3818,
"totalSize": 20893520
}
```

src/tests/apps.cpp
jerinphilip: Moving to an issue outside this PR: https://github.com/browsermt/bergamot-translator/issues/221, where this will be handled, along with other such occurences. Marking as resolved.

src/translator/cache.cpp
jerinphilip: I would expect so. The following is the example provided by L4 to work with a Cache, which is adapted for our use-case:

* https://github.com/browsermt/L4/blob/master/Examples/main.cpp

The string might be to get the same instance in a multiproc setting.

src/translator/batcher.cpp
XapaJIaMnu: Was confused about what "bucket" referred to here. It's about batching, not cache.

src/translator/cache.cpp
```diff
@@ -0,0 +1,157 @@
...
+  size_t numBuckets = options->get<size_t>("cache-buckets");
+  hashTableIndex_ = service_.AddHashTable(L4::HashTableConfig(
```
XapaJIaMnu, line 23: What is the role of the cache buckets here? is there a reason to have more than one? Are there performance implications?

src/translator/cache.cpp
```diff
@@ -0,0 +1,157 @@
...
+      cacheIdentifier, L4::HashTableConfig::Setting{static_cast<uint32_t>(numBuckets)}, cacheConfig_));
+}
+
+bool ThreadSafeL4Cache::fetch(const marian::Words &words, ProcessedRequestSentence &processedRequestSentence) {
+  auto &hashTable = context_[hashTableIndex_];
```
XapaJIaMnu, line 28: Is there a reason for using a reference here as opposed to a pointer? Just a bit of an unusual design pattern.

XapaJIaMnu: Another minor comment. Excessive use of `auto` makes readability worse.

src/translator/cache.h
```diff
@@ -0,0 +1,196 @@
...
+/// unsuitable for WASM's current blocking workflow.  Keys are marian::Words, values are ProcessedRequestSentences.
+///
+/// There is a serialization (converting to binary (data*, size)) of structs overhead with this class as well, which
+/// helps keep tight lid on memory usage. This should however be cheaper in comparison to recomputing through the graph.
+
+#ifndef WASM_COMPATIBLE_SOURCE
+class ThreadSafeL4Cache {
+ public:
+  // L4 has weird interfaces with Hungarian Notation (hence IWritableHashTable). All implementations take Key and Value
+  // defined in this interface. Both Key and Value are of format: (uint8* mdata, size_t size). L4 doesn't own these - we
```
XapaJIaMnu, line 62: A bit confusing. Just say "L4 makes a copy of the key and the value and stores them."

src/translator/cache.cpp
```diff
@@ -0,0 +1,157 @@
...
+  stats.misses = perfData.Get(L4::HashTablePerfCounter::CacheMissCount);
+  stats.activeRecords = perfData.Get(L4::HashTablePerfCounter::RecordsCount);
+  stats.evictedRecords = perfData.Get(L4::HashTablePerfCounter::EvictedRecordsCount);
+  stats.totalSize = perfData.Get(L4::HashTablePerfCounter::TotalIndexSize);
+  return stats;
+}
+
+#endif  // WASM_COMPATIBLE_SOURCE
+
+ThreadUnsafeLRUCache::ThreadUnsafeLRUCache(Ptr<Options> options)
```
XapaJIaMnu, line 92: Just thoughts, thinking about real world usecases. We probably don't particularly care if the cache is LRU, because the principal usecases are a person editing something they've input, or, alternatively, loading additional page content. Good to benchmark the cost though.

src/translator/cache.cpp
XapaJIaMnu: I assume all threads go to the same bucket, otherwise you would have a lot of cache misses.

src/translator/cache.h
```diff
@@ -0,0 +1,196 @@
...
+  /// L4 Service. There's a string key based registration through service_ and accesses from multiple processes to get
+  /// the same map.
+  L4::LocalMemory::HashTableService service_;
+
+  /// An L4 Context, does magic with epochs everytime accessed;  "Once a context is retrieved, the operations such as
+  /// operator[] on the context and Get() are lock-free."
+  L4::LocalMemory::Context context_;
+
+  /// context_[hashTableIndex_] gives the hashmap for Get(...) or Add(...) operations
+  size_t hashTableIndex_;
```
XapaJIaMnu, line 116: Do we have a use case where hashTableIndex != 1

src/translator/processed_request_sentence.h
```diff
@@ -0,0 +1,280 @@
...
+
+ public:
+  /// To be used in conjunction with delayedAllocate. A ProcessedRequestSentence is created uninitialized. Upon incoming
+  /// History, we compute the requireSize and allocate later with delayed allocate.
+  Storage() : data_{nullptr}, size_(0) {}
+
+  void delayedAllocate(size_t requiredSize) {
+    assert(!initialized());
+    size_ = requiredSize;
+    data_ = malloc(sizeof(char) * size_);
```
XapaJIaMnu, line 95: Any reason why C++ `New` is not used here?

src/translator/request.cpp
```diff
@@ -11,38 +11,66 @@ namespace marian {
...
+      // 2. Also, if cache somehow manages to decrease all counter prefilling histories, then we'd have to trigger
+      // ResponseBuilder as well. No segments go into batching and therefore no processHistory triggers.
+      if (counter_.load() == 0) {
+        responseBuilder_(std::move(processedRequestSentences_));
+      }
+    }
   }
 }
 
+bool Request::isCachePrefilled(size_t index) { return !(processedRequestSentences_[index].empty()); }
```
XapaJIaMnu, line 49: What is prefilled cache? Are we supporting persistent cache?

XapaJIaMnu (changes requested):
Getting better. I have a few questions. Also, could you update the performance numbers so that we can see what is the impact of the last changes that you did?

src/translator/cache.h
jerinphilip: > A bit confusing. Just say "L4 makes a copy of the key and the value and stores them."

I'm confused. As an explanation for what? 

This comment is in response to your query regarding the aliasing of types, for:

```cpp
using KeyBytes = L4::<incomprehensible-thing>
using ValueBytes = L4::<incomprehensible-thing>
```

src/translator/cache.cpp
jerinphilip: > Is there a reason for using a reference here as opposed to a pointer? 

Reason: L4 Example.

* https://github.com/microsoft/L4/blob/master/Examples/main.cpp#L9-L24

This is a remnant of copying over from the example and getting it to work. I can try to find the actual type and fix it if you insist, but I'd prefer to keep closer to the example and point there. Maybe adding a comment linking would help as an alternative?

src/translator/cache.h
jerinphilip: I hope they used 0-based indexing. `hashTableIndex_ != 0` in this case? To answer your query, we probably don't. There's only 1 Cache we use which we have an adapter wrapping around. L4 however in the example provides only doc on how to operate it with the service, string and indices.

src/translator/processed_request_sentence.h
jerinphilip: This is untyped memory. Does [this answer](https://stackoverflow.com/a/184568/4565794) look like something I should be doing?

XapaJIaMnu: You can do that, I guess. I see no reason `data_` to be a `void *`, as it is used as a byte array (`char` or `unsigned char`) and then reinterpreted to the appropriate type. 

Also `malloc` can fail silently and you're not checking for that, whereas `new` will throw if it fails.

src/translator/cache.h
XapaJIaMnu: Yes, I meant 0.  We should hardcode that to 0 and remove the unnecessary array lookups (for the clarity of the person reading the code)

src/translator/request.cpp
XapaJIaMnu: OOokay, I would have expected to see the word "hit".

src/translator/cache.cpp
XapaJIaMnu: I'll fire up my editor over the weekend and propose some changes.

src/translator/cache.cpp
jerinphilip: I'll have to look closer. My guess would be that buckets contain granularity of locking coarse/fine-grained to reduce contention. I've seen mutexes per bucket.. so.

jerinphilip:
> Also, could you update the performance numbers so that we can see what is the impact of the last changes that you did?


Edit workflow:

| Cache     |  Hits  | Misses | WallTime (Before) | WallTime (After) |
| --------- | ------ | ------ | -----------------:| ----------------:|
| without   | -      |   -    |          1213.36s |      1296.27719s | 
| with (L4) | 1765175|  4682  |           192.18s |       152.97415s | 


Other metrics (contention) on WNGT20:


40 threads:

|    | 40-threads.cache | 40-threads.no-cache |
|----| ----------------:| -------------------:|
|real|      4m34.87s    |            4m34.73s |
|user|    161m51.31s    |          162m37.45s |
|sys |      0m46.02s    |            0m49.42s |


1-thread:

|    |1-thread.cache| 1-thread.no-cache |
|----|-------------:| -----------------:|
|real|     6m21.37s |          6m13.78s |
|user|     6m19.92s |          6m13.17s |
|sys |      0m1.11s |           0m0.80s |

XapaJIaMnu:
Great! Much lower overhead than before.

src/translator/cache.cpp
```diff
@@ -0,0 +1,157 @@
...
+  /// represented by size_t (unverified) and convert it into a uint64_t for hashing with more bits)
+  ///
+  /// Note that there can still be collisions, in which case the translations returned from cache can be wrong. It's
+  /// practically very unlikely (claims @XapaJIaMnu). To guarantee correctness, comparisons need to be made with the
+  /// actual key, which is cannot be done with this approach, as the cache is unaware of the original marian::Words key.
+  //
+  /// TODO(@jerinphilip): Empirical evaluation that this is okay.
+
+  KeyBytes keyBytes;
+  auto key = hashFn_(words);
```
XapaJIaMnu, line 37: uint64_t pls

src/translator/cache.cpp
XapaJIaMnu: We definitely want this to be a singleton object.

src/translator/cache.cpp
XapaJIaMnu: After reading the code more carefully, buckets are a cache hyperparameter that needs to be benchmarked. More buckets might mean better performance in multithreaded case.

src/translator/cache.cpp
XapaJIaMnu: auto here resolves to `IWritableHashTable` and gives a reference, so I guess you don't have much choice regarding the reference pattern here.

src/translator/cache.h
```diff
@@ -0,0 +1,207 @@
...
+      HashWordType hashWord = static_cast<HashWordType>(word.toWordIndex());
+      util::hash_combine<HashWordType>(seed, hashWord);
+    }
+    return seed;
+  }
+};
+
+}  // namespace cache_util
+
+struct CacheConfig {
```
XapaJIaMnu, line 36: Could i have a small sweep of the bucket hyperparameter? Could you try a few values? I suspect lower bucket number might mean lower memory overhead, higher bucket number will mean less locking

XapaJIaMnu (changes requested):
I am mostly happy with this. The major comment that I have is that the cache object should be a singleton. We should have only one cache per service and all threads should go to it. There is no case where we want to have more than one cache. Hardcoding that will tell future programmers that there should be just one of these.

Minor: `auto key` -> `uint64_t key`.

I think you have coded tests already, so a small parameter sweep on bucket and I'm ready to accept this.

.gitmodules
```diff
@@ -7,3 +7,6 @@
 [submodule "bergamot-translator-tests"]
 	path = bergamot-translator-tests
 	url = https://github.com/browsermt/bergamot-translator-tests
+[submodule "3rd_party/L4"]
+	path = 3rd_party/L4
+	url = https://github.com/microsoft/L4
```
XapaJIaMnu, line 12: I'm very confused here, where is your fork located?

jerinphilip: https://github.com/browsermt/L4/pull/1. This probably happened because I started with a Microsoft submodule, opened a branch but eventually pushed to browsermt/L4. GitHub is able to lookup by commit across forks, so no fail.

jerinphilip: (I will fix this).

src/translator/cache.cpp
jerinphilip: All functions are reliant on std::hash, which returns `std::size_t` [per cppreference](https://en.cppreference.com/w/cpp/utility/hash). I have updated things to size_t instead.

kpu (changes requested):
This appears to be attempting to cache translation results across a model being unloaded and loaded again.  I think that's too much optimization at the cost of making potentially very confusing errors.  

Honestly, `options_->asYamlString()` is probably slower than hashing the entire model memory using something like MurmurHash for the size of models we're working with.  And it doesn't control for other things that we might be messing with, like lexical shortlist or vocab changing.  

An alternative option is `TranslationModel` gets a sequence number from an atomic integer on construction. Then we just hope that, by the time this integer rolls over, everything has been evicted from cache.

jerinphilip:
> Honestly, options_->asYamlString() is probably slower than hashing the entire model memory using something like MurmurHash for the size of models we're working with.

If translateLocally can retire the following string manipulation and construction of `Options`, I'd prefer deprecating the Options based constructor making it private. This will make `options_->asYamlString` go away. 

https://github.com/XapaJIaMnu/translateLocally/blob/master/src/MarianInterface.cpp#L14-L31

There are string manipulations anyway, why not supply these as YAML in an `std::string`? 

In which case, the `computeUniqueId` can move to the string-based constructor and be faster. Keep in mind that this is a one-off cost at the construction of translation model. `Id` is stored after for model's lifetime. Note that `computeUniqueId` isn't dependent on the conversion.

> And it doesn't control for other things that we might be messing with, like lexical shortlist or vocab changing.

The rest of the code relies only on tail bytes of model AlignedMemory (vocab swaps can happen, model with different vocab is unlikely, correct) and configString. I don't know about shortlist, but grabbing tail from that as well isn't hard? Translations for a `(model, source-sentence)` cached are valid even if shortlist changes I suppose, as functional translations? 

QE operates outside Cache Layer. 

> https://github.com/browsermt/bergamot-translator/issues/104#issue-861311748: I'm beginning to think we should have a unified binary file like @XapaJIaMnu was suggesting.

Also, Why not have the `Id` in the binary when we get to `MemoryBundle` to bin instead of individual pieces? 


The integer Id and hoping for alternative is even more unattractive. It's code that sticks out now that we have to start assigning Ids to translation models, while it's entirely possible to have an internal unique-Id as f(modelFile/Memory).

XapaJIaMnu:
For translatelocally, we will want to be able to manipulate the flags with which marian/bergamot-translator is loaded. If you provide a saner way to create the configuration, that's fine.

I think kenneth's suggestion to have each model receive an unique id via an atomic integer is the best solution. Add that atomic id to hash combine and we're good to go.

jerinphilip:
On the `modelId`, we're working with the running `Id` generation solution for now. This will guarantee correctness. @kpu has pointed out I might just be hashing tail padding instead of actual model content and could lead to all sorts of weird errors. :-|

I have got WASM to acknowledge L4 exists and get it to compile to fix build errors. Note that this won't work at runtime, because of the background thread. I just didn't want to create on more ifdef and go into ifdef hell all the way from cache to service to tests. 

The tests now require that a separate cache, service be tested. I've mostly been ignoring `BlockingService` in favour of `AsyncService` (only active test in former was that faithful translation works). For now, rather than doing one-off I have templated the test-suite to work with both services, bridging at a constraint that single-source blob of text in single Response out (this involves . This will compile-time guarantee that all test-code is compatible to work with both Services given the bridge exists. However, I'm not going into the laborious task of maintaining tests on both paths (test at runtime for outputs). Differences in BLAS in `WASM_COMPATIBLE_SOURCE` makes it painful enough. 

For now, I'm adding one cache-test copy (from native) on `WASM_COMPATIBLE_SOURCE` path to demonstrate the same cache mechanism works there as well for purposes of this PR. 

That said, all builds are succeeding now unlike before with some code-moving around. Some benchmarking is left to do.

src/translator/cache.cpp
XapaJIaMnu: Wrong no need for singleton.

src/translator/cache.cpp
XapaJIaMnu: Buckets don't really matter all that much in testing

src/translator/cache.h
jerinphilip: Hmmm, do we want to do this? We're taking up the responsibility of manually verifying things here - why not simply keep L4 as the single-source of truth?

src/translator/cache.cpp
```diff
@@ -0,0 +1,169 @@
+#include "cache.h"
+
+#include <cstdlib>
+
+#include "translation_model.h"
+
+namespace marian {
+namespace bergamot {
+
+size_t HashCacheKey::operator()(const CacheKey &key) const {
```
XapaJIaMnu, line 10: Does this work with -m32 option? `size_t` is `uint32_t` there, whereas we expect `uint64_t` keys.
I added `m32` build option to marian: https://github.com/browsermt/marian-dev/blob/master/CMakeLists.txt#L40

src/translator/cache.h
XapaJIaMnu: All i am saying is `const size_t constexpr hashTableIndex_ = 1;`. That way when you open any IDE you would hover it and find "oh, we just have one hash table, ok".

src/translator/cache.h
```diff
@@ -0,0 +1,205 @@
...
+namespace bergamot {
+
+class TranslationModel;
+
+struct CacheKey {
+  const TranslationModel *model;
+  const marian::Words &words;
+};
+
+struct HashCacheKey {
```
XapaJIaMnu, line 24: Why is this a struct, when it clearly has no state? Every time you call it with a the full `CacheKey`. Some comments on the above function as well.

src/translator/cache.h
```diff
@@ -0,0 +1,205 @@
...
+  size_t operator()(const CacheKey &key) const;
+};
+
+struct CacheConfig {
+  size_t sizeInMB{20};
```
XapaJIaMnu, line 29: Those can't change once the cache is initialised, right? In that case they should be `const`'d. Also, they should be initialised in the constructor. The ones which are not modifiable by the constructor should get `const constexpr` treatment.

src/translator/cache.h
```diff
@@ -0,0 +1,205 @@
...
+  RecordList storage_;
+  std::unordered_map<Record::Key, RecordPtr> cache_;
+
+  /// Active sizes (in bytes) stored in storage_
+  size_t storageSize_;
+
+  // Limit of size (in bytes) of storage_
+  size_t storageSizeLimit_;
+
+  HashCacheKey hashFn_;
```
XapaJIaMnu, line 197: Again this is strange, as it holds no state.

XapaJIaMnu (changes requested):
Might have missed something as this is 1.5k lines of code. Looks better, take a look at the comments.

src/translator/cache.cpp
jerinphilip: Everything's made `size_t`. 

[`hash_combine`](https://github.com/marian-nmt/marian-dev/blob/9fa166be885b025711f27b35453e0f2c00c9933e/src/common/hash.h) mixes up `size_t` seeds and such unless explicitly specified. Inconsistency would lead to issues when going between WASM and Native. 

The rationale applied is to follow `size_t` everywhere (even cpp `std::hash` returns `size_t`). For WASM this will consistently be 32-bit while for Native it'll be 64-bit, which should avoid any narrowing errors and inconsistencies.

src/translator/response_builder.cpp
```diff
@@ -1,27 +1,22 @@
...
-    // Alignments
-    // TODO(jerinphilip): The following double conversion might not be
-    // necessary. Hard alignment can directly be exported, but this would
-    // mean WASM bindings for a structure deep within marian source.
-    auto hyp = std::get<1>(result);
-    auto softAlignment = hyp->tracebackAlignment();
+void ResponseBuilder::buildAlignments(const ProcessedRequestSentences &processedRequestSentences, Response &response) {
+  // There are no means to test this works due to tests being removed. TagTree will have to take over after cache,
+  // tentatively.
+  /*
```
jerinphilip, line 17: Backref. Do we want this kept for translateLocally?

jerinphilip: @qianqianzhu Alignments are still available here for you to pick-up. The section is commented. I'd expect your tag-tree to show up somewhere here and displace "alignments", assuming the transformed tree through inside-out is superior.

`convertSoftAlignToHardAlign` is not range based and demands a vector, so it won't work. Until then (softAlignments) should be intact. 

What is available to you is a pseudo-vector than a real C++ `std::vector`. It's backed by read-only binary storage. You have `cbegin()`, `cend()` or `size()` and index based for loops to access.

src/translator/cache.h
jerinphilip: const is treated at the whole object passing (members where these are kept as a whole are consts, incoming args are consts as a whole).

src/translator/cache.h
jerinphilip: This is now a static member function at the interface, with CacheKey and hash(CacheKey) being protected members. Anyone outside is thus forbidden from using these, and the static member function should eliminate any problems with "state".

src/translator/cache.h
jerinphilip: > `const size_t constexpr hashTableIndex_ = 1;` 

You mean zero again, I take. That this human error happens again is reason enough for me not to change this. xD

XapaJIaMnu: Yes, because passing that parameter from somewhere else is so much less error prone. Logic.

src/translator/cache.cpp
```diff
@@ -0,0 +1,165 @@
...
+      cacheConfig_(config.sizeInMB * 1024 * 1024, std::chrono::seconds(config.timeToLiveInMilliseconds),
+                   config.removeExpired),
+      service_(epochManagerConfig_),
+      context_(service_.GetContext()) {
+  // There is only a single cache we use. However, L4 construction API given by example demands we give it a string
+  // identifier.
+  const std::string cacheIdentifier = "global-cache";
+  hashTableIndex_ = service_.AddHashTable(L4::HashTableConfig(
+      cacheIdentifier, L4::HashTableConfig::Setting{static_cast<uint32_t>(config.numBuckets)}, cacheConfig_));
+}
```
jerinphilip, line 33: Umm.. on the `hashTableIndex_` thread.
```
hashTableIndex_ = service.AddHashTable(...) // On construction
```

The above variable is used everywhere else. I think it maybe possible for me to `const` this through some construction ordering. `cacheIdentifier` can be constexpr. 

`hashTableIndex` *will* come from L4, obtained value used ahead. i.e, I'm not doing `constexpr` on `hashTableIndex_`being 0`. 

Does this work for you? @XapaJIaMnu

src/translator/processed_request_sentence.h
```diff
@@ -0,0 +1,280 @@
...
+  size_t size() const { return size_; }
+
+ private:
+  char *data_;
+  size_t size_;
+};
+
+class StorageIO {
+ public:
+  StorageIO(Storage &storage) : storage_(storage) { ptr_ = storage_.data_; }
```
kpu, line 143: explicit

src/translator/cache.cpp
```diff
@@ -0,0 +1,163 @@
...
+  return seed;
+}
+
+ThreadSafeL4Cache::ThreadSafeL4Cache(const ThreadSafeL4Cache::Config &config)
+    : epochManagerConfig_(config.ebrQueueSize, std::chrono::milliseconds(config.ebrIntervalInMilliseconds),
+                          config.ebrNumQueues),
+      cacheConfig_(config.sizeInMB * 1024 * 1024, std::chrono::seconds(config.timeToLiveInMilliseconds),
+                   config.removeExpired),
+      service_(epochManagerConfig_),
+      context_(service_.GetContext()),
```
kpu, line 27: I don't understand the lifecycle management of the L4 values you're reading enough.  Currently you have one read `context_` that is used for all read operations across all threads.  

Supposedly `context_` keeps things alive:

https://github.com/Microsoft/L4/wiki/Epoch-Queue

Does that mean reads are just accumulating?  What is decrementing the reference count?

jerinphilip: > Does that mean reads are just accumulating? What is decrementing the reference count?

This is correct. They are accumulating. Fixes have been pushed, however, L4 is not lock-free on read-anymore in this case.

jerinphilip: @XapaJIaMnu 
1. Buckets matter now. [test_cache_hparam.sh](https://github.com/browsermt/bergamot-translator-tests/blob/cache-rt/cache-tests/test_cache_hparam.sh) http://ix.io/3AUP 
2. The multi/single (40/1) cache (on/off) 1M/100K experiment [test_cache_overhead.sh](https://github.com/browsermt/bergamot-translator-tests/blob/cache-rt/cache-tests/test_cache_overhead.sh) http://ix.io/3AUQ

XapaJIaMnu: I'm confused, were the previous experiments 1thread?

jerinphilip: > were the previous experiments 1thread?

Writes weren't flushed until destruction of `Service` -> `Cache` -> `L4:`, so not a proper contention setting. L4 doesn't allow lock-free reads anymore, limitation of API @ `context_`. 

L4 allows a second reader to accept context and read "multiple" times without locks as demonstrated with multiple key-value pairs in https://github.com/browsermt/L4/blob/master/Examples/main.cpp, while something else writes I suppose. For other purposes, our code is no-longer "lock-free" (be it read or write).

XapaJIaMnu: I have missed this conversation. Why is the code no-longer lock-free on read? Was what was done before wrong?

kpu:
Here's some code for a cache.  

1. It's actually lock free
2. Manages a `shared_ptr` without any of the serialization stuff so you can keep the types as is
3. Downgrades gracefully to bad platforms like webassembly
4. Doesn't have overengineered options configuration
5. No new submodule

It probably has a worse eviction policy.  

```
#include <memory>
#include <vector>

template <class Entry, class Hash = std::hash<Entry>, class Equals = std::equal_to<Entry> > class SimpleCache {
  public:
    explicit SimpleCache(std::size_t size) : entries_(size) {}

    template <class Key> std::shared_ptr<Entry> Find(const Key &key) const {
      std::size_t offset = hash_(key) % entries_.size();
      std::shared_ptr<Entry> ret = 
#ifdef WASM
         entries_[offset];
#else
        std::atomic_load(&entries_[offset]);
#endif

      if (equals_(key, *ret)) {
        return ret;
      } else {
        return std::shared_ptr<Entry>();
      }
    }

    void Store(std::shared_ptr<Entry> entry) {
      std::size_t offset = hash_(*entry) % entries_.size();
#ifdef WASM
      entries_[offset] = entry;
#else
      atomic_store(&entries_[offset], entry);
#endif
    }

  private:
    std::vector<std::shared_ptr<Entry> > entries_;

    Hash hash_;
    Equals equals_;
};
```

Issue #57, opened 2021-03-18
ByteRange capability WASM Bindings


jerinphilip:
TranslationResult as it stands now only has sentence-mappings which are string_views. However string_view is not supported by embind per https://github.com/mozilla/bergamot-translator/issues/8, and has since been changed to the following structure. 

Proposed alternative is `ByteRange`:
```
ByteRange {
    size_t begin_byte_offset;
    size_t end_byte_offset;
}
```

There are no means yet to operate with this in 1) `QualityScore` (uses string_view), 2) `TranslationResult` (uses sentence-mappings for string_views, missing word byteranges). Alignments are ready with the [above specified datatype](https://github.com/browsermt/bergamot-translator/blob/607dbc6f85d43f8ced5c8572175e5b3fb0156cfa/src/translator/sentence_ranges.h#L14-L18).

I propose we export [Annotation](https://github.com/browsermt/bergamot-translator/blob/607dbc6f85d43f8ced5c8572175e5b3fb0156cfa/src/translator/sentence_ranges.h#L23-L59) in WASM instead to reduce work. Annotation is a simple view on top of the reference string composed of the ByteRange primitive and can provide access to both sentences and words using a flat container. These can one to one map with indices in an Alignment Structure and reduce overall export. An alternate annotation (view on string) can simply be exported for a QualityScore reducing work.

abhi-agg:
> There are no means yet to operate with this in 1) QualityScore (uses string_view), 2) TranslationResult (uses sentence-mappings for string_views, missing word byteranges)


@jerinphilip I am replacing all the references of `std::string_view` in unified APIs with 

```
ByteRange {
    size_t beginByteOffset_;
    size_t endByteOffset_;
}
```

(following the [style guideline](https://github.com/browsermt/bergamot-translator/issues/54#issuecomment-809402068))  where `ByteRange` follows half open interval [beginByteOffset_, endByteOffset_).

If you have any concerns then please raise them otherwise I will submit a PR soon.

jerinphilip:
@abhi-agg You're writing the conversion code as well (`Response` -> `TranslationResult`) so feel free to do whatever suits your use-case. Don't say I didn't warn you your current edits won't suffice. I'm absolving myself from the trouble of editing to incomplete structs (and unified API altogether, seeing where it's headed). 🤦‍♂️

[...]

PR #203, opened 2021-07-06
Ability to extract individual sentences in JS

abhi-agg:
This PR fixes https://github.com/browsermt/bergamot-translator/issues/57

Details:
- Moved `ByteRange` structure to `definitions.h` file as this is a base structure that is used to represent a sentence, word, token etc.
- Added public methods to return sentence byte ranges from Response class
    - Doesn't make sense to expose an internal structure `AnnotatedText` just for returning this information to JS
- JS bindings to access sentence byte ranges from wasm module
- wasm test page changes to extract individual sentences from response using these newly developed JS bindings

### DIFF
--- a/src/translator/response.h
+++ b/src/translator/response.h
@@ -65,6 +65,16 @@ struct Response {
   /// to (sub-)words accessible through Annotation.
   std::vector<Alignment> alignments;
 
+  /// Returns the source sentence (in terms of byte range) corresponding to sentenceIdx.
+  ///
+  /// @param [in] sentenceIdx: The index representing the sentence where 0 <= sentenceIdx < Response::size()
+  ByteRange getSourceSentenceAsByteRange(size_t sentenceIdx) const { return source.sentenceAsByteRange(sentenceIdx); }
+
+  /// Returns the translated sentence (in terms of byte range) corresponding to sentenceIdx.
+  ///
+  /// @param [in] sentenceIdx: The index representing the sentence where 0 <= sentenceIdx < Response::size()
+  ByteRange getTargetSentenceAsByteRange(size_t sentenceIdx) const { return target.sentenceAsByteRange(sentenceIdx); }
+
   const std::string &getOriginalText() const { return source.text; }
 
   const std::string &getTranslatedText() const { return target.text; }
--- a/wasm/test_page/bergamot.html
+++ b/wasm/test_page/bergamot.html
@@ -81,6 +81,8 @@ En consecuencia, durante el año 2011 se introdujeron 180 proyectos de ley que r
   }
 
   var translationService, responseOptions, input = undefined;
+  const encoder = new TextEncoder(); // string to utf-8 converter
+  const decoder = new TextDecoder(); // utf-8 to string converter
   const constructTranslationService = async (from, to) => {
 
     const languagePair = `${from}${to}`;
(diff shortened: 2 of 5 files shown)


wasm/test_page/bergamot.html
```diff
@@ -125,6 +127,7 @@
 quiet: true
 quiet-translation: true
 gemm-precision: int8shift
+alignment: true
```
jerinphilip, line 130: This option is of no relevance to this PR, at least as of now. Sentence alignments are available irrespective of this option. Actually, legal values are "hard", "soft"  or empty where there is no alignments I think. This is a marian-option, so follow docs at https://marian-nmt.github.io/docs/cmd/marian-decoder/

```
--alignment TEXT                      Return word alignment. Possible values: 0.0-1.0, hard, soft
```

The option indicates marian-backend should compute alignments during model construction. However, the Response will have (word) alignments only if explicitly indicated through `ResponseOptions`.

wasm/test_page/bergamot.html
```diff
@@ -188,18 +191,56 @@
...
+    return sentences;
+  }
+
+  // This function returns a substring of text (a string). The substring is represented by
+  // byteRange (begin and end endices) within the utf-8 encoded version of the text.
+  const _getSentenceFromByteRange = (text, byteRange) => {
+    const utf8BytesView = encoder.encode(text);
+    const utf8SentenceBytes = utf8BytesView.subarray(byteRange.begin, byteRange.end);
+    return decoder.decode(utf8SentenceBytes);
+  }
```
jerinphilip, line 242: If encode, cut, decode is how I'm going to use ByteRanges everytime in WASM/JS I'd be unhappy about it. Is there no better way to give something to JS with which `.slice(range.begin, range.end)` can happily work?

src/translator/response.h
```diff
@@ -65,6 +65,16 @@ struct Response {
...
+  ByteRange getSourceSentenceAsByteRange(size_t sentenceIdx) const { return source.sentenceAsByteRange(sentenceIdx); }
+
+  /// Returns the translated sentence (in terms of byte range) corresponding to sentenceIdx.
+  ///
+  /// @param [in] sentenceIdx: The index representing the sentence where 0 <= sentenceIdx < Response::size()
+  ByteRange getTargetSentenceAsByteRange(size_t sentenceIdx) const { return target.sentenceAsByteRange(sentenceIdx); }
+
   const std::string &getOriginalText() const { return source.text; }
 
   const std::string &getTranslatedText() const { return target.text; }
```
jerinphilip, line 80: Why do we have the `getX` family of functions, what value does this add? I'm confused.

jerinphilip (changes requested):
Prefer the bigger ByteRange PR coming in, with holistic take and examples on how to use this for cases like a tag-tree construction (how do I create a ByteRange from JS and give it to C++, with the current API?), maybe source/target mapping indication at word level etc. The current encode-cut-decode is very cumbersome to work with thinking from a JS dev perspective (I'd like IndexRanges equivalent to the ByteRange which I can do `.slice(...)` with).

From #57:

> The question is really which is better:
>
> 1. Copy an array of size_t to javascript then it accesses that array.
For each element of the array, call from javascript to a C++ accessor returning a size_t.
> 2. I'm guessing 1 is better? In general, we should be benchmarking where all the time is going.

Looks like this PR going for 1 here without benchmarking? Are there numbers to support this is better?

wasm/test_page/bergamot.html
abhi-agg: You are right. I am removing alignment option from the config here.

wasm/test_page/bergamot.html
abhi-agg: For now, we will go with this approach because JS strings are UTF-16 encoded while byte ranges that we are passing are for strings in utf-8 encoding. Once tag-translation is complete, I will refactor this part.

src/translator/response.h
abhi-agg: > Doesn't make sense to expose an internal structure AnnotatedText just for returning this information to JS

The description of this PR explains the motivation behind this change 👍

abhi-agg:
> Prefer the bigger ByteRange PR coming in, with holistic take and examples on how to use this for cases like a tag-tree construction (how do I create a ByteRange from JS and give it to C++, with the current API?), maybe source/target mapping indication at word level etc. The current encode-cut-decode is very cumbersome to work with thinking from a JS dev perspective (I'd like IndexRanges equivalent to the ByteRange which I can do `.slice(...)` with).
> 
> From #57:
> 
> > The question is really which is better:
> > 
> > 1. Copy an array of size_t to javascript then it accesses that array.
> >    For each element of the array, call from javascript to a C++ accessor returning a size_t.
> > 2. I'm guessing 1 is better? In general, we should be benchmarking where all the time is going.
> 
> Looks like this PR going for 1 here without benchmarking? Are there numbers to support this is better?

As per the comments https://github.com/browsermt/bergamot-translator/issues/57#issuecomment-871142555 and https://github.com/browsermt/bergamot-translator/issues/57#issuecomment-874568459, we will do the benchmarking once we fix on the changes required in the interface for tag-translation.

andrenatal (approved):

Issue #78, opened 2021-03-30
Outbound Translation Requirements


abhi-agg:
A placeholder to gather and document the requirements of OT feature

andrenatal:
We need an official list of features required for the outbound translations feature. We have a slide deck with few UI mockups and some annotations: https://docs.google.com/presentation/d/15ah2n58GfbP8B97nUrfl9HdKSP3Vh71-W6vZNr37eOk/edit#slide=id.ge3232ed478_0_5, but we'd like to see all the requirements listed on a github repo.

PR #210, opened 2021-07-30
Multiple TranslationModels Implementation

### LINKED ISSUE #209: Multiple Translation models
We're going for solving support from the translator side for outbound translation models (#78) treating this as a multiple translation models solution (Not just forward and backward models, possibly more than 2).

Based on the discussions I've had with @kpu, the following planned design will not be backward compatible due to now additionally requiring information on which model to use for each translate calls. 

The entities that are involved in the translation of a single direction will be abstracted together and separated from the translation process (which is effectively going to be reduced to a thread pool after).

The following components are associated with the translation of a single direction

* Marian Model
* TextProcessor: [Vocabularies, SentenceSplitter]
* ShortList
* Ordered-by-Length-storage for Batch Generation (to generate homogeneous w.r.t single-direction batches)
* QualityEstimator (*under development; @abarbosa94 @mfomicheva @felipesantosk)

Consequently the above will be packed together in a single class, which @kpu prefers to call a `TranslationModel` (https://github.com/browsermt/bergamot-translator/issues/149#issuecomment-842309639).

The following could be shared among the `TranslationModel`s that are active:

* A thread pool of workers to produce batches from batch pools and translate (consume). 
* Translation-Cache (*under development, @jerinphilip)

`Service` will change to additionally accepting a reference to a `TranslationModel` instance in the `translate(...)` API to provide model information. `Service`, `BatchTranslator` will be cut and moved into `TranslationModel` and `Service`. `BatchTranslator` will in all likelihood, cease to exist with data going to `TranslationModel` and methods ~~moving into `Service`~~ becoming free functions.

Note that in this design, the API user (WASM+JS; @abhi-agg or translateLocally; @XapaJIaMnu ) is expected to own and manage `TranslationModel`s. Ownership will have to be shared while the translation process in other threads is active, so expect some `shared_ptr<TranslationModel>` floating around.

The interaction of producing batches and clearing out orphan sentences when a model is deleted among several `TranslationModel`s is a fuzzy area w.r.t ownership (at least to me). I will possibly have a clearer picture once I try out the implementation, look into the nuances. We may end up taking up managing models if clearing out orphan batch/sentences/models get unnecessarily tricky.

jerinphilip:
Fixes #209. Please note that accompanying changes in BRT (https://github.com/browsermt/bergamot-translator-tests/pull/43) will have to be reviewed in context of this PR.

- [x] Separate `TranslationModel`, Threaded vs Non-Threaded workflows operating with `TranslationModel`
- [x] WASM Bindings temporary editing to get green on CI. Let's pretend this works (it most likely does, FWIW). 
- [x] Marian's config-parsing doesn't support multiple configs to be used in non-trivial workflows (pivoting, multiple models; `marian --configs` is nested env); CLI Rework. Bergamot has it's own CLI now. Individual `TranslationModel `configs are supplied as `--model-configs <path>` for testing purposes. Library code can load from `std::string &config` or `Ptr<Options> options = TranslationModel::Config` as before.
- [x] Tests in BRT for Outbound translation, demonstrating the modifications work for #78.


src/translator/parser.cpp
```diff
@@ -0,0 +1,160 @@
...
+      "data-weighting",
+      "log",
+      "sqlite",     // except: 'temporary', handled in the processPaths function
+      "shortlist",  // except: only the first element in the sequence is a path, handled in the
+                    //  processPaths function
+      "ssplit-prefix-file",
+  };
+
+  marian::cli::makeAbsolutePaths(configCopy, pathsInSameDirAs, PATHS);
+  options.merge(configCopy, /*overwrite=*/true);
```
jerinphilip, line 146: Note: These multiple conversions are costly and affects library code, fix.

src/translator/parser.cpp
```diff
@@ -0,0 +1,160 @@
...
+    app_.parse(argc, argv);
+    handleSpecialOptions();
+  } catch (const CLI::ParseError &e) {
+    exit(app_.exit(e));
+  }
+
+  if (version_) {
+    std::cerr << buildVersion() << std::endl;
+    exit(0);
+  }
```
jerinphilip, line 52: Move to handleSpecialOptions

jerinphilip:
```
jphilip@var:~/code/bergamot-translator/bergamot-translator-tests$ /home/jphilip/code/bergamot-translator/build/bergamot-test --bergamot-mode test-forward-backward --model-config-paths models/enes/enes.student.tiny11/config.intgemm8bitalpha.yml.bergamot.yml models/esen/esen.student.tiny11/config.intgemm8bitalpha.yml.bergamot.yml --cpu-threads 4 < data/simple/bergamot/input.txt
    The Bergamot project will add and improve client-side machine translation in a web browser.

    Unlike current cloud-based options, running directly on users’ machines empowers citizens to preserve their privacy and increases the uptake of language technologies in Europe in various sectors that require confidentiality. Free software integrated with an open-source web browser, such as Mozilla Firefox, will enable bottom-up adoption by non-experts, resulting in cost savings for private and public sector users who would otherwise procure translation or operate monolingually.

    Bergamot is a consortium coordinated by the University of Edinburgh with partners Charles University in Prague, the University of Sheffield, University of Tartu, and Mozilla.



----------------
    El proyecto Bergamot agregará y mejorará la traducción de la máquina del lado del cliente en un navegador web.

    A diferencia de las actuales opciones basadas en la nube, funcionar directamente en las máquinas de los usuarios permite a los ciudadanos preservar su privacidad y aumenta la adopción de tecnologías de idiomas en Europa en diversos sectores que requieren confidencialidad. El software libre integrado con un navegador web de código abierto, como Mozilla Firefox, permitirá la adopción de abajo hacia arriba por parte de los no expertos, lo que resultará en ahorros de costos para los usuarios privados y del sector público que de otra manera adquirirían traducción u operarían monolinguidamente.

    Bergamot es un consorcio coordinado por la Universidad de Edimburgo con socios de la Universidad Charles en Praga, la Universidad de Sheffield, la Universidad de Tartu y Mozilla.



----------------
    The Bergamot project will add and improve the translation of the client-side machine into a web browser.

    Unlike current cloud-based options, working directly on users' machines allows citizens to preserve their privacy and increases the adoption of language technologies in Europe in various sectors that require confidentiality. Free software integrated with an open-source web browser, such as Mozilla Firefox, will allow the adoption from bottom up by non-experts, resulting in cost savings for private and public sector users who would otherwise acquire translation or operate monolinguily.

    Bergamot is a consortium coordinated by the University of Edinburgh with partners from Charles University in Prague, the University of Sheffield, the University of Tartu and Mozilla.
```

abhi-agg:
I left a comment pertaining to this PR in the main issue https://github.com/browsermt/bergamot-translator/issues/209#issuecomment-903847697 👍

CMakeLists.txt
```diff
@@ -87,9 +87,8 @@ endif()
 
 if(COMPILE_WASM)
   set(WORMHOLE ON CACHE BOOL "Use WASM wormhole in intgemm https://bugzilla.mozilla.org/show_bug.cgi?id=1672160")
-  list(APPEND WASM_COMPILE_FLAGS -pthread -O3 -g2 -fPIC -mssse3 -msimd128)
+  list(APPEND WASM_COMPILE_FLAGS -O3 -g2 -fPIC -mssse3 -msimd128)
```
jerinphilip: @abhi-agg This branch has pulled the latest cmake changes with (attempt) to remove `-lpthread`? [WASM CI passes without any additional `ifdef`](https://github.com/browsermt/bergamot-translator/runs/3560336970?check_suite_focus=true). `¯\_(ツ)_/¯`

src/translator/parser.cpp
```diff
@@ -0,0 +1,164 @@
...
+                                 "Maximum input tokens to be processed in a single sentence.", 128);
+
+  // The following is a complete hijack of an existing option, so no need to add explicitly.
+  // configParser.addOption<size_t>("--mini-batch-words", "Bergamot Options",
+  //                                "Maximum input tokens to be processed in a single sentence.", 1024);
+
+  configParser.addOption<std::string>("--ssplit-prefix-file", "Bergamot Options",
+                                      "File with nonbreaking prefixes for sentence splitting.");
+
+  configParser.addOption<std::string>("--ssplit-mode", "Bergamot Options", "[paragraph, sentence, wrapped_text]");
```
abhi-agg, line 113: @jerinphilip The default value for `--ssplit-mode` option has been removed. Is it intended?

jerinphilip: Not really. Will set back in a follow-up commit.

CMakeLists.txt
abhi-agg: Then we don't need to do anything here 👍

abhi-agg (approved):

abhi-agg (approved):

src/tests/apps.cpp
```diff
@@ -2,41 +2,41 @@
...
 
-  if (options->get<bool>("bytearray")) {
-    // Load legit values into bytearrays.
-    memoryBundle = getMemoryBundleFromConfig(options);
-  }
-
-  Service service(options, std::move(memoryBundle));
+namespace {
 
+std::string readFromStdin() {
```
kpu, line 8: Ouch, convoluted to have so many copies but it is in a test.

src/translator/translation_model.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+  // Old code.
+  if (options_->hasAndNotEmpty("shortlist")) {
+    int srcIdx = 0, trgIdx = 1;
+    bool shared_vcb =
+        vocabs_.sources().front() ==
+        vocabs_.target();  // vocabs_->sources().front() is invoked as we currently only support one source vocab
+    if (memory_.shortlist.size() > 0 && memory_.shortlist.begin() != nullptr) {
+      bool check = options_->get<bool>("check-bytearray", false);
+      slgen_ = New<data::BinaryShortlistGenerator>(memory_.shortlist.begin(), memory_.shortlist.size(),
```
kpu, line 46: This is purely const.  You don't need one per thread.

src/translator/service.h
```diff
@@ -1,146 +1,116 @@
...
+    size_t numWorkers;
+  };
+  /// Construct an AsyncService with configuration loaded from Options. Expects positive integer value for
+  /// `cpu-threads`. Additionally requires options which configure AggregateBatchingPool.
+  AsyncService(const AsyncService::Config &config);
+
+  /// Create a TranslationModel compatible with this instance of Service. Internally assigns how many replicas of
+  /// backend needed based on worker threads set. See TranslationModel for documentation on other params.
+  template <class ConfigType>
+  Ptr<TranslationModel> createCompatibleModel(const ConfigType &config, MemoryBundle &&memory = MemoryBundle{}) {
```
kpu, line 77: Since your `TranslationModel` already has elements of service embedded (the batching pool), just make `TranslationModel` take the `AsyncService` or `BlockingService` in the constructor.

src/translator/translation_model.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+  using CorpusBatch = marian::data::CorpusBatch;
+  Ptr<CorpusBatch> corpusBatch = New<CorpusBatch>(subBatches);
+  corpusBatch->setSentenceIds(sentenceIds);
+  return corpusBatch;
+}
+
+void TranslationModel::translateBatch(size_t deviceId, Batch &batch) {
+  auto &backend = backend_[deviceId];
+  auto search = New<BeamSearch>(options_, backend.scorerEnsemble, vocabs_.target());
```
kpu, line 171: Always prefer stack variables over memory allocation.

src/translator/translation_model.h
```diff
@@ -0,0 +1,122 @@
...
+/// workflow for translation.
+
+class TranslationModel {
+ public:
+  using Config = Ptr<Options>;
+
+  /// Equivalent to options based constructor, where `options` is parsed from string configuration. Configuration can be
+  /// JSON or YAML. Keys expected correspond to those of `marian-decoder`, available at
+  /// https://marian-nmt.github.io/docs/cmd/marian-decoder/
+  TranslationModel(const std::string& config, MemoryBundle&& memory, size_t replicas = 1)
```
kpu, line 49: This replicas thing is ugly.  I get why you are doing it this way since we don't have shared workspace memory yet.

src/translator/aggregate_batching_pool.cpp
```diff
@@ -0,0 +1,33 @@
...
+  // TODO(@jerinphilip): Set aggregate limits
+}
+
+size_t AggregateBatchingPool::enqueueRequest(Ptr<TranslationModel> model, Ptr<Request> request) {
+  model->enqueueRequest(request);
+  aggregateQueue_.push(model);
+  return request->numSegments();
+}
+
+size_t AggregateBatchingPool::generateBatch(Ptr<TranslationModel>& model, Batch& batch) {
```
kpu, line 17: I think these are output parameters.  So somebody could provide a non-null model here under the expectation that it will be overwritten.

src/translator/aggregate_batching_pool.cpp
```diff
@@ -0,0 +1,33 @@
...
+  while (model == nullptr && !aggregateQueue_.empty()) {
```
kpu, line 18: model should always be nullptr here.  Therefore why are you checking it?

src/translator/aggregate_batching_pool.cpp
```diff
@@ -0,0 +1,33 @@
...
+namespace marian {
+namespace bergamot {
+
+AggregateBatchingPool::AggregateBatchingPool() {
+  // TODO(@jerinphilip): Set aggregate limits
+}
+
+size_t AggregateBatchingPool::enqueueRequest(Ptr<TranslationModel> model, Ptr<Request> request) {
+  model->enqueueRequest(request);
+  aggregateQueue_.push(model);
```
kpu, line 13: If I understand correctly, a model can be enqueued multiple times.  So if there's a continuous stream of requests coming in, the first element of `aggregateQueue_` will never be popped and `agreegateQueue_` will grow unbounded.

src/translator/aggregate_batching_pool.h
```diff
@@ -0,0 +1,54 @@
...
+#include <memory>
+#include <queue>
+
+#include "data/types.h"
+#include "translation_model.h"
+
+namespace marian {
+namespace bergamot {
+
+/// Aggregates request queueing and generation of batches from multiple TranslationModels (BatchingPools within,
```
kpu, line 13: Comments should say this class is not threadsafe.

src/translator/threadsafe_batcher.cpp
```diff
@@ -1,38 +1,49 @@
...
 
 #include <cassert>
 
 namespace marian {
 namespace bergamot {
 
-ThreadsafeBatcher::ThreadsafeBatcher(Ptr<Options> options) : backend_(options), enqueued_(0), shutdown_(false) {}
+template <class BatchingPoolType>
+template <class... Args>
+GuardedBatchingPoolAccess<BatchingPoolType>::GuardedBatchingPoolAccess(Args &&... args)
```
kpu, line 13: Why was this renamed from Threadsafe + the class it wraps?

src/translator/translation_model.cpp
```diff
@@ -0,0 +1,177 @@
...
+#include "data/corpus.h"
+#include "data/text_input.h"
+#include "parser.h"
+#include "translator/beam_search.h"
+
+namespace marian {
+namespace bergamot {
+
+TranslationModel::TranslationModel(const Config &options, MemoryBundle &&memory /*=MemoryBundle{}*/,
+                                   size_t replicas /*=1*/)
```
kpu, line 15: Some TODOs about how this is a horrible hack to be replaced when workspaces are extracted from graphs.

kpu (changes requested):
Unbounded memory consumption.

src/translator/translation_model.h
```diff
@@ -0,0 +1,122 @@
...
+  void translateBatch(size_t deviceId, Batch& batch);
+
+ private:
+  Config options_;
+  MemoryBundle memory_;
+  Vocabs vocabs_;
+  TextProcessor textProcessor_;
+
+  /// Maintains sentences from multiple requests bucketed by length and sorted by priority in each bucket.
+  BatchingPool batchingPool_;
```
kpu, line 97: It's weird to have a batching pool inside the TranslationModel.  What you could have had is something like a TranslationModel object without a pool, then a larger struct that has BatchingPool and TranslationModel as member variables.  But then I get that you'd have to look up the BatchingPool.  So from a user perspective I guess it's right to have them together.  Maybe slough off a class that just has the model stuff and make it a member variable here?

src/tests/apps.cpp
jerinphilip: Carrying over to https://github.com/browsermt/bergamot-translator/issues/221, to be fixed later.

src/translator/aggregate_batching_pool.cpp
jerinphilip: This is replaced with an `unordered_set`.

src/translator/service.h
jerinphilip: Consensus in meeting if I understood correctly is to solve this via separating workspaces and binding notions of number of workers with Service. 

Alternatives discussed where overloading with both Service and packing it behind TranslationModel, but this would make pivoting rather unintuitive.

src/translator/translation_model.h
jerinphilip: In the current setting, providing the worker specifics or how-to translate, one can use API provided by `TranslationModel` to achieve translation of a request. The `BatchingPool` is a part of a given `TranslationModel` instance, aimed at generating homogeneous batches collating from multiple requests. Ownership is with `TranslationModel` as it is not possible to translate without the graph and other preproc existing. I perceive `TranslationModel` as "all things needed to translate in a single-direction", which includes the `BatchingPool`. 

I'm confused to how to separate `TranslationModel = ActualTranslationModel + BatchingPool` into meaningful units. `BatchingPool` handles batching (raw sentences in, optimally* packed batches out), which is understandable and keeping this concern separate makes sense. But regarding the other half - what is `ActualTranslationModel` intended to do? I'd need to know this to create a proper separation. It's never used by the client. It's one more layer deeper for library use (library works with `TranslationModel`).

jerinphilip: Carrying this outside this PR, if this needs to be addressed. Since it does not affect anything external, we can always come back and change this... so.

PR #217, opened 2021-08-23
BRT: Update sacrebleu to get speed tests back working

jerinphilip:


XapaJIaMnu:
Does older sacrebleu still work? Is it better to have some version check and then based on that change the sacrebleu line?

jerinphilip:
Perhaps I should have described this better. Starting version 2.0.0, sacrebleu now provides JSON output by default: structured. Since the setup scripts use `python3 -m pip install` they update to the latest version hence breaking BRTs that run from scratch. I gain old behaviour by adding `-f text` (format text) with the new sacrebleu, which is what is done for now. The current test which uses this sacrebleu is one that runs only on var (although approx BLEU comparisons might be broken as well). 

> Does older sacrebleu still work? 

With the new script? No. 

> Is it better to have some version check and then based on that change the sacrebleu line?

I believe it's not best to create if-else branches here based on versioning. It's just testing after all, we don't need to support all versions for all packages involved, just keep it up-to-date enough such that GitHub CI and Jenkins on var runs. Expecting BRTs to run with latest WMT sets (if at all), I'd guess we are best off living at the head with sacrebleu. 

A standard way to solve this is to have a fixed environment that can be reproduced with `virtualenv`, with a [`requirements.txt`](https://github.com/browsermt/bergamot-translator-tests/blob/ee534f7507966efe3199ac84e56bdd4b3950b736/requirements.txt) specifying versions compatible with (currently we don't specify versions).

XapaJIaMnu:
If else branch takes literally 5 minutes to code, including parsing the version output.... But as you decide, I don't particularly care either way.

PR #223, opened 2021-09-22
Separating workspace from TranslationModel

jerinphilip:
WIP.

### DIFF
--- a/src/translator/service.cpp
+++ b/src/translator/service.cpp
@@ -10,7 +10,30 @@
 namespace marian {
 namespace bergamot {
 
-BlockingService::BlockingService(const BlockingService::Config &config) : requestId_(0), batchingPool_() {}
+Ptr<Options> horribleOptionsHack() {
+  Ptr<Options> options = std::make_shared<Options>();
+  options->set<std::string>("gemm-precision", "int8shiftAlphaAll");
+  options->set<bool>("dump-quantmult", false);
+  options->set<float>("clip-gemm", 1.0);
+  options->set<bool>("use-legacy-batching", false);
+  return options;
+}
+
+Graph createGraph(size_t workspaceSizeInMB, size_t cpuId) {
+  marian::DeviceId device_(cpuId, DeviceType::cpu);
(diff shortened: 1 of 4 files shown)


XapaJIaMnu:
A small test app? I want to see how exactly it gets used.

jerinphilip:
Umm.. this is a draft PR so I can take something concrete to discussions.

There is a forward backward test-app (demonstrating working for outbound translation) which passes in tests. 

https://github.com/browsermt/bergamot-translator/blob/cf541c68f9b43bce8c68e2292007a1573cfaa38e/src/tests/apps.cpp#L59

However, I don't trust the current state in a multi-threaded high volume multiple models setting. So we are looking at a larger experiment where `n > 1` models are in play on `sources.shuf` - translating to `translated.tgt`. I was thinking run each models individually to get individual.out, and compare it with output in a pooled model setting.

Due to batching differences and floating-point approximations I don't expect exact matches. Therefore it will be compared with relaxations (BLEU?).

In any case, there are known issues:

1, in namespacing which will cause overwrites which will need to be fixed by altering createScorers taking in a prefix perhaps. https://github.com/browsermt/marian-dev/issues/57#issuecomment-925082770
2, Freeing scorer created tensors/params upon model deletion. I'm not sure if this happens automatically (I suspect lifetime for these are beyond scorer's lifetimes hence leaking).

XapaJIaMnu:
What's the problem with multiple threads? Each thread has its own scorer and graph?

jerinphilip:
> Each thread has its own scorer and graph?

No, scorer is a property of `TranslationModel`, tied to architecture. Only graph has moved to thread (along with it the workspace). 

> What's the problem with multiple threads?
 
There are no problems with multiple threads. translateBatch is triggered from the worker who has the graph, so thread-safety should be okay. We have duplicates of scorers on each graph (for each thread), so that should be okay as well I suppose.

The problem is graph has a variable store - (string name, Tensor(?) value). name can conflict between "multiple translation models". It won't show up right now I'm guessing but might show up in high volume? In the simple case the first one might have gone to thread-1 and the second to thread-2 never colliding and this overwrite creating funny behaviour being small inputs and done in series (not async).

PR #224, opened 2021-09-28
Wasm test page for new Service interface (Multiple Models)

abhi-agg:
This PR modifies the wasm test page to verify whether the new feature (loading and working with multiple translation models) works in the engine for wasm target.

It fixes a part of the issue https://github.com/mozilla/bergamot-translator/issues/58

### DIFF
--- a/wasm/test_page/bergamot.js
+++ b/wasm/test_page/bergamot.js
@@ -32,14 +32,20 @@ document.querySelector("#load").addEventListener("click", async() => {
 const translateCall = () => {
     const text = document.querySelector('#from').value;
     const paragraphs = text.split("\n");
-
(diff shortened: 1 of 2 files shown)


andrenatal (approved):

XapaJIaMnu (approved):
I don't really understand what is happening here, but I guess it's ok?

abhi-agg:
> I don't really understand what is happening here, but I guess it's ok?

@XapaJIaMnu If you are performing translation for En->De and now also want to translate for Es->En then both En->De and Es->En models can co-exist in the engine now. Earlier, we had to destruct the first model before we could use any other model in the engine. I just updated the wasm test page to test this feature where you can load multiple models and translate using all those models. I hope this helps 👍

PR #227, opened 2021-10-09
Cache with std::atomic on shared-pointers

jerinphilip:
Mostly copied over from https://github.com/browsermt/bergamot-translator/pull/202#issuecomment-937609245, with a few modifications.

### DIFF
--- a/src/translator/cache.h
+++ b/src/translator/cache.h
@@ -1,0 +1,41 @@
+#pragma once
+#include <memory>
+#include <vector>
+
+namespace marian::bergamot {
+
+template <class Key, class Value, class Hash = std::hash<Key>, class Equals = std::equal_to<Key>>
+class AtomicCache {
+ public:
+  struct Record {
+    Key key;
+    Value value;
+  };
+
+  using FloatingRecord = std::shared_ptr<Record>;
+
+  explicit AtomicCache(std::size_t size) : records_(size, nullptr) {}
+
+  FloatingRecord Find(const Key &key) const {
+    const FloatingRecord &record = records_[hash_(key) % records_.size()];
+    FloatingRecord ret = std::atomic_load(&record);
+    if (ret && equals_(key, ret->key)) {
+      return ret;
+    } else {
+      return FloatingRecord{nullptr};
+    }
+  }
+
+  void Store(FloatingRecord &recordIn) {
+    FloatingRecord &record = records_[hash_(recordIn->key) % records_.size()];
+    atomic_store(&record, recordIn);
+  }
+
+ private:
+  std::vector<FloatingRecord> records_;
+
+  Hash hash_;
+  Equals equals_;
+};
+
+}  // namespace marian::bergamot
--- a/src/tests/units/CMakeLists.txt
+++ b/src/tests/units/CMakeLists.txt
@@ -1,6 +1,7 @@
 # Unit tests
 set(UNIT_TESTS
     annotation_tests
+    cache_tests
     quality_estimator_tests)
 
 foreach(test ${UNIT_TESTS})
--- a/src/tests/units/cache_tests.cpp
+++ b/src/tests/units/cache_tests.cpp
@@ -1,0 +1,44 @@
+
+#include <random>
+#include <thread>
+
+#include "catch.hpp"
+#include "translator/cache.h"
+
+using namespace marian::bergamot;
+
+TEST_CASE("Test Cache in a threaded setting") {
+  size_t numThreads = 100;
+  size_t numIters = 10000;
+  using TestCache = AtomicCache<int, int>;
+
+  TestCache cache(300);
+
+  // Insert (x, x) into cache always. Tries to fetch, if hit assert the value is (x, x).
+  // Not sure if this catches any ABA problem.
+
+  auto op = [numIters, &cache]() {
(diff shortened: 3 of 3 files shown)


kpu:
I had designed it with the intention you could take an existing `shared_ptr` object and stuff it in the cache.  Has that changed?

jerinphilip:
> Is that a useful case?

Yes? Start with `Ptr<History>` don't convert it since you fiercely hate conversions? For us:

```cpp
using TranslationCache = AtomicCache<size_t, Ptr<History>>;
```

I'm still suspicious how there won't be any ABA in this cache, but doesn't seem to violate anything in the unit-tests (which are not that big, TBH). We are not editing one memory location, but a whole lot of it. Instructions (which enable lock-free atomics) operate at 1 mem-loc at a time, does it not? If it were this simple, why cant I find this everywhere online, I wonder?

kpu:
I hate unnecessary conversions.  I also hate unnecessary memory allocations especially if they are `shared_ptr` that doubles the memory allocations.  

Do I really need `shared_ptr<AtomicCache::Record<size_t, shared_ptr<History> >` ?

Do not understand your concern about ABA.  The object pointed to is kept alive the entire time it's in custody of a `shared_ptr` so nobody can create a second object there.

jerinphilip:
> Do I really need shared_ptr<AtomicCache::Record<size_t, shared_ptr<History> > ?

I ran into `std::atomic<..>` requires [`is_trivially_copyable`](https://en.cppreference.com/w/cpp/types/is_trivially_copyable) with `AtomicCache::Record<size_t, shared_ptr<History>`. [`15ee836` (#227)](https://github.com/browsermt/bergamot-translator/pull/227/commits/15ee8365bb35d229f79df4bd75b740adfe29f497)

Even without `Ptr<History>` as value (`AtomicCache<int, int>`) instead, I am getting linking failures to `atomic_load` and `atomic_store`. [`aa6320a` (#227)](https://github.com/browsermt/bergamot-translator/pull/227/commits/aa6320a5d9cf03dbb2013d579948f7056f2d567c).


I'm not sure where I'm headed, but here's some documentation.  Looks like we have conversion, and ProcessedRequestSentence / storage is staying.

https://en.cppreference.com/w/cpp/atomic/atomic#Primary_template

> The primary std::atomic template may be instantiated with any TriviallyCopyable type T satisfying both CopyConstructible and CopyAssignable. The program is ill-formed if any of following values is false:

https://en.cppreference.com/w/cpp/types/is_trivially_copyable:

> Objects of trivially-copyable types that are not potentially-overlapping subobjects are the only C++ objects that may be safely copied with std::memcpy or serialized to/from binary files with std::ofstream::write()/std::ifstream::read().

kpu:
What was the problem with the version I wrote?

kpu:
There are two ways to make a store's interface.  
1. Have a key-value struct which should be `std::pair` just for consistency with the standard library.  
2. Have a single opaque `Entry` class.  The `Hash` functor is overloaded to accept both `Entry` and `Key`.  Similarly the `Equal` functor is overloaded to accept both `Entry` and `Key` making a comparison on that.  This is what you were missing.  

But I think the issue is that the key is not a function of `History` so much as a function of the input sentence and model used.  Therefore you need a `size_t` key alongside the `History` object. In that case a key-value store makes sense.  It may as well be `std::pair<std::size_t, History>`.  The `History` object looks moveable.  You may as well move it out of `std::shared_ptr<History>` misery into the `std::pair<std::size_t, History>`.

src/translator/cache.h
```diff
@@ -0,0 +1,85 @@
...
+
+  void atomicStore(const Key &key, Value value) {
+    // No probing, direct map onto records_
+    size_t index = hash_(key) % records_.size();
+    size_t mutexId = hash_(key) % mutexBuckets_.size();
+
+    std::lock_guard<std::mutex> lock(mutexBuckets_[mutexId]);
+    Record &candidate = records_[index];
+
+    if (!used_[index]) {
```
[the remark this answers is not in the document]
jerinphilip: I have no means to tell if I hit cache or not placing reasonable asserts (at test-apps) without this. Are there alternatives I can use?

The current assert is put in a sentence, get translation. Put same sentence again. assert(cache hits the second time).

src/translator/cache.h
```diff
@@ -0,0 +1,75 @@
...
+  bool atomicLoad(const Key &key, Value &value) const {
+    // No probing, direct map onto records_
+    size_t index = hash_(key) % records_.size();
+    size_t mutexId = index % mutexBuckets_.size();
+
+    std::lock_guard<std::mutex> lock(mutexBuckets_[mutexId]);
+    const Record &candidate = records_[index];
+    if (equals_(key, candidate.first)) {
+      value = candidate.second;
+      stats_.hits += 1;
```
XapaJIaMnu, line 43: I am a bit late to this party, but shouldn't that be atomic?

jerinphilip: There's a mutex above making this section atomic..?

jerinphilip: Oh wait, you're right with regards to stats..

XapaJIaMnu: `std::atomic<int>`

Issue #212, opened 2021-08-04
First class pivot translation capability


jerinphilip:
https://github.com/mozilla/bergamot-translator/issues/62 is an ask for a "pivoting" through English to obtain non-English translations feature between languages `xx` and `yy`. 

```
[xx-source] -> [en-intermediate] -> [yy-target]
```

Just thinking out loud here, to initiate discussion. Implementation will take a while and will have to wait for #210. A relevant feature is support for parameterisation by the model in methods in `Service`, being tracked in #209 and implementation ongoing in #210 as part of supporting multiple models. Using multiple models through JS API provided by #210 alone will lead to subpar performance and incomplete outputs (the pivot feature is already 2x slower, comm back and forth will make it worse).  

Instead, we can support provision of the following additional method which sequences the underlying computations from a list of models supplied externally from JS. A loose API sketch below:

```cpp
class Service {
public:
// Applies models in series to translate source to target.
// Models[ [x_1 -> x_2] - [x_ 2-> x_3] ... [x_{n-1} -> x_{n}] ] = Model[x_1 -> x_n]
void sequencedTranslate(
	std::vector<std::shared_ptr<TranslationModel>> models, // < models to be applied in series
	std::string &&source, 
	CallbackType clientCallback,
	const ResponseOptions responseOptions) {
	// Reduce application of pivotHalfTranslate(...) on models 
	
	// Reduce needs to happen with chained-callbacks, because we don't 
	// immediately know when a request finishes.
	
	// (n-1)th Response.target is the n-th Response.source (for consistent tokenizations)
	
	// Response.alignment matrix aggregations are binary op, matrix multiplied(?)
	// over each pair of pivotHalfTranslate ops.
}

private:	

void pivotHalfTranslate(
	std::shared_ptr<TranslationModel> xx_to_yy, 
	AnnotatedText source, // need to supply consistent sentence/tokens at intermediates as source, target
	CallbackType callback,
	const ResponseOptions responseOptions,
);

}
```

`models.size()` times slower for a lower bound, additional overheads of matrix-multiplies, greater error guaranteed. On the bright side... still better than no `xx->yy` translation, I guess. 

* The sentences need to align one to one from source to intermediates to target. 
[...]

PR #231, opened 2021-10-18
Wasm test page UI for translating b/w non-English language pairs

abhi-agg:
This PR modifies the UI of the wasm test page to demonstrate translation b/w non-English language pairs. This is an improvement on the https://github.com/browsermt/bergamot-translator/pull/224.


wasm/test_page/js/worker.js
```diff
@@ -0,0 +1,294 @@
...
+  translationModels.set(languagePair, translationModel);
+}
+
+const translate = (from, to, paragraphs) => {
+  // If none of the languages is English then perform translation with
+  // English as a pivot language.
+  if (from !== 'en' && to !== 'en') {
+    log(`Translating '${from}${to}' via pivoting: '${from}en' -> 'en${to}'`);
+    let translatedParagraphsInEnglish = translateInvolvingEnglish(from, 'en', paragraphs);
+    return translateInvolvingEnglish('en', to, translatedParagraphsInEnglish);
```
jerinphilip, line 218: This works only for plaintext, is not the recommended or intended way to do pivot translation. I'm unsure who communicated this is the solution for https://github.com/mozilla/bergamot-translator/issues/62#issue-958664129. Going forward with this for HTML will incur serious debt ahead -it won't work with links. But a targeted by the library pivot translation will be compatible with other algorithms like those which require TagTree.

https://github.com/browsermt/bergamot-translator/issues/212 is unsolved and will take more development time after which the proper solution will come about, if at all.

abhi-agg: We know that for the links/tags, the current setup wouldn't work. Once the API supports pivoting for html text, this page will be updated accordingly. This is quite clear from the https://github.com/mozilla/bergamot-translator/issues/62#issuecomment-932391160

andrenatal (approved):

wasm/test_page/js/worker.js
```diff
@@ -0,0 +1,294 @@
...
+// This function extracts all the source sentences from the Response and returns them.
+const getAllSourceSentencesOfParagraph = (response) => {
+  const sentences = [];
+  const text = response.getOriginalText();
+  for (let sentenceIndex = 0; sentenceIndex < response.size(); sentenceIndex++) {
+    const utf8SentenceByteRange = response.getSourceSentence(sentenceIndex);
+    sentences.push(_getSentenceFromByteRange(text, utf8SentenceByteRange));
+  }
+  return sentences;
+}
```
jerinphilip, line 286:
```js

getAllTranslatedSentencesOfParagraph -> response.getTranslatedSentence(sentenceIndex)
getAllSourceSentencesOfParagraph -> response.getSourceSentence(sentenceIndex)

getSentences(response.target)
getSentences(response.source)

// Or just export annotation, wrap it nicely. 

```

Let's make it a point to call out how bad design is creating more bloat further down, making more work for everyone.

wasm/test_page/js/worker.js
```diff
@@ -0,0 +1,294 @@
...
+    let translatedParagraphsInEnglish = translateInvolvingEnglish(from, 'en', paragraphs);
+    return translateInvolvingEnglish('en', to, translatedParagraphsInEnglish);
+  }
+  else {
+    log(`Translating '${from}${to}'`);
+    return translateInvolvingEnglish(from, to, paragraphs);
+  }
+}
+
+const translateInvolvingEnglish = (from, to, paragraphs) => {
```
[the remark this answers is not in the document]
abhi-agg: > Your code should generalize to any pivot (or any sequence of models applied in series to get to the target language although it won't be more than 1 hop).

It doesn't make sense to me at this point to change the wasm test page to accommodate a general case which currently doesn't exist. I can generalize it once we have trained models that pivot without involving English.

wasm/test_page/js/worker.js
abhi-agg: > I'm telling you this is not the intended way to do pivot translation

Once you have the pivot translation ready the intended way, I will change the wasm test page accordingly.

andrenatal: > I'm telling you this is not the intended way to do pivot translation, demonstrating this as a reference has more downsides than upsides

@jerinphilip would you please point us to what's the `intended way to do pivot translation` for both html tags plain text then? I searched for implementation documentation but to no avail, and it would be helpful if you could provide those. 
Thanks!

jerinphilip: The intended way is you to call a new function mentioned in https://github.com/browsermt/bergamot-translator/issues/212#issue-960451083, where the issue is being tracked. You tell the C++ API, "Hey, I have these models which you can apply in series to translate.". This will give you a `Response` that is consistent, usable the same way as you would have in an English model, hence working with tags and what not, when the feature gets to you. 

This will shift the responsibility of resolving inconsistencies at alignment matrices, vocabulary mismatches, sentence misalignments to the C++ library. You will run into trouble doing this at JS at some point, and by then if this example code is used the efforts built on top of this example would go down the drain.

The feature is not implemented, hence still open. If I may ask, why are you under the impression that this has been implemented? I didn't communicate pivot translation is solved, but I believe it's my responsibility to point out pitfalls of the current approach when it's brought to my attention.

abhi-agg: > The intended way is you to call a new function mentioned in #212 (comment), where the issue is being tracked.
> The feature is not implemented, hence still open. 

@kpu Could you please confirm that until https://github.com/browsermt/bergamot-translator/issues/212 resolves, it is okay to perform translation b/w language pairs not involving English by using English as a pivot? I changed the wasm test page just to demonstrate that multiple models work in JS and used pivoting as a use case to demonstrate this. We are aware that the existing solution for pivoting works for only for the plain text and not for the html text. We would be happy to change things on our side once https://github.com/browsermt/bergamot-translator/issues/212 resolves. Please let me know if I missed something here. Thanks. 👍

kpu: We do have systems lying around for Kazakh-English with Russian as a pivot (which have not been optimized for Bergamot).  

There will be API changes in the future so we can get word alignment working in pivot translation.  It will probably be done on C++ side.  As long as you accept that your code will have to change in the future to accommodate that (and Edinburgh is unlikely to be helpful with this since it's JavaScript), I have no problem with you putting out something that works for the time being.

Issue #235, opened 2021-10-26
Marian emits no EOS 


jerinphilip:
Discovered when working with pivot translation, where the two pointer algorithm broke due to absence of an EOS in the intermediate. 

Isolated an abort where-in it's visible that sentence-piece does not decode an EOS (possibly because marian never predicted it). More documentation will appear here.

[`9e2477d` (#19)](https://github.com/jerinphilip/bergamot-translator/pull/19/commits/9e2477d53208193d2672254e3cf90d66ccff1221)

```
jphilip@var:~/code/bergamot-translator/bergamot-translator-tests$ INTGEMM_CPUID=AVX2 /mnt/Storage/jphilip/bergamot-build/bergamot-test --bergamot-mode test-pivot --model-config-paths /home/jphilip/code/bergamot-translator/bergamot-translator-tests/models/enes/enes.student.tiny11/config.intgemm8bitalpha.yml.bergamot.yml /home/jphilip/code/bergamot-translator/bergamot-translator-tests/models/esen/esen.student.tiny11/config.intgemm8bitalpha.yml.bergamot.yml --cpu-threads 4 < data/simple/bergamot/input.txt
[2021-10-26 08:21:53] Error: No EOS in targetSentenceMappings
[2021-10-26 08:21:53] Error: Aborted from void marian::bergamot::ResponseBuilder::buildTranslatedText(marian::Histories&, marian::bergamot::Response&) in /home/jphilip/code/bergamot-translator/src/translator/response_builder.cpp:44

[CALL STACK]
[0x55aeb06890f9]                                                       + 0x1130f9
[0x55aeb0684e3d]                                                       + 0x10ee3d
[0x55aeb0683f0a]                                                       + 0x10df0a
[0x55aeb0691944]                                                       + 0x11b944
[0x55aeb0668f39]                                                       + 0xf2f39
[0x55aeb06942e4]                                                       + 0x11e2e4
[0x7f2bd4f7dd80]                                                       + 0xd0d80
[0x7f2bd46df6db]                                                       + 0x76db
[0x7f2bd4408a3f]    clone                                              + 0x3f
```

The above is running on var, I have not been able to reproduce this on CI running AVX2/AVX512 machines as seen in the link above.

Issue #19, opened 2021-02-08
Jenkins bergamot-translator #18 failed


marianminion:
Build 'bergamot-translator' is failing!

Last 50 lines of build output:

```
Started by user Jerin Philip
Running as SYSTEM
Building remotely on var (vnni) in workspace /var/lib/jenkins/workspace/bergamot-translator
Cloning the remote Git repository
Cloning repository https://github.com/browsermt/bergamot-translator
 > git init /var/lib/jenkins/workspace/bergamot-translator # timeout=10
Fetching upstream changes from https://github.com/browsermt/bergamot-translator
 > git --version # timeout=10
 > git fetch --tags --progress https://github.com/browsermt/bergamot-translator +refs/heads/*:refs/remotes/origin/*
 > git config remote.origin.url https://github.com/browsermt/bergamot-translator # timeout=10
 > git config --add remote.origin.fetch +refs/heads/*:refs/remotes/origin/* # timeout=10
 > git config remote.origin.url https://github.com/browsermt/bergamot-translator # timeout=10
Fetching upstream changes from https://github.com/browsermt/bergamot-translator
 > git fetch --tags --progress https://github.com/browsermt/bergamot-translator +refs/heads/*:refs/remotes/origin/*
 > git rev-parse jp/match-marian-decoder^{commit} # timeout=10
 > git rev-parse refs/remotes/origin/jp/match-marian-decoder^{commit} # timeout=10
Checking out Revision 4e26aa32bc009c4a4d809927f224f59ae130f03e (refs/remotes/origin/jp/match-marian-decoder)
Commit message: "Duplicating bergamot-translator for benchmarks"
 > git config core.sparsecheckout # timeout=10
 > git checkout -f 4e26aa32bc009c4a4d809927f224f59ae130f03e
 > git rev-list 4e26aa32bc009c4a4d809927f224f59ae130f03e # timeout=10
[bergamot-translator] $ /bin/sh -xe /tmp/jenkins7386354468175542899.sh
+ . /etc/environment
+ PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games
+ rm -rf build
+ mkdir -p build
+ cd build
+ cat /var/lib/jenkins/cuda-10.2/version.txt
cat: /var/lib/jenkins/cuda-10.2/version.txt: No such file or directory
Build step 'Execute shell' marked build as failure

```

Changes since last successful build:
No changes

[...]

PR #236, opened 2021-10-26
First class pivot translation capability

jerinphilip:
This PR brings first-class pivot translation in the C++ API, solving #212.

### Description

Translates a text from source-language to target-language through a pivot-language. Effectively runs models in series, while having the following additional benefits compared to when `Service::translate(...)` would be used repeatedly.

1. Consistency in sentences between source and target. Consistent creation of the alignment matrix for use in downstream tasks like tag-translation (#199). 
2. Efficient sentence-splitting (does not sentence-split twice, creating inconsistencies). 
3. The `Response` generated can be used as if it were coming through `translate(...)`, eliminating any need for additional code for clients in JS or python or C++. 

`AsyncService::pivot(...)` is provisioned for C++ multi-threaded setting and `BlockingService::pivotMultiple(...)` provisioned for blocking use-case targeted at WebAssembly. 

### Implementation 

The core additions are:

1. `BlockingService::pivotMultiple(first, second, ...)`: Implemented sequentially as imperative code.
2. `AsyncService::pivot(first, second, ...)`: Implemented by callback chaining.
3. `remapAlignments(Response &first, Response &second)`: Uses two alignment matrices and token information with the intermediate being represented in the same alphabet to aggregate alignments. Aggregation is done by spreading the probability over characters and then collecting them over the differing token representations using a two pointer algorithm.

For a more friendly notation in math formulations for alignments, check [this overleaf](https://www.overleaf.com/read/bgvydnqbxmsk). 

**Known Caveat(s)**

1. There's a slack introduced which breaks some existing assumptions on the bounding memory constraints imposed by `max-length-break` by allowing the translated text to exceed `max-length-break` before being sent to the second model. This is required as marian can generate sequences of length `max-length-factor*max-length-break` and I did get bit by the extreme case in WNGT20.  However, the above compromise is good-enough for now, as the alternative will be introducing a notion of "subsentences" in annotation and doing more correction code recursively. It is on the client that the models provided can work with the growing sequence length.

2. There is an issue on where marian does not predict an EOS token. I have triggered it at [`9e2477d` (#19)](https://github.com/jerinphilip/bergamot-translator/pull/19/commits/9e2477d53208193d2672254e3cf90d66ccff1221) on var, but unable to do the same on CI. This will be looked into separately by throwing all sort of sentences and fuzzing later. For now there's a workaround of redistributing the probability mass of the target eos token over all pivot tokens equally so the sum adds up to 1, which is acceptable to the problem scenario. The issue is tracked in #235.

### Testing

For `AsyncService` for a test-case involving of en->es, es->en (same vocabulary, another one might be more coverage but is too much work).

1. Asserts the Alignment generated after pivoting is a probability distribution over source tokens given target. 
2. Outputs the sentences going from en->en, which should stay consistent over continuous development to ensure nothing breaks. 
3. An accuracy minimum of 70% of token matches from source to target calibrated on the standard bergamot input text is additionally present, ensuring that the English tokens at start and end match exactly.

`BlockingService` is tested to work manually at https://github.com/jerinphilip/bergamot-translator/pull/23 but however will remain outside and can be used by @abhi-agg (cc @andrenatal) for reference. It is tested to be working.


jerinphilip:
@abhi-agg Sorry for the delayed response. This PR would've created some extra hurdles for HTML, which seemed more important for the past few weeks and was hence on hold.

> Is it technically feasible to combine these 2 APIs into one? Something like this:
> `BlockingService::translateMultiple(std::vector<std::shared_ptr<TranslationModel>> models, ...)`

I prefer something like what you're suggesting. In fact, without the constraints brought about by the actual underlying code, I proposed something similar in #212. The concern, upon visiting the implementation details is of violating bounded memory (currently implicitly enforced by a max-length-break). Each added model in the vector adds a `max-length-factor` multiplier on the slack space required to be accounted for in `max-length-break` based bucketing. I hope you understand why this is reason to worry. The code internally will have to accomodate the same as well. There are means to achieve this, but I believe it is not in the interest for the project and developer time.

Realistically, each added model is also a major slowdown, because you're running a large (despite compressed) neural model twice. Running it multiple times will result in a slow product, and we're always looking at pivoting through English or the next reasonable pivot language (eg: Russian for certain European languages, Hindi maybe for certain Indian languages). To summarize, 'pivot(model1, model2, ...)` seems to be what's ideal given our situation.

kpu:
@jerinphilip What have the efficiency issues got to do with how the API is made, other than making it easier to use?

src/translator/translation_model.cpp
```diff
@@ -95,16 +95,27 @@ Ptr<Request> TranslationModel::makeRequest(size_t requestId, std::string &&sourc
   Segments segments;
   AnnotatedText annotatedSource;
 
-  HTML html(std::move(source), responseOptions.HTML);
+  // HTML html(std::move(source), responseOptions.HTML);
```
jerinphilip, line 98: Note to self: remove commented code.

src/translator/batching_pool.cpp
```diff
@@ -49,8 +59,18 @@ size_t BatchingPool::enqueueRequest(Ptr<Request> request) {
     if (!request->cacheHitPrefilled(i)) {
       RequestSentence sentence(i, request);
       size_t bucket_id = sentence.numTokens();
-      assert(bucket_id < bucket_.size());
+
+      // Due to a workaround for pivoting, unless we can discipline the
```
graemenail, line 63: What is the motivation for changing the behaviour for non-pivoting requests?

`makeRequest` and `makePivotRequest` exist as separate methods, but not `enqueuePivotRequest`?

Also, do we need the resize in the constructor as well as here?

graemenail (commented):
@jerinphilip, I think this looks good; I've left some minor comments, as well as some questions to try and better understand different aspects

src/translator/text_processor.cpp
```diff
@@ -138,5 +138,39 @@ void TextProcessor::wrap(Segment &segment, std::vector<string_view> &wordRanges,
...
+
+  for (size_t s = 0; s < source.numSentences(); s++) {
+    // This is our sentenceStream
+    ByteRange sentenceByteRange = source.sentenceAsByteRange(s);
+
+    // Fool tokenization using ByteRanges into looking at replacement. They're same, so okay.
+    marian::string_view sentence{&replacement.text[sentenceByteRange.begin], sentenceByteRange.size()};
+
+    std::vector<string_view> wordRanges;
+    Segment segment = tokenize(sentence, wordRanges);
```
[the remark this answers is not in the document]
jerinphilip: Yes, there was some talk in a meeting about controlling the growth of memory. Given we're working around so many things to provide pivoting - different vocabs, no way of any control over tokens - it made sense to simply relax it for pivoting.

Not that the previous control on the non-pivot path still exists by means of truncation in the text processor. It's just the generated text (pivot in-betweens) that we have no control over. The trouble arises due to constraints on alignments and mappings which is required for features like HTML translation.

It might be possible to constrain sequence lengths by cutting and merging - introducing notions of subsentences, but it's too much work and will further delay in this PR.

src/translator/service.h
```diff
@@ -47,6 +47,24 @@ class BlockingService {
   std::vector<Response> translateMultiple(std::shared_ptr<TranslationModel> translationModel,
                                           std::vector<std::string> &&source, const ResponseOptions &responseOptions);
 
+  /// With the supplied two translation models, translate using first and then the second generating a response as if it
+  /// were translated from first's source language to second's target langauge. Requires first's target to be second's
+  /// source to work correctly - effectively implementing pivoting translation via an intermediate language.
+  ///
+  /// Requires that `max-length-break` * `max-length-factor` for the second to be greater than the value for the first
```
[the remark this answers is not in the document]
jerinphilip: It's dynamically growing now and hence will not fail, I gave up temporarily on constraining memory.

src/translator/response.cpp
```diff
@@ -0,0 +1,141 @@
...
+  return remapped;
+}
+
+std::vector<Alignment> remapAlignments(const Response &first, const Response &second) {
+  std::vector<Alignment> alignments;
+  for (size_t sentenceId = 0; sentenceId < first.source.numSentences(); sentenceId++) {
+    const Alignment &sourceGivenPivots = first.alignments[sentenceId];
+    const Alignment &pivotGivenTargets = second.alignments[sentenceId];
+
+    // TODO: Allow range iterators and change algorithm, directly tapping into AnnotatedText
```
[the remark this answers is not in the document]
jerinphilip: Um yes, when the dust settles it might be possible to place this on AnnotatedText, #298 also pointing in that direction.

src/translator/batching_pool.cpp
jerinphilip: Less duplication? The if is expected to be called very less?

graemenail: Ok, the idea being the single resize in the constructor with `pivotSlack` should avoid (most of) the incremental resizes.

Is there some noticeable impact on non-pivoting with this change?

src/translator/batching_pool.cpp
```diff
@@ -24,11 +34,11 @@ size_t BatchingPool::generateBatch(Batch &batch) {
   batch.clear();
   size_t paddedBatchSize = 0;
 
-  for (size_t length = 0; length < bucket_.size(); length++) {
+  for (size_t length = 0; length <= maxActiveBucketLength_; length++) {
```
graemenail, line 37: Can this be reverted back to `bucket_.size()`?

src/translator/response.cpp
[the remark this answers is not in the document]
graemenail: Now that this PR is mentioned I'll resolve this.

src/translator/batching_pool.cpp
jerinphilip: There's `maxLengthBreak_ + pivotSlack` sizing initially. `bucket_.size()` could be `1.5*maxLengthBreak_` looping a bit more although empty. It's not structured and one more point of error, I don't like it much either, but....

src/translator/batching_pool.cpp
jerinphilip: Seconds for WNGT20, 40 threads.

pivot | main
-- | --
321.8 ± 4.08σ | 322.1 ± 5.2σ

src/translator/service.h
```diff
@@ -47,6 +47,24 @@ class BlockingService {
...
+  ///
+  /// @param[in] first: TranslationModel capable of translating from source language to pivot language.
+  /// @param[in] second: TranslationModel capable of translating between pivot and target language.
+  /// @param[move] sources: The input source texts to be translated.
+  /// @param[in] options: Options indicating whether or not to include optional members in response and pass additional
+  /// configurations. See ResponseOptions.
+  ///
+  /// @returns responses corresponding to the source-text which can be used as if they were translated with
+  /// translateMultiple.
+  std::vector<Response> pivotMultiple(std::shared_ptr<TranslationModel> first, std::shared_ptr<TranslationModel> second,
```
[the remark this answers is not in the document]
jerinphilip: It's already mentioned in some capacity. Resolving.

src/tests/common-impl.cpp
```diff
@@ -203,3 +225,74 @@ void TestSuite<Service>::translationCache(Ptr<TranslationModel> model) {
...
+  // Measure accuracy of word match.
+  float accuracy = static_cast<float>(favourableOutcomes) / static_cast<float>(totalOutcomes);
+
+  // This is arbitrary value chosen by @jerinphilip, but should be enough to check if things fail.
+  // This value is calibrated on bergamot input in BRT. All this is supposed to do is let the developers know if
+  // something's largely amiss to the point alignments are not working.
+  ABORT_IF(accuracy < 0.70, "Accuracy {} not enough. Please check if something's off.", accuracy * 100);
+
+  std::cout << response.source.text;
+  std::cout << response.target.text;
```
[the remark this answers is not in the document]
jerinphilip: This is a test-app where currently inputs holds extra `\n`. Doesn't make a different to the library user. Carrying forward for later.

graemenail (approved):
Thanks for addressing the comments @jerinphilip

PR #245, opened 2021-11-01
cache: threadsafety-fixes; optional stats collection

### LINKED ISSUE #246: Remove cache stats in default build
The default build should not be collecting caching statistics, especially when doing so starts requiring its own atomic operations.

jerinphilip:
Fixes #246. Also fixes an atomicity issue in cache-stats collection.


XapaJIaMnu:
I also think stats computation should be hidden behind a flag as atomic operations are much much more expensive than their non-atomic counterparts. I am not sure how much this matters in the real world. Would it be too much of a pain to add an extra flag?

jerinphilip:
@XapaJIaMnu No, think what you're suggesting is reasonable. Will do.

kpu:
The goal was to remove cache statistics in release builds; the statistics code shouldn't be there at all.

XapaJIaMnu:
@jerinphilip if you don't feel like fixing it properly, I'm ok to merge this now as a "better than nothing" solution. Merge conflicts need to be resolved as I merged the interface changes.

jerinphilip:
> Merge conflicts need to be resolved as I merged the interface changes.

Merge conflicts are resolved and this PR is now ready for review. It appears to me that the path being available in default build is useful for some diagnosis at [XapaJIaMnu/translateLocally@`6b2fb1d` (#75)](https://github.com/XapaJIaMnu/translateLocally/pull/75/commits/6b2fb1d160b901f1034ae7b27e9fac82ec9025aa), so let's leave this here and then eventually deprecate this for a complicated build based (through macros or whatever) one.

XapaJIaMnu:
@jerinphilip this PR is a draft, @jelmervdl was testing parameters. Atomic operations must be hidden behind a flag at the very least. @kpu prefers ifdef, but I don't mind either as long as I can easily turn them off.

jelmervdl:
For my/translateLocally use case it would be enough to have the cache implementation be an optional template parameter of AsyncService. I can then just use the implementation with stats, while the default interface (and the one exposed through the wasm bindings) use an optimised one. This will not make stats available to WASM, but is Mozilla interested in those stats anyway?

jerinphilip:
@XapaJIaMnu Can you let me know if [`6f955fc` (#245)](https://github.com/browsermt/bergamot-translator/pull/245/commits/6f955fc4422ab5d13f98fa266c9e35351dcc09c0) works for the compile-switch? 

Essentially what I have don is use the `COMPILE_TESTS` cmake variable to create a `TESTS_ENABLED` macro, which then ifdefs the content away when tests are not enabled.

XapaJIaMnu:
I don't think stats should be hidden behind compile tests, because there's a number of scenarios in which you'd want to see the stats, but not compile tests. 

Just give stats their own ifdef and set it to on in CMakeLIsts if compile tests is on. That way you get the current and people that want to get a different behaviour can enable it via a `-DENABLE_CACHE_STATS` or something?

src/translator/cache.h
```diff
@@ -26,7 +27,13 @@ class AtomicCache {
 
   void store(const Key &key, Value value) { atomicStore(key, value); }
 
-  const Stats stats() const { return stats_; }
+  const Stats stats() const {
+#ifdef ENABLE_CACHE_STATS
+    return Stats{hits_.load(), misses_.load()};
+#else
+    return Stats{0, 0};
```
XapaJIaMnu, line 34: How about adding an `std::cerr << "Warning: Cache stats disabled. To enable recompile with -DENABLE_CACHE_STATS=ON"<< std::endl;` (or use a logger, I think there is WARN_ONCE or something).

That way anyone who uses the `stats()` call when the stats are not tracked will know not to trust the result, as well as get a hint on how to enable them.

XapaJIaMnu (approved):
I've added a small suggestion, then you can merge.

src/translator/cache.h
jerinphilip: If somebody calls this without enabling the same in build, it's converted into a fatal abort, which should serve the same purpose of informing the user.

XapaJIaMnu: Not sure if it needs to be fatal, but as you decide. Merge when you feel like it

PR #250, opened 2021-11-01
Deprecate hardAlignment in favour of softAlignment

### LINKED ISSUE #248: Deprecate hard-alignments in favour of soft-alignments
Heads up to translateLocally people, @jelmervdl, @XapaJIaMnu.

For tag-tree operations, I've been informed to kill hardalignments (sparse) in favour of softalignments (dense).

jerinphilip:
Fixes #248.

### DIFF
--- a/src/translator/response.h
+++ b/src/translator/response.h
@@ -14,18 +14,6 @@
 namespace marian {
 namespace bergamot {
 
-/// Alignment is stored as a sparse matrix, this pretty much aligns with marian
-/// internals but is brought here to maintain translator
-/// agnosticism/independence.
-struct Point {
-  size_t src;  ///< Index pointing to source ByteRange
-  size_t tgt;  ///< Index pointing to target ByteRange
-  float prob;  ///< Score between [0, 1] on indicating degree of alignment.
-};
-
-/// Alignment is a sparse matrix, where Points represent entries with values.
-typedef std::vector<Point> Alignment;
-
 /// Response holds AnnotatedText(s) of source-text and translated text,
 /// alignment information between source and target sub-words and sentences.
 ///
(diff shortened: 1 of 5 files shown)


jerinphilip:
@abhi-agg Since I'm editing WASM bindings, providing some context. The original alignments were from a Fred era where there was a request for some mappings and were provisioned for experimentation. This however has not been exported for about 6 months via ResponseOptions, and the old extension as I understand has been deprecated and will not use these sparse-alignments. As far as I see, you haven't exposed alignments yet via WASM, so it's okay. 

To proceed with hard-alignments for a TagTree which @kpu, @qianqianzhu are coordinating currently for, I'm replacing the existing unified alignment with marian's hard-alignment, which is the dense-matrix structure with most information that they need.

@qianqianzhu's already done this delta for Inside-Outside (#199) and I have the same in my working copy of pivoting (#236). @kpu / @qianqianzhu please feel free to pull this into your working branch or collect it at main after review and merge.

qianqianzhu (dismissed):

src/translator/response.h
```diff
@@ -65,10 +53,10 @@ struct Response {
...
   std::vector<SentenceQualityScore> qualityScores;
 
-  /// Alignments between source and target. Each Alignment is a
-  /// sparse matrix representation with indices corresponding
-  /// to (sub-)words accessible through Annotation.
-  std::vector<Alignment> alignments;
+  /// Alignments between source and target. This is a collection of dense matrices providing
+  ///    P[t][s] = p(source-token s  | target token t)
+  /// with an alignment matrix for each sentence.
+  std::vector<marian::data::SoftAlignment> alignments;
```
kpu, line 59: `std::vector<std::vector<float> >` for better or worse

jerinphilip: You mean `std::vector<std::vector<std::vector<float>>>` (I'm sorry....)? This is `#sentences * # source-tokens * #target-tokens` (3D Tensor)

kpu: Yes, what I meant is you shouldn't refer to Marian class names, just substitute what it is defined as.

PR #258, opened 2021-11-05
Update ssplit cpp, pcre2 source compile to fix broken builds

jerinphilip:


jerinphilip:
This will be merged by @jerinphilip if all workflows succeed without awaiting review.

kpu:
Merging because it fixes most CI

PR #259, opened 2021-11-05
HTML input/output

jelmervdl:
Working on top of  #253 to add the missing bits, and get rudimentary HTML input and output working.
- [x] Unescape HTML entities https://github.com/bitextor/warc2text/blob/master/src/entities.cc  
Only `"'<>&` are decoded/encoded right now, I don't expect `DOMNode.innerHTML` to return any other.
- [ ] Convert some tags to spaces, possibly skip some with non-text content https://github.com/bitextor/warc2text/blob/39cb9d1eacefeb47c698fc97faf5204633b8d9d5/src/html.cc
- [x] Track were tags were removed and adjust source offsets
- [ ] Have some HTML object in Request that has the information to restore them.
- [ ] Restoring algorithm that, in the first iteration, will just insert them back in text proportion and escape entities.
- [ ] Actual alignment based insertion

This adds an `HTML` object that extracts text from HTML, and keeps track of these texts spans in the original HTML so it can insert all the removed bits back in when it sees these in the form of `AnnotatedText` instances.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -39,7 +39,7 @@ jobs:
     - name: Prepare vcpkg
       uses: lukka/run-vcpkg@v7.4
       with:
-        vcpkgArguments: protobuf pcre2
+        vcpkgArguments: protobuf 
         vcpkgGitCommitId: 8dddc6c899ce6fdbeab38b525a31e7f23cb2d5bb
         vcpkgDirectory: ${{ github.workspace }}/vcpkg/
         vcpkgTriplet: x64-windows-static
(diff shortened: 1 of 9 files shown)


jelmervdl:
Some notes of trouble I'm expecting when trying to reconstruct the HTML this way. I expect some of the whitespace to flip from outside to inside tags because sentencepiece prefers spaces at the start of tokens, and I'm only slicing HTML at the beginning or end of a token. This will pretty likely case some layout issues with websites that have these elements styled as inline HTML elements…

This is also why I'm using the original input source string for the source text, as opposed to the cut&glue reconstruction that will be necessary on the target side. 

(Notes below; bleu is text spans extracted from html, green is HTML spans, red is sentencepiece boundaries c.q. tokens for which we have alignment information)
6B7E497B-D4CB-4BAC-B570-F5A5A01CD99E copy

kpu:
Agreed being strictly at spm token boundaries will cause trouble with whitespace.  This seems like something we can worry about once there is alignment then track whitespace manually.

kpu:
Merging into branch.

Issue #264, opened 2021-11-16
Use optimized gemm implementation when available for wasm target


abhi-agg:
We want to be able to use the optimized gemm implementation when available in the environment, otherwise resort to using the portable fallback implementation of the [interface](https://github.com/browsermt/marian-dev/blob/master/src/tensors/cpu/wasm_intgemm_interface.h) for wasm target

kpu:
Related https://github.com/browsermt/marian-dev/issues/63

Issue #205, opened 2021-07-14
A built-in wasm 8-bit matrix multiply primitive


abhi-agg:
A summary of all the tasks that need to be done in this repository (and its submodules) to import matrix multiply function (based on 4x8-bit-to-32-bit dot product primitive) from a separate wasm module.

JS code will be able to instantiate a separate wasm module that exports a matrix multiply function and Bergamot code can, then, link against that instance to access that function.

Tasks Stage-1:
- [ ] [Import matrix-multiply from a separate wasm module](https://github.com/browsermt/bergamot-translator/issues/204)
- [ ] Develop a native matrix-multiply implementation optimized for Intel SSE4.1 architecture
- [ ] Test cases for this implementation
- [ ] Assist in landing Intel SSE4.1 optimized native matrix-multiply implementation in Firefox
- [ ] Test and Validate for SSE4.1 architecture that this solution works and provides performance improvements as expected

Tasks Stage-2:
- [ ] Develop a native portable (architecture agnostic) fallback code for matrix-multiply function
- [ ] Assist in landing this portable code in Firefox

Tasks Stage-3:
- [ ] Develop native matrix-multiply implementations for other architectures (Armv8.0/AVX2/Armv8.5? + additional to be decided later)
- [ ] Test cases for the implementations
- [ ] Assist in landing these implementations in Firefox
- [ ] Test bergamot for these architectures

cc @yurydelendik @andrenatal @lonnen @kpu @eqrion

Please suggest if I missed anything. I have created separate issues for 1 of the task to have the specific discussions there. Same can be done for other tasks as well as we go ahead.

P.S.: Stage-2 and Stage-3 can be ordered differently. Stage-1 has to be executed first.

jerinphilip:
Hello, just wondering if some work I'm doing at arm-playground relates/overlaps with the todos here. Could you let know if the links mentioned below are related?

1. Portability
	a. *Develop a portable (architecture agnostic) fallback implementation for matrix-multiply function for wasm*: <br/> I assume architecture agnostic means no presence of SIMD. Available via ruy as [`kStandardCppPath`](https://github.com/google/ruy/blob/8c3fd3f266b4a22d542d4aa41329b5018d6b87e1/ruy/path.h#L62)?
[...]

PR #265, opened 2021-11-16
Import optimized gemm implementation (when available) for wasm target

abhi-agg:
Fixes https://github.com/browsermt/bergamot-translator/issues/264

List of changes:
- Added file `wasm/import-gemm-module.js` that houses the logic to import and use the optimized gemm when available, otherwise fallback gemm
- Associated refactoring in `wasm/patch-artifacts-import-gemm-module.sh` for this change

Ran [wasm test page](https://github.com/browsermt/bergamot-translator/tree/main/wasm#demo) to test this logic. The optimized gemm is called when run on locally compiled version of firefox browser that exports a dummy implementation for gemm intrinsics. This proves that the logic written here works correctly to invoke an optimized gemm when available in the environment.

cc @yurydelendik @eqrion

### DIFF
--- a/wasm/import-gemm-module.js
+++ b/wasm/import-gemm-module.js
@@ -1,0 +1,23 @@
+
+/* Use an optimized gemm implementation if available, otherwise use the fallback
+ * implementation.
+ */
+function createWasmGemm() {
+    const OPTIMIZED_GEMM = "mozIntGemm";
+    const FALLBACK_GEMM =  "asm";
+
+    if (WebAssembly[OPTIMIZED_GEMM]) {
+        return new WebAssembly.instance(WebAssembly[OPTIMIZED_GEMM](), {"": {memory: wasmMemory}}).exports;
+    }
+    else {
+        return {
+            "int8_prepare_a": (...a) => Module[FALLBACK_GEMM]["int8PrepareAFallback"](...a),
+            "int8_prepare_b": (...a) => Module[FALLBACK_GEMM]["int8PrepareBFallback"](...a),
+            "int8_prepare_b_from_transposed": (...a) => Module[FALLBACK_GEMM]["int8PrepareBFromTransposedFallback"](...a),
+            "int8_prepare_b_from_quantized_transposed": (...a) => Module[FALLBACK_GEMM]["int8PrepareBFromQuantizedTransposedFallback"](...a),
+            "int8_prepare_bias": (...a) => Module[FALLBACK_GEMM]["int8PrepareBiasFallback"](...a),
+            "int8_multiply_and_add_bias": (...a) => Module[FALLBACK_GEMM]["int8MultiplyAndAddBiasFallback"](...a),
+            "int8_select_columns_of_b": (...a) => Module[FALLBACK_GEMM]["int8SelectColumnsOfBFallback"](...a)
+        }
+    }
+}
--- a/wasm/import-gemm-module.js
+++ b/wasm/import-gemm-module.js
@@ -8,7 +8,7 @@ function createWasmGemm() {
 
     if (WebAssembly[OPTIMIZED_GEMM]) {
         console.log(`Using optimized gemm (${OPTIMIZED_GEMM}) implementation`);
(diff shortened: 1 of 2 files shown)


kpu:
Is this the slow fallback that bounces WASM -> JS -> WASM with those bindings?

yurydelendik:
> Is this the slow fallback that bounces WASM -> JS -> WASM with those bindings?

By reading the code, it is both. If `WebAssembly.mozIntGemm` is present, the created instance will use intrinsic functions.

kpu:
Ok I guess this gets it to link which is good.  We should avoid calling these functions in the fallback case, but that can come later.

kpu (approved):
Approving this but opening https://github.com/browsermt/marian-dev/issues/66 to track the dangling performance issue if the intrinsic is missing.

jerinphilip:
Hello. I'm trying to better understand how the sources in JS and C++ interoperate for achieving speed up as I progress with ruy. 

> The optimized gemm is called when run on locally compiled version of firefox browser that exports a dummy implementation for gemm intrinsics. This proves that the logic written here works correctly to invoke an optimized gemm when available in the environment.

@abhi-agg  Can I be pointed to a diff that holds this dummy implementation? Is there a sharable link, perhaps as a mercurial diff? 

> Is this the slow fallback that bounces WASM -> JS -> WASM with those bindings?

@kpu I'm slightly confused by what is intended here, may I request get some details? The fallback is currently intgemm path compiled by emscripten to WASM (which would be SSSE3) and all of this is in WASM? Where does the JS bit mentioned above come in? If `createWasmGemm` is called at the first run setting the switch between WebAssembly Gemm (emitted by emscripten) and native gemm available via firefox, doesn't this happen only at load and not every call? What is meant by "bindings" in the above text?

yurydelendik:
>  Can I be pointed to a diff that holds this dummy implementation? Is there a sharable link, perhaps as a mercurial diff?

* I prototyped idea intrinsic at https://gist.github.com/yurydelendik/a913d0894615c7712babb43e842044ef
* Fully working prototype for old version of firefox and bergamot at https://gist.github.com/yurydelendik/cc02ba86128ed46d622e8c3099c8a510

> Where does the JS bit mentioned above come in? If createWasmGemm is called at the first run setting the switch between WebAssembly Gemm (emitted by emscripten) and native gemm available via firefox, doesn't this happen only at load and not every call? What is meant by "bindings" in the above text

`createWasmGemm` will be called once during wasm instance creation. Look at "bindings" as static linking (of shared library).


> Is this the slow fallback that bounces WASM -> JS -> WASM with those bindings?

For phase one, `WASM_{bergamot_module}` -> `JS` -> `WASM_{bergamot_module}` just to make implementation quicker and provide some proves that there no internal CPU-related code passes in the `bergamot_module` that we could miss. For next phase, it will be `WASM_{bergamot_module}` -> `WASM_{intgemm_core_module}`. I recommend to add the `memory` parameter to the createWasmGemm function to make thing more descriptive, instead of using `wasmMemory`. Notice `intgemm_core_module` is a separate build artifact and will require time to create, but it will remove JS bits.

kpu:
Can we just eliminate

`WASM_{bergamot_module}` -> `JS` -> `WASM_{bergamot_module}`
`WASM_{bergamot_module}` -> `WASM_{intgemm_core_module}` 

by having the C++ code call itself internally bypassing the C functions?  Then you would tell Marian if it has the intrinsic.  Or it would be detected by way of having an intentionally incorrect C function implementation.

yurydelendik:
> Can we just eliminate [them]?

It depends on architectural decision you are going to make. I see the only way to achieve that: via `if(some_global_setting) { } else {}` near every intrinsic call. This approach has only cons: increases cyclomatic complexity, code duplication, the optimal code path will not be tested, the wasm intgemm code will be present in the bergamot executable. The latter allows possibility that some code path will invoke that when the intrinsics in use, which will make benchmark results incorrect. From correctness of the result point of view, all of that really does not matter.

kpu:
Let's take this where it belongs in #205 instead of some random commit.

PR #266, opened 2021-11-27
HTML handling improvements

jelmervdl:
This merge request contains fixes for several issues:

1. If the model did not have `alignment:soft` in its configuration, `HardAlignment()` would crash & burn.
2. If you had a model that cut the first word of an output into multiple tokens, and the first of those tokens was not the token that had the highest alignment score of all target tokens for that word, there was an invalid array access (because `(size_t)0 - 1 >= 0`…)
3. If you had an empty (i.e. what in XHTML was a self-closing) element at the end of a parent's element contents, or the end of a sentence, you'd get a parse error because that empty element would stay on the stack too long.
4. `&nbsp;` was not recognised. It is now, but translated to a single "␣". So if there was a non-breaking space in the source HTML, it will not correctly be reconstructed. If this is a problem, we could treat `&nsbp;` as a special sort of formatting/taint for spaces and handle it that way.
5. This contains a partial fix for losing elements that have no text in them, e.g. `hello <u></u> world`. Previously the `<u></u>` would be lost. Now an empty span is added with the `u` as markup, which then gets opened & closed before `_world`. These empty spans are not yet copied over to the target side, so translated output will still lose empty tags.

### DIFF
--- a/src/translator/html.cpp
+++ b/src/translator/html.cpp
@@ -166,6 +166,10 @@ AnnotatedText Apply(AnnotatedText const &in, Fun fun) {
 
 bool IsContinuation(string_view str) { return !str.empty() && str.compare(0, 1, " ", 1) != 0; }
 
+bool HasAlignments(Response const &response) {
+  return !response.alignments.empty() && !response.alignments[0][0].empty();
+}
+
 void HardAlignments(Response const &response, std::vector<std::vector<size_t>> &alignments) {
   // For each sentence...
   for (size_t sentenceIdx = 0; sentenceIdx < response.target.numSentences(); ++sentenceIdx) {
@@ -512,7 +516,7 @@ void HTML::Restore(Response &response) {
   // tokens with the tags from their source token counterpart. If there is no
   // alignment information available, we just interpolate based on sentence
   // length (badly).
-  if (!response.alignments.empty()) {
+  if (HasAlignments(response)) {
     // DebugPrintAlignmentScores(std::cerr, response);
     HardAlignments(response, alignments);
   } else {
--- a/src/translator/html.cpp
+++ b/src/translator/html.cpp
@@ -94,7 +94,7 @@ bool IsBlockElement(std::string const &name) {
   return inline_ish_elements.find(name) == inline_ish_elements.end();
 }
 
-bool IsEmtpyElement(std::string const &name) {
+bool IsEmptyElement(std::string const &name) {
   // List of elements for which we do not expect a closing tag, or self-closing
   // elements in XHTML. See also https://developer.mozilla.org/en-US/docs/Glossary/Empty_element
   static std::unordered_set<std::string> empty_elements{"area",  "base", "br",   "col",   "embed",  "hr",    "img",
@@ -130,6 +130,10 @@ void FilterEmpty(HTML::Taint &stack) {
     if (!(*src)->empty) *(dst++) = *src;
 
   stack.resize(dst - stack.begin());
+}
+
+bool ContainsTag(HTML::Taint const &stack, HTML::Tag const *tag) {
+  return std::find(stack.rbegin(), stack.rend(), tag) != stack.rend();
 }
 
 template <typename Fun>
@@ -266,6 +270,14 @@ AnnotatedText RestoreSource(AnnotatedText const &in, std::vector<HTML::Taint> &t
     size_t offset = 0;  // Size added by prepending HTML
     size_t whitespace_size = CountPrefixWhitespaces(token);
 
+    // Close tags we want to show up left (before) the token, but open tags
+    // ideally come directly after any prefix whitespace. However, some tokens
+    // match multiple spans. If a previous span has added an open tag, after any
+    // whitespace, and the next span closes said tag again, we need to close
+    // it after the whitespace. So after the first open tag, any closing tag
+    // should also align right, after whitespace, not before. Hence this bool.
+    bool close_left = true;
+
     // Potential issue: spans and tokens can intersect, e.g.
     //
     //    text  <p> h <u> e </u> ll o </p>
@@ -281,7 +293,7 @@ AnnotatedText RestoreSource(AnnotatedText const &in, std::vector<HTML::Taint> &t
 
       for (auto cit = closing.crbegin(); cit != closing.crend(); ++cit) {
         std::string close_tag = format("</{}>", (*cit)->name);
-        html.insert(offset, close_tag);
+        html.insert(offset + (close_left ? 0 : whitespace_size), close_tag);
         offset += close_tag.size();
       }
 
@@ -289,6 +301,7 @@ AnnotatedText RestoreSource(AnnotatedText const &in, std::vector<HTML::Taint> &t
         std::string open_tag = format("<{}{}>", tag->name, tag->attributes);
         html.insert(offset + whitespace_size, open_tag);
         offset += open_tag.size();
+        close_left = false;
       }
 
       if (span_it + 1 != span_end && ((span_it + 1)->begin < range.end || last)) {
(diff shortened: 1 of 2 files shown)


jerinphilip (commented):
Leaving one tiny nit. The code is already too complex for me to line by line review. I will experiment a bit with my lemonade fork through python and if all OK, I will merge directly by midnight. 

Thanks for taking the time out for a hotfix.

src/translator/html.cpp
```diff
@@ -449,13 +468,19 @@ HTML::HTML(std::string &&source, bool process_markup) {
         // separate words
         if (IsBlockElement(scanner.get_tag_name()) && !source.empty() && source.back() != ' ') source.push_back(' ');
 
-        pool_.emplace_back(new Tag{
-            scanner.get_tag_name(), std::string(),
-            IsEmtpyElement(scanner.get_tag_name())  // TODO: detect empty elements by doing a second pass and detecting
-                                                    // non-closed elements?
-        });
+        tag = new Tag{scanner.get_tag_name(), std::string(), IsEmptyElement(scanner.get_tag_name())};
```
jerinphilip, line 471: unique_ptr, github outage ate my nit comment, so pointing out again.

jerinphilip: On a closer look, turns out `pool_` is an `std::vector<std::unique_ptr>>`. This is my bad, the `Tag *` unassigned above and the `new` keyword here raised some alarms. Is it safe to assume no leak?

Would `std::make_unique<...>(...)` followed by a `std::move(...)` be more readable? Alternatively, how about we move the `new Tag{...}` block inside `emplace_back(...)` with a note `pool_` is a container holding unique-pointer?

jerinphilip:
Merging owing to the critical nature. Comments related to new can be addressed in a later change.

src/translator/html.cpp
jelmervdl: There is a note saying pool_ takes ownership, but I agree the previous code where `Tag *tag` didn't exist and I just referenced `bool_.back()` might have been clearer about ownership. `Tag *tag` was added to make the attribute parsing not depend on an element being on the stack.

I'll flip it around, do something like `tag = pool_.emplace_back(new Tag{}).get()` (or C++11 equivalent) again to make it clearer `pool_` owns them.

The idea of `pool_` was to have something like [kpu/preprocess/util/pool.hh:Pool](https://github.com/kpu/preprocess/blob/master/util/pool.hh#L15) eventually to save on dynamic allocations but it would be a tiny optimisation given how much string allocations I'm already doing in this bit of code anyway.

jerinphilip: Please excuse my lack of context regarding the details. 

If I may ask a conceptual question without knowing the full details, what is the concern here that keeps things at `std::vector<std::unique_ptr<T>>` instead of `std::vector<T>`? If it's pointer invalidation due to container growth `bool_.back()` might not be possible. If it's not pointer invalidation, you can probably get rid of the repeated allocations, I think by achieving the same with `std::vector<T>`?

jelmervdl: Just pointer invalidation. The `Taint` structure (which is `std::vector<Tag*>`) that's attached to each span refers to and compares tags by their address. Mainly for efficiency. And because at some point it made sense to have tags refer to other tags through parent/child relationships (but that's removed in this version). The `pool_` is there to just make sure all Tags are freed once `HTML` goes out of scope.

Technically I think I could just copy Tag structs to where they are needed, I'm currently not mutating tags after they have been assigned to a span. But that might be pretty expensive due to each tag also having a couple of strings associated with them, and as far as I understand, `std::string` is not copy-on-write.

kpu:
Hmmm I thought we were doing space-separated word rounding so it's weird to see stuff split in the middle like that.  The system is probably also struggling since these are rare tokens.  Our of curiosity,

jelmervdl:
In this particular case (since I can't exactly replicate it) could you check it is actually using alignment information? The model configuration in https://github.com/browsermt/bergamot-translator/blob/main/wasm/test_page/js/worker.js#L172 might need a `alignment: soft` entry.

If alignment info is not available, it falls back to just guessing positions based on offsets, and it's pretty bad at that. That might be why it's cutting words in half here. When alignment info is available, it generally does a better job at keeping words consistent at least.

Although, in this case, for me it completely loses the Spectre (en) -> Spectre (de) relationship and the link is gone when I try it with the online model. I'll have a look at what's going on there. But it might be what Kenneth implied, related to rare tokens.  Maybe they and untranslated entities would need some special treatment.

jerinphilip:
I can reproduce:

```html
[src] >  <div id="0"><br><b id="1">(hace 50 años</b>): Fallece <b id="2">Chandrasekhara Raman</b>, físico indio (n. 1888; <b id="3">en la imagen</b>), premio Nobel de física en 1930.</div>
[hyp] >  <div id="0"><b id="1">(Hace 50 a-os</b>): Fallece <b id="2">Chandrasekhara Raman</b>, f-sico indio (n. 1888; <b id="3">en la imagen</b>), prämio Nobel de fesica en 1930.</div>
[tgt] >  <div id="n0"><br><b id="n1">(50 years ago):</b> Chandrasekhara Raman, <b id="n2">Indian physicist,</b> dies (n 1888; in pictures), <b id="n3">Nobel laureate in</b> 1930.</div>

[src] >  In early 2018, security researchers disclosed two major vulnerabilities, known as <a href="https://en.wikipedia.org/wiki/Meltdown_(security_vulnerability)" target="_blank">Meltdown</a> and <a href="https://en.wikipedia.org/wiki/Spectre_(security_vulnerability)" target="_blank">Spectre</a>.
[hyp] >  Anfang 2018 enthüllten Sicherheitsforscher zwei große Schwachstellen, bekannt als Meltdown <a href="https://en.wikipedia.org/wiki/Meltdown_(security_vulnerability)" target="_blank">und Spec</a>tre<a href="https://en.wikipedia.org/wiki/Spectre_(security_vulnerability)" target="_blank">.</a>
```

https://github.com/jerinphilip/tagtransfer/blob/bf76a17aa0de7d60c02c9a3ade51e8b48595f566/tagtransfer/examples.py#L156-L160

jelmervdl:
There's also a bug in the token-to-word generalisation code: it only goes backwards, not forwards.  That would at least explain the half word links.

I've been looking at the alignment scores between spectre and spectre and that one might be a bit more challenging to solve.

jerinphilip:
> Do I need to change the ResponseOptions::alignment as well?

For now please go ahead with a yes, setting it in JS so development can parallel progress, thank you for pointing out. https://github.com/browsermt/bergamot-translator/blob/eea5554b91dceea0e51a628adc844d7fc1e7ae85/src/translator/response_builder.h#L64-L67. 

This is not strictly what is desired, we will work towards a solution for fixing this.

abhi-agg:
> > Do I need to change the ResponseOptions::alignment as well?
> 
> For now please go ahead with a yes, setting it in JS so development can parallel progress, thank you for pointing out.
> 
> https://github.com/browsermt/bergamot-translator/blob/eea5554b91dceea0e51a628adc844d7fc1e7ae85/src/translator/response_builder.h#L64-L67
> .
> 
> This is not strictly what is desired, we will work towards a solution for fixing this.

Wait. The whole point of keeping [ResponseOptions::alignment](https://github.com/browsermt/bergamot-translator/blob/main/src/translator/response_options.h#L20) was whether to return alignments in the response or not. The `alignment: soft` config for translator construction is only for the alignment matrix to be present from marian translation pipeline (as you said).

I imagine that setting [ResponseOptions::alignment](https://github.com/browsermt/bergamot-translator/blob/main/src/translator/response_options.h#L20) to `true` would require a translator constructed with `alignment: soft` config but I can specify `alignment: soft` config for translator and not set [ResponseOptions::alignment](https://github.com/browsermt/bergamot-translator/blob/main/src/translator/response_options.h#L20) to `true`. Right?

PR #275, opened 2021-12-14
Proper arch setting on win32

XapaJIaMnu:
@jerinphilip  with this, `DBUILD_ARCH` should work properly for all platforms. I assume you're already passing  it for MacOS build?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,6 +20,35 @@ if(NOT CMAKE_BUILD_TYPE)
   message(WARNING "CMAKE_BUILD_TYPE not set; setting to Release")
   set(CMAKE_BUILD_TYPE "Release")
 endif()
+
+if(NOT COMPILE_WASM)
+  # Setting BUILD_ARCH to native invokes CPU intrinsic detection logic below.
+  # Prevent invoking that logic for WASM builds.
+  set(BUILD_ARCH native CACHE STRING "Compile for this CPU architecture.")
+
+  # Unfortunately MSVC supports a limited subset of BUILD_ARCH flags. Instead try to guess
+  # what architecture we can compile to reading BUILD_ARCH and mapping it to MSVC values
+  # references: https://clang.llvm.org/docs/UsersManual.html https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html https://gcc.gnu.org/onlinedocs/gcc-4.8.5/gcc/i386-and-x86-64-Options.html
+  # https://docs.microsoft.com/en-us/cpp/build/reference/arch-x86?redirectedfrom=MSDN&amp;amp;view=vs-2019&view=msvc-170 https://devblogs.microsoft.com/oldnewthing/20201026-00/?p=104397
+  # This is by no means an exhaustive list but should match the most common flags Linux programmers expect to parse to MSVC
+  if(MSVC)
+    if(BUILD_ARCH STREQUAL "native") # avx2 is good default for native. Very few desktop systems support avx512
+      set(MSVC_BUILD_ARCH "/arch:AVX2")
+    elseif(BUILD_ARCH STREQUAL "skylake-avx512" OR BUILD_ARCH STREQUAL "cannonlake" OR BUILD_ARCH STREQUAL "x86-64-v4" OR BUILD_ARCH STREQUAL "tigerlake" OR BUILD_ARCH STREQUAL "cooperlake" OR BUILD_ARCH STREQUAL "cascadelake")
+      set(MSVC_BUILD_ARCH "/arch:AVX512")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -20,10 +20,39 @@ if(NOT CMAKE_BUILD_TYPE)
...
+  set(BUILD_ARCH native CACHE STRING "Compile for this CPU architecture.")
+
+  # Unfortunately MSVC supports a limited subset of BUILD_ARCH flags. Instead try to guess
+  # what architecture we can compile to reading BUILD_ARCH and mapping it to MSVC values
+  # references: https://clang.llvm.org/docs/UsersManual.html https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html https://gcc.gnu.org/onlinedocs/gcc-4.8.5/gcc/i386-and-x86-64-Options.html
+  # https://docs.microsoft.com/en-us/cpp/build/reference/arch-x86?redirectedfrom=MSDN&amp;amp;view=vs-2019&view=msvc-170 https://devblogs.microsoft.com/oldnewthing/20201026-00/?p=104397
+  # This is by no means an exhaustive list but should match the most common flags Linux programmers expect to parse to MSVC
+  if(MSVC)
+    if(BUILD_ARCH STREQUAL "native") # avx2 is good default for native. Very few desktop systems support avx512
+      set(MSVC_BUILD_ARCH "/arch:AVX2")
```
jerinphilip, line 36: As someone not aware of the source, I'd expect `BUILD_ARCH=native` to be equivalent to `-march=native`.

jerinphilip (commented):
> I assume you're already passing it for MacOS build?

I don't fully understand what's going on here. Everything compiles with `-march=native` on clang and GCC at the moment, which is whatever is highest on the particular hardware (`-mavx512 ...`  etc added by the compiler). This was inherited from Marian at the first integration PR. 

The windows half is untested by means of BRT. Mozilla achieves cross-platform by means of WASM and import of gemm into firefox sources, so windows/mac/Linux is probably a problem only at intgemm level - as WASM already compiles. 

I find it confusing to give `arch=native` a different meaning than [1] by means of hard-coding it in CMake as `avx2`, which the MSVC half here appears to be doing, but I'm not very experienced in maintaining cross-platform builds. 

{1] https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html

CMakeLists.txt
```diff
@@ -80,11 +109,6 @@ include(GetVersionFromFile)
 message(STATUS "Project name: ${PROJECT_NAME}")
 message(STATUS "Project version: ${PROJECT_VERSION_STRING_FULL}")
 
-if(NOT COMPILE_WASM)
-  # Set BUILD_ARCH to native only while compiling for non wasm platform
-  set(BUILD_ARCH native CACHE STRING "Compile for this CPU architecture.")
-endif()
```
jerinphilip, line 86: https://github.com/browsermt/bergamot-translator/runs/4523341672?check_suite_focus=true#step:9:78

Any idea what's triggering `native` is coming in despite the above lines being removed?

CMakeLists.txt
XapaJIaMnu: MSVC doesn't have an equivalent of native. The current code (before my changes uses AVX2 for windows builds. 
https://docs.microsoft.com/en-us/cpp/build/reference/arch-x86?redirectedfrom=MSDN&amp=&view=msvc-170

Since we don't have native on windows, I'm just mimicking old behaviour.

CMakeLists.txt
XapaJIaMnu: The lines above are moved to the top of the `CMakeLists.txt`. Essentially what this patch does, is keep old behaviour unchanged and try to get Windows compilation to accept some clang/g++ like `-march` flags.

jerinphilip: Oops. My bad. Resolving.

XapaJIaMnu:
What is going on here is that it makes it possible for you to properly compile marian/bergamot-translator with `'march=XXXX`, propagating it down to Windows and  removing the hardcoded avx2 windows version.

As a bonus you can also provide `-DBUILD_ARCH=core-avx-i`  (for example) to get a binary that is compiled against the AVX instruction set.

Sadly there isn't a 1:1 correspondence between gcc/clang architectures and  MSVC architectures so I have done some partial mapping that will work for most cases.

jerinphilip:
Understanding this to be a net improvement, looks good to me for merge. 

ARM builds don't seem to support the `native` argument default. This is not a problem yet but soon will be. The current organization in `CMakeLists.txt` does not look ideal (I don't know what is ideal, tbh). If the position is we don't need strong backwards compatibility in build (retaining previous behaviour seems to be taken into consideration), could what's being done here be achieved in a better way?

XapaJIaMnu:
I'm not sure what could be a better organisation. The windows specific bits are partially redundant but I'm unwilling to spend time changing this away from the way done in upstream marian as I don't understand the windows build well enough.

However I would like to make sure arm is supported. What flags does it use? I'll do the modifications here

jerinphilip (approved):
> What flags does it use? I'll do the modifications here

Let's address this later. Trying to do future stuff without a build is going to be quite wasteful. Just wanted to highlight the ARM incoming to initiate some thought and to point we'll be reworking parts of this eventually.

PR #276, opened 2021-12-14
More robust logic to import wasm gemm

abhi-agg:
- Import optimized gemm implementation only if all the necessary functions
   are provided by it, othewise use the fallback gemm

@andrenatal This should fix the issue that you were encountering in the extension


abhi-agg:
Merging this one as documentation CI failure has got nothing to do with this PR and it is blocking extension.

PR #279, opened 2021-12-16
Increase HTML test coverage

jelmervdl:
This pull request covers adding unit tests until I'm close to 100% test coverage of the xh_scanner and HTML code. And fix any issues I identify while doing that.

Issues found and fixed:
- Accessing `scanner.value()`, `attr_name()` or `tag_name()` before a relevant token had been identified would point to random memory, which could lead to segfaults.
- Parsing `<script>` and `<style>` tags. These were treated as normal elements, but HTML spec allows use of unescaped entities inside them. It's a bit pointless to send these to the translation service anyway.
- Parsing around unquoted and valueless attributes, specifically when mixed. I don't think `Element.innerHTML` would ever give you any of these, but if external HTML is ever allowed (I see a web service being created…) support might come in handy. xh_scanner already had some support for it anyway.

### DIFF
--- a/src/translator/html.cpp
+++ b/src/translator/html.cpp
@@ -174,10 +174,17 @@ bool HasAlignments(Response const &response) {
   // Test for each sentence individually as a sentence may be empty (or there)
   // might be no sentences, so just testing for alignments.empty() would not be
   // sufficient.
-  for (size_t sentenceIdx = 0; sentenceIdx < response.target.numSentences(); ++sentenceIdx)
+  for (size_t sentenceIdx = 0; sentenceIdx < response.target.numSentences(); ++sentenceIdx) {
+    // If response.alignments is just empty, this might catch it.
     if (response.alignments.size() <= sentenceIdx ||
         response.alignments[sentenceIdx].size() != response.target.numWords(sentenceIdx))
       return false;
+
+    // If response.alignments is "empty" because the model did not provide alignments,
+    // it still has entries for each target word. But all these entries are empty.
+    for (size_t wordIdx = 0; wordIdx < response.target.numWords(sentenceIdx); ++wordIdx)
+      if (response.alignments[sentenceIdx][wordIdx].size() != response.source.numWords(sentenceIdx)) return false;
+  }
   return true;
 }
 
--- a/src/translator/xh_scanner.h
+++ b/src/translator/xh_scanner.h
@@ -52,7 +52,10 @@ class scanner {
   };
 
  public:
-  explicit scanner(instream &is) : input_(is), got_tail(false) { c_scan = &scanner::scan_body; }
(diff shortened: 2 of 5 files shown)


src/translator/xh_scanner.h
```diff
@@ -80,6 +109,9 @@ class scanner {
   // Consumes <!-- ... -->
   token_type scan_comment();
 
+  // Consumes <?name [attrs]?>
+  token_type scan_pi();
```
jerinphilip, line 113: Why is this called 'pi'?

jerinphilip (approved):
PR looks good to me, I want to merge when ready so more extensive testing is possible in the wild (through python and HTML). Currently, there are fixes sitting in multiple branches.

> This pull request covers adding unit tests until I'm close to 100% test coverage of the xh_scanner and HTML code. And fix any issues I identify while doing that.

Would it be possible to integrate the coverage stats with Catch be integrated first class into CI system? 

Few nits/queries are inline. There are at least three styles (`snake_case`, `CamelCase`, `camelBack`) in this PR. The clang-tidy check is configured to complain although not enforced (must be in the formatting CI check logs). 

Let's merge in the interest of easing things for everyone and bring the remaining changes incrementally.

src/translator/xh_scanner.h
jelmervdl: It was named `scan_pi` in the imported codebase. I assume because scan_processing_instruction was too long to type.

src/translator/html.cpp
```diff
@@ -94,11 +94,13 @@ bool IsBlockElement(std::string_view const &name) {
   return inline_ish_elements.find(std::string(name)) == inline_ish_elements.end();
 }
 
-bool IsEmptyElement(std::string_view const &name) {
+bool IsVoidTag(std::string_view const &name) {
   // List of elements for which we do not expect a closing tag, or self-closing
   // elements in XHTML. See also https://developer.mozilla.org/en-US/docs/Glossary/Empty_element
-  static std::unordered_set<std::string> empty_elements{"area",  "base", "br",   "col",   "embed",  "hr",    "img",
-                                                        "input", "link", "meta", "param", "source", "track", "wbr"};
+  // More relevant source of this list: https://searchfox.org/mozilla-central/source/dom/base/FragmentOrElement.cpp#1791
```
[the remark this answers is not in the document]
jelmervdl: Good catch! I'll replace it with a link to a commit instead of a branch.

jelmervdl:
You're right the naming conventions were a mess. I've tried to match the style of the rest of the codebase. I stopped half-way through html.cpp though because #283 will already rename a bunch of variables. Making my merge harder, as well as wasted effort renaming variables now.

PR #280, opened 2021-12-17
Adding circle ci job to push the wasm artifacts to github releases

andrenatal:
We need a way to auto publish builds to github releases to facilitate the integration with the extension, so this patch introduces that to circle ci whenever a tag is created. This job can remain disabled here, since we intend to use it only on [github.com/mozilla/bergamot-translator](https://github.com/mozilla/bergamot-translator).

Artifacts and releases will be created like this: https://github.com/andrenatal/bergamot-translator/releases

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,7 +27,7 @@ jobs:
               cp bergamot-translator-worker.wasm ../artifacts/bergamot-translator-worker-with-wormhole.wasm
               cp bergamot-translator-worker.js ../artifacts/bergamot-translator-worker-with-wormhole.js
               shasum -a 256 ../artifacts/* > ../artifacts/SHA256-1
-              cp BERGAMOT_VERSION ../artifacts/
+              cp ../BERGAMOT_VERSION ../artifacts/
             else
               echo "Failure: Artifacts Not Present"
               exit 1
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,11 +23,11 @@ jobs:
             if ls bergamot*.wasm &>/dev/null && ls bergamot*.js &>/dev/null
             then
               echo "Artifacts Successfully Generated"
-              mkdir artifacts
-              cp build-wasm/bergamot-translator-worker.wasm artifacts/bergamot-translator-worker-with-wormhole.wasm
-              cp build-wasm/bergamot-translator-worker.js artifacts/bergamot-translator-worker-with-wormhole.js
-              shasum -a 256 ./artifacts/* > ./artifacts/SHA256-1
-              cp BERGAMOT_VERSION ./artifacts/
+              mkdir ../artifacts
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -11,9 +11,14 @@ jobs:
...
-          command: bash build-wasm.sh WORMHOLE
-
+          name: Build WASM WORMHOLE
+          command: |
+            bash build-wasm.sh WORMHOLE
+            mkdir artifacts
+            cp build-wasm/bergamot-translator-worker.wasm artifacts/bergamot-translator-worker-with-wormhole.wasm
+            cp build-wasm/bergamot-translator-worker.js artifacts/bergamot-translator-worker-with-wormhole.js
+            shasum -a 256 ./artifacts/* > ./artifacts/SHA256-1
+            cp BERGAMOT_VERSION ./artifacts/
```
abhi-agg, line 21: @andrenatal I believe, the steps to copy the artifacts to "artifacts" folder should happen after the next step "Check artifacts" which checks whether artifacts are present or not.

.circleci/config.yml
```diff
@@ -43,8 +53,12 @@ jobs:
...
       - run:
           name: Build WASM
-          command: bash build-wasm.sh
-
+          command: |
+            bash build-wasm.sh
+            mkdir artifacts
+            cp build-wasm/bergamot-translator-worker.wasm artifacts/bergamot-translator-worker-without-wormhole.wasm
+            cp build-wasm/bergamot-translator-worker.js artifacts/bergamot-translator-worker-without-wormhole.js
+            shasum -a 256 ./artifacts/* > ./artifacts/SHA256-2
```
abhi-agg, line 61: same here.

.circleci/config.yml
```diff
@@ -57,13 +71,55 @@ jobs:
...
-          - build-without-wormhole
\ No newline at end of file
+          - build-with-wormhole:
+              filters:
+                tags:
+                  only: /^v.*/
+          - build-without-wormhole:
+              filters:
+                tags:
+                  only: /^v.*/
```
abhi-agg, line 114: A question. Does these steps mean "build-with-wormhole" and "build-without-wormhole" jobs will run only when a new tag is created? We want this behavior only for "publish_to_github" job. Right?

abhi-agg: Never mind. This is correct.

.circleci/config.yml
andrenatal: ok, fixed

abhi-agg (dismissed):
Lgtm.

andrenatal:
I need this to land so I could turn on on mozilla's fork. It's hurting our workflow. As I said, it can remain disabled in this repo

jerinphilip (approved):

andrenatal:
Thanks @jerinphilip !

abhi-agg (approved):

PR #283, opened 2021-12-19
HTML transfer empty elements

### LINKED ISSUE #267: Empty elements not copied to translated output
This bug is to track a missing feature in the HTML parsing/reconstruction.

Empty elements (both things like `` and `<u></u>` without any text between them) are not properly copied over to the translated output.

jelmervdl:
Naive implementation of tag insertion. Fixes #267.

Brief description of how it works:

1. The HTML is chopped up into spans. Each HTML element is described as markup applied to a span of text. That text is the input for the translation model. Empty and void elements such as `` are described as markup applied to an always empty span. That way their position is remembered relative to the surrounding text.
2. During translation, the input text is chopped into tokens. These are then translated, and using alignment information we can link source and target tokens to each other.
3. For each source token we figure out which span they fall in, and then associate that span (and its markup) with the linked target token.
4. When reconstructing the target HTML, we walk through all target tokens, and the spans associated with them, to figure out when to open and close an HTML element. Newly added is the bit that we also keep an eye on these spans' positions in the original input HTML. If between the spans of the previous target token and the current target token there are any of these empty spans, (and they haven't been inserted yet), we include their HTML before the current target token.

Known "issues" atm (assuming #279 is merged already):

1. `<script>` and `<style>` tags will be empty, their contents lost. For the browser extension use-case this shouldn't be an issue, why would you send these elements for translation? For use cases like https://github.com/jerinphilip/tagtransfer/pull/1 it is useful.
2. Happily translates the contents of `<code>`, `<kbd>`, `<samp>`, `<var>`, elements that are probably better left untranslated. (Just took a quick peek, Google Translate seems to exclude `<applet>` `<area>` `<base>` `<br>` `<code>` `<frame>` `<frameset>` `<hr>` `` `<input>` `<kbd>` `<link>` `<map>` `<meta>` `<noframes>` `<noscript>` `<object>` `<param>` `<script>` `<style>` `<svg>` `<textarea>` `<title>` `<wbr>`)

I'm waiting with marking this one ready for #279 to be merged. That will also make it easier to fix any outstanding issues.


src/translator/html.cpp
```diff
@@ -336,48 +330,86 @@ AnnotatedText RestoreTarget(AnnotatedText const &in, std::vector<HTML::Taint> co
...
+    // First we scan through spans_ to catch up to the span assigned to this
+    // token. We're only interested in empty spans (empty and void elements)
+    for (auto span_it = prev_span + 1; span_it < *target_span_it; span_it++) {
+      // We're only interested in empty spans between the spans in target_span_it
+      if (span_it->size() != 0) continue;
+
+      DiffTags(prev_span->tags, span_it->tags, opening, closing);
+
+      for (auto cit = closing.crbegin(); cit != closing.crend(); ++cit) {
+        std::string close_tag = format("</{}>", (*cit)->name);
```
jerinphilip, line 344: `fmt::format` is available via spdlog, although a bit old. This is what is used by marian's `LOG(fmtstring, Args...)`. Are there speed implications for maintaining a new one, if not let's reuse that one? I've tried getting in the updated one, but ends up violating ODR without a trickle down.

jerinphilip (commented):
> I'm waiting with marking this one ready for #279 to be merged.

Given #279 is merged, taking a quick pass. I'm still mostly relying on black-box testing via https://github.com/jerinphilip/tagtransfer/pull/1 and some programmatic validation of generated HTML than line-by-line source review. Can you please synchronize with main so this branch can be picked up by `tagtransfer` and I can try to throw a few more HTML fragments in an attempt to break stuff ^_^? 

> For each source token we figure out which span they fall in, and then associate that span (and its markup) with the linked target token.

What's the complexity here? Are we at a stage where we can measure what sort of additional time per request the HTML pipeline adds on top of raw translation? This should be better than JS originally intended given the feature had to be done anyway and is now written in C++.

> `<script>` and `<style>` tags will be empty, their contents lost.

Is this temporary? May I query (for understanding, not an immediate resolution) what's making this difficult? Is it not possible to keep information on nodes to copy as-is from source on what eventually becomes `html.Restore(...)`? Asking because I previously saw an unordered set keeping track of tag information somewhere in another PR. 

Tiny `format` related query inline.

src/translator/html.cpp
jelmervdl: It's there mainly for readability because string concatenations are so verbose in C++.

It's a very simple implementation that in itself should be pretty fast. However, it uses std::ostringstream under the hood which is maybe quite the hog. If profiling points that out, we should probably replace these calls with just some raw string operations, and the usage around the exceptions with `fmt::format`.

jelmervdl:
I suggest (or really you suggested ;) ) to move `<style>` and `<script>` treatment to another pull request. Same thing for the elements-should-break-up-words thing I just discussed with @andrenatal. Makes experimenting with it a bit easier :)

jelmervdl:
> What's the complexity here? Are we at a stage where we can measure what sort of additional time per request the HTML pipeline adds on top of raw translation?

The complexity is the same of the previous hard alignment implementation. I don't think I introduced anything more complex.

 > This should be better than JS originally planned given the feature had to be done anyway and is now written in C++.

The slow part of the Javascript implementation was navigating the DOM tree, and maintaining large tables for DOM nodes. Something like that will still be necessary in the browser extension probably, although hopefully those tables will be smaller. All in all the speed up will probably depend a lot on how the new extension will be implemented, and how large the chunks of HTML it sends to the translator will be.

src/translator/html.cpp
jerinphilip: Resolving. Is not strictly related to this PR anyway, but just poked a query given we rushed the first PR.

jerinphilip (approved):
I have a few errors in mobile Wikipedia, which is shared with @jelmervdl offline and is expected to have something to do with script/style tags having trouble.

image

The facebook/twitter icons seem to have trouble at image




This PR also appears to fix problems with image rendering and consequently, web-page render issues I had with Wikipedia elements and such being out of place. 

image

PR #285, opened 2021-12-21
HTML transfer script/style/etc elements

jelmervdl:
This pull requests handles `<script>`, `<style>`, `<!-- -->` and `<? ?>` instructions. These were previously silently ignored, but are now treated as void elements like ``: they are copied to the output based on their position in relation to other elements that do contain text.

Not entirely necessary for block-element-level translation, but nice to have when you're doing full-page translation like our little experiments with Python.

### DIFF
--- a/src/translator/html.cpp
+++ b/src/translator/html.cpp
@@ -47,11 +47,19 @@ size_t countPrefixWhitespaces(string_view const &input) {
   return size;
 }
 
+// Formatters used for exception messages combined with format()
 std::ostream &operator<<(std::ostream &out, HTML::Tag const *tag) {
   if (tag == nullptr) return out << "[nullptr]";
-  out << '<' << tag->name << tag->attributes;
-  if (tag->empty) out << '/';
-  return out << '>';
+  switch (tag->type) {
+    case HTML::Tag::ELEMENT:
+      return out << '<' << tag->name << tag->attributes << '>';
+    case HTML::Tag::VOID_ELEMENT:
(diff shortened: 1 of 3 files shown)


src/translator/html.cpp
```diff
@@ -47,11 +47,19 @@ size_t countPrefixWhitespaces(string_view const &input) {
...
+  switch (tag->type) {
+    case HTML::Tag::ELEMENT:
+      return out << '<' << tag->name << tag->attributes << '>';
+    case HTML::Tag::VOID_ELEMENT:
+      return out << '<' << tag->name << tag->attributes << "/>";
+    case HTML::Tag::COMMENT:
+      return out << "<!--" << tag->data << "-->";
+    case HTML::Tag::PROCESSING_INSTRUCTION:
+      return out << "<?" << tag->data << "?>";
+  }
```
jerinphilip, line 62: There's a compiler warning here on reaching a point of no return, any way to make it happy?

jerinphilip (dismissed):
Having script contents break our python pipeline in complicated ways, but we now know the script content transfer works. :'D 

Minor nit inline.

src/translator/html.cpp
jelmervdl: I would have hoped that the compiler was intelligent enough to see that these are all possible enum values that `tag->type` can have. But I can add a `default: return out << "[Unknown tag]";` It's only used for exception messages anyway.

jerinphilip (approved):

PR #286, opened 2021-12-21
Treat most HTML elements as word-breaking

jelmervdl:
Re discussion with @andrenatal this is a little experiment where inline elements are also treated as spaces. Because fronted webdevelopers do not write semantic HTML.

- [x] Interpret `word<sup><a href="#ref">1</a></sup>` as `word 1` (main: `word1`)
- [x] Interpret `under<u>line</u>d` as `under line d` (main: `underlined`)
- [x] Interpet `under<br>line` as `under\n\nline` where `\n\n` signifies a new sentence (main: `under line`)
- [ ] Test more use cases
  - [x] Test on void elements, e.g. `<br>`
  - [ ] Test on non-void elements, e.g. `<sup>`
  - [ ] Test on non-void elements that are transferred/duplicated (e.g. one word that yields two non-consecutive words in the translation. Need to come up with an example.)

Changes:
- Heuristic to add \n\n to separate sentences from block elements
- Also look at closing tag i.e. `a</div>b` should not become `ab`
- Added `img` to inline tags
- Made `isContinuation` less aggressive in copying markup from a single token to a whole "word"
- Made all above configurable through `HTML::Options` (which is only used by [html-testbench](https://github.com/browsermt/bergamot-translator/tree/html-testbench))

Solves #288 and #302.


src/translator/html.cpp
```diff
@@ -494,11 +504,10 @@ HTML::HTML(std::string &&source, bool process_markup) {
       } break;
 
       case markup::Scanner::TT_TAG_START:
-        // If it makes sense to treat this element as a break in a word (e.g.
-        // <br>, <img>, <li>) make sure it does so in this text as well.
+        // Treat HTML tags always as spaces that break up words.
         // TODO: Strong assumption here that the language uses spaces to
         // separate words
-        if (isBlockElement(scanner.tag()) && !source.empty() && source.back() != ' ') source.push_back(' ');
+        if (!source.empty() && source.back() != ' ') source.push_back(' ');
```
jerinphilip, line 510:
```
$ cat | bergamot translate -m en-es-tiny --html=1 --alignment=1 2> /dev/null
<a href="Tintin">Tintin</a>'s dog is a good dog.
El perro de <a href="Tintin">Tintín</a> es un buen perro.

$ cat | bergamot translate -m en-es-tiny --html=1 --alignment=1 2> /dev/null
<a href="Tintin">Tintin</a> 's dog is a good dog.
<a href="Tintin">El</a> perro <a href="Tintin">de Tintín</a> es un buen perro.
```

jelmervdl:
I'm now adding a marker to tags that introduced a space. But this could be problematic as these tags can be moved around or duplicated. They could move away from the space they introduced, and eat the wrong space(s).

Another way of doing this might be to add a void non-element to the stack, an element that simply renders as 1 negative space. void elements are never duplicated (as non-void elements can be). They can still move around (if pushed forward by a non-void element that occurs earlier in the translation than in the input text) and thus still consume the wrong space…

jerinphilip:
Just trying to raise some concerns and brainstorm here. Can we construct an experiment in which we can automated measure:

1. Is there any degradation to the quality of the translation due to the space introduced? Perhaps BLEU (or friends) on a bunch of entries? Blocker might be ground-truths.
2. Is there any degradation to HTML transferring? Am I correct in plugging the second input in Tintin case where the second HTML transfer assigns tags at two places while the first one assigns it correctly.
3. What percentage of web page nodes have this problem, that devs writing malformed HTML necessitate taking up a complex fix that can do harm as well? Run without this fix on a few pages, run with the fix, how many cases did we get something different. % nodes diff to whole nodes? What portion of this would even matter to disrupt page flow as claimed by extension-devs? 

We can at least do 3, given our python tooling (and if there's a temporary switch within `ResponseOptions`)? Does it make sense to try? I can help out in some capacity.

jelmervdl:
> Is there any degradation to the quality of the translation due to the space introduced? Perhaps BLEU (or friends) on a bunch of entries? Blocker might be ground-truths.

Ideally it would improve the quality of translations because it breaks words apart that should have been broken apart to begin with.

> Is there any degradation to HTML transferring? Am I correct in plugging the second input in Tintin case where the second HTML transfer assigns tags at two places while the first one assigns it correctly.

I'm still investigating this one. I don't think it is related to this change, as currently it doesn't introduce spaces because of closing tags (only opening and void tags. An oversight on my part!) so effectively the input is still `Tintin's` vs `Tintin_'s`. This does seem to cause different alignment scores.

> What percentage of web page nodes have this problem, that devs writing malformed HTML necessitate taking up a complex fix that can do harm as well? Run without this fix on a few pages, run with the fix, how many cases did we get something different. % nodes diff to whole nodes? What portion of this would even matter to disrupt page flow as claimed by extension-devs?

We could do something like that. (We're gonna need a chrome://flags like thing this way 😭) Some of the machinery is definitely necessary as we can assume we will encounter `foobar` like things, where `foobar` is not ment to be a single word. I also need a way to remove that introduced space again, to avoid making it `foo bar` when it comes out. Although, that might be more of a "lab issue" than something that would cause trouble in real world pages.

jelmervdl:
> We can at least do 3, given our python tooling (and if there's a temporary switch via ResponseOptions)? Does it make sense to try? I can help out in some capacity.

I think one way to make this easily testable would be to make the list of elements that introduce spaces temporarily configurable. Then we can experiment with None, All, or a subset more easily.

src/translator/html.cpp
jelmervdl: it seems to be unrelated. Something about the model thinking the `'` aligning best with `El`.

Without space:
without-space

With space: (sorry it's very hard to see but here you can see a bit of space between `</a>` and `'` in the top row)
with-space

jerinphilip: @jelmervdl Is there a source that generates the images above? I'm interested in saving some time at integrating it into an HTML rendering of sorts if it's in python or even C++ something. Thank you for providing the visualization, I have opened #298 about this which I believe could improve the library.

To be honest, I was just engineering a valid counter-example where modifying HTML will have an effect (the comment is not about the exact line if you're looking at it this way). I can also see benefits to this PR's line of thinking. The following fragment used to work well when `tidy` indented the divs (although I think the solution should be fixing the non-semantic HTML by the author). 

```
<div class="person"><a href="https://jerinphilip.github.io/"><img src="people/jerin.jpg" height="150" width="150" alt="Jerin Philip"/><div>Jerin Philip</div></a>Research Assistant</div>
```

`PhilipResearch Assistant` affects translation, once HTML is stripped.

Perhaps we should try selective dropping of spaces (`<div>`, `<sup>` etc) with `libtidy` as a reference which interrupts the flow and is guaranteed to be unaffected by spaces. But I guess whatever `libtidy` is doing as this is based on tags and not requiring stats NLP to resolve ambiguity can and should be the responsibility of the web browser.

jelmervdl: That's what the original implementation did: Everything not `{"abbr", "a", "b", "em", "i", "kbd", "mark", "math", "output", "q", "ruby", "small", "span", "strong", "sub", "sup", "time", "u", "var", "wbr", "ins", "del"}` added whitespace (if it wasn't already there) but those tags didn't.

Note that there was a bug where a closing tag does not add whitespace, regardless of the element.

jelmervdl:
Note that I've force-pushed this branch to just contain HTML processing improvements (especially around adding spaces and keeping markup connected to the right bit of translation). The test tools are in the `html-testbench` branch.

src/translator/html.h
```diff
@@ -18,40 +21,84 @@ class BadHTML : public std::runtime_error {
...
+
+    // List of characters that occur at the start of a token that indicate that
+    // the this token is probably *not* a continuation of a word. Set to empty
+    // to never mark a token as a continuation of the word.
+    // std::string continuationDelimiters = "\n ,.(){}[]";
+    std::string continuationDelimiters;
+
+    // Should we always add spaces to the places where tags used to be? I.e.
+    // `un<u>der</u>line` should become `un der line`?
+    bool substituteInlineTagsWithSpaces = true;
```
jerinphilip, line 45: make this part of response options? I feel defaulting to false is safer. Wouldn't mind making the whole `Options` part of `ResponseOptions`. The experiments you're doing by trial and error can be automated on a volume of HTML by means of crawling. WebAssembly via nodejs still looks primitive for the use-case. No clue about wasm-runtime.

src/translator/html.cpp
```diff
@@ -463,24 +295,41 @@ HTML::HTML(std::string &&source, bool process_markup) {
...
+        // If the previous segment was the open or close tag of a block element
+        // we treat the text after it as a new sentence.
+        if (addSentenceBreak) {
+          if (!(source.empty() || (source.size() > 2 && source.substr(source.size() - 2) == ""))) {
+            stack.push_back(makeTag({Tag::WHITESPACE}));
+            // Important: span->size() == 0 to make it behave as a void element.
+            // Also important: position before the \n\n tokens, not after, to
+            // make it easier to remove them later through apply().
+            spans_.push_back(Span{source.size(), source.size(), stack});
+            source.append("\n\n");  // TODO assumes ssplit-mode = wrapped_text
```
jerinphilip, line 307: split-mode default isn't `wrapped_text`, it's `paragraph`. But I don't think this affects the ultimate outcome when HTML.

jerinphilip (changes requested):
Doing a pass since this is marked as ready for review. 

The web assembly binary needs to go, obviously. 

@jelmervdl What do you think about forking https://mozilla.github.io/translate/ and shifting the (JS) changes there. I think it's best to configure an additional dropdown somewhere in the page where an artefact generated via GitHub CI as part of release here can be picked up (by means of tag, branch, or URL). If this repository is somewhere, perhaps more explanatory visualizations explaining HTML transfer can be added there as a diagnostic tool. 

Not done reviewing, I will try to check a few pages on this via python tooling and review the HTML modifications part.

src/translator/html.h
jerinphilip: Just went and checked test-bench. I believe https://github.com/browsermt/bergamot-translator/commit/5dfeef874422f6dd4e5f4b482de66ef1dea50f53 should be beneficial to more people and a resolution here as well?

app/bergamot.cpp
```diff
@@ -20,6 +20,7 @@ int main(int argc, char *argv[]) {
   std::shared_ptr<TranslationModel> model = service.createCompatibleModel(options, std::move(memoryBundle));
 
   ResponseOptions responseOptions;
+  responseOptions.HTML = config.serviceConfig.HTML;
```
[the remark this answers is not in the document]
jelmervdl: Yeah that's my fault, that's left overs from debug code. If we need html options exposed to the CLI we should do that in a simpler stand-alone pull request :)

(Sorry I'm always confused by the "resolve conversation" button not also adding my comment, which differs from Gitlab)

src/translator/html.cpp
jelmervdl: Hmm, you're right. I've misjudged where that default came from. `\n\n` will still be picked up correctly, but now I'm a bit worried about the quality of [pages that wrap their HTML manually](https://neural.mt) because those `<p>...</p>` will now be interpreted as multiple paragraphs?

src/translator/html.h
kpu: I'm ok with putting the knobs in main even if just there to help development.  Doubt exposing them all the way through the browser is worth it.

jelmervdl: Oh btw I noticed Google Translate also has this behaviour of breaking words at tags, even more aggressively. I suspect they do this for a reason (which was my main reason for "let's go with *on* by default).

It aso doesn't seem to work all that well. For the `<u>d</u><i>o</i><b>g</b>` example it is a solution. But I noticed some of our translation models are really good at fixing mistakes with spaces: Both `transatlantic` and `trans atlantic` are often (correctly) translated as a single word. (Only `trans<br>atlantic` is successful in forcing the split in the translations as well.)

jelmervdl:
> @jelmervdl What do you think about forking https://mozilla.github.io/translate/ (or even opening a PR there with a new webpage as test-bench) and shifting the (JS) changes there.

Not sure. My html-testbench repo solves the "I need a way to see and tweak the knobs", combined with a bit of "I need to explain this to my peers somehow". Mozilla's page seems to be much more targeted to "I need show this is a Google Translate replacement". Moving code from testbench into that would be for the purpose of "I need to explain how this works to a layperson". But that's quite a bit different from my use cases.

For me personally, I much rather just have two branches here, one with the C++ improvements that will at some point make it into main, and one with my little GUI and all the necessary WASM bindings hacked on top of it. Makes it very easy to experiment and pull/pick commits from one another. Easier than trying to keep multiple (public!) repos in sync. And I really don't want to care about the html-testbench code quality or how pretty or explanatory it is.

src/translator/html.h
jerinphilip: 1. Let's not block this on evaluations. We'll decouple it, give it a lower priority and give you more room to experiment.
2. Let's go with the Google default. 

@jelmervdl you know the tooling I have and how the said tooling can make things configurable. If you deem something should be good enough to give a knob outside for python-based things, please do add along with this PR. Since the value from bindings would be 1, we can look into this perhaps when the flags are more mature and you understand it well.

jerinphilip (dismissed):
> Not done reviewing, I will try to check a few pages on this via python tooling and review the HTML modifications part.

The changes appear to be working around the previously undesirable cases (cutting divs, wikipedia `<sup>` tags and heading edit buttons). I hope non-matching tags flush should also be a plus, although I can't find legit places to reproduce.

Tests are failing due to HTML generation being different now. `isspace(...)` might've caused this, if I remember correctly. 

Leaving to @jelmervdl do decide when and where to bring in the Bidnings + JS + HTML.

PR #287, opened 2021-12-21
CI: Circle CI config script update

abhi-agg:
- Robust artifact presence check
 - Variable name refactoring
 - Storing only those artifacts that are required
 - Remove commit sha from the names of the Github Releases
 - Use BERGAMOT_VERSION file contents for Git Tag names

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,13 +19,12 @@ jobs:
           name: Check artifacts
           working_directory: build-wasm
           command: |
-            ls -all bergamot*
-            if ls bergamot*.wasm &>/dev/null && ls bergamot*.js &>/dev/null
-            then
+            ARTIFACT_BASE="bergamot-translator-worker"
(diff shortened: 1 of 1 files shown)


andrenatal:
@abhi-agg did you test this to see if these changes to the artifact sharing between jobs will work?

How we'll identify which version to use if all tags and releases will have the same name? Will you bump the version at every release?

abhi-agg:
> @abhi-agg did you test this to see if these changes to the artifact sharing between jobs will work?

I tested it in my fork of bergamot-translator and it works 👍🏾 

> How we'll identify which version to use if all tags and releases will have the same name? Will you bump the version at every tagging?

Whenever a new tag is required, I will bump the version in [BERGAMOT_VERSION](https://github.com/browsermt/bergamot-translator/blob/main/BERGAMOT_VERSION) file and push it to main via a commit and tag this commit to this version.

PR #289, opened 2021-12-22
vcpkg deprecate temporary ports solution in favour of upstream

jerinphilip:
[WIP] Checking via CI if the temporary workaround can be removed.


XapaJIaMnu:
yeah I have the same issue, but no idea why...

XapaJIaMnu:
I don't have access to your repo, but adding `--debug-find` to the cmake command will tell us where exactly PCRE2 is looked at.

XapaJIaMnu:
We only consider static libs for PCRE2, but for some reason it's looking in `x64-windows` and not in `x64-windows-static` installed directories...

PR #291, opened 2021-12-23
HTML basic integration tests

jerinphilip:
Integration tests for the HTML translation pipeline, which currently only has unit tests. 

Keeping input HTML simple so as to not get in the way of ongoing feature development fixing edge-cases of HTML by means of rule-based corrections.

(https://github.com/jerinphilip/bergamot-translator/pull/28 takes off HTML first and reintegrates it outside through callbacks to bring pivoting up-to-date. Need something to ensure I'm not breaking integration in between).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,13 +19,12 @@ jobs:
           name: Check artifacts
           working_directory: build-wasm
           command: |
-            ls -all bergamot*
-            if ls bergamot*.wasm &>/dev/null && ls bergamot*.js &>/dev/null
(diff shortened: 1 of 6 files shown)


jelmervdl:
The approx-diff path doesn't look like it would work with the HTML output. Other than that, seems like a good addition to the tests.

jelmervdl (approved):
The approx-diff.py path doesn't look like it would be compatible with the HTML output (it's line-based scoring, right?) but other than that this seems like a good addition.

Issue #270, opened 2021-11-30
Setting HTML options towards ease of usage


jerinphilip:
So, there are two options for alignment which HTML is dependent on - the first which decides if a (Marian) model is alignment capable or not decided at model-construction, the second which decides if `Response` - a bergamot primitive should deal with alignments or not effects of which appear at runtime on a translate call.

1. It is safe to assume Mozilla always will need models capable of HTML translation (a few plain-text without `HTML=false` options).
2. What `alignment`: `soft` serves is the purpose of a switch to save some computation on the construction of Alignments. This in Marian is managed by the configuration. Currently, firefox through bindings also manages this through configuration. 
     a. Are there clients who would like to retain an `alignment`: `none` setting? Perhaps `translateLocally`? If not we might default to this always being true and incur any additional computation costs (https://github.com/browsermt/bergamot-translator/pull/269#issuecomment-982877484). 
3. We can read the configuration at runtime at the `translate(...)` call and check if it is capable of providing alignments. This can be used in throwing an error or abort, essentially telling the client "you tried to translate HTML without giving me alignments". We cannot write (set/unset, because we also support a multithreaded setting and the model config is shared among multiple translation requests). 
4. @jelmervdl can have HTML pipeline cry if no alignments. Currently, it provides an illusion that things work if no alignment matrix, we should consider this a bug, given we have something superior to length-based tag transfer which @kpu originally requested. 
5. `ResponseOptions` is a plain-old-datatype that makes things easy for bindings(?). We can change it to a class with constraints on construction (HTML => alignment true always). 
6. We can at call time edit `ResponseOptions.alignment = true if ResponseOptions.HTML = true` as we get it with the translation-request. There will be some duplication for a few upcoming PRs, but this is OK. 

6 seems to be the easiest from a Mozilla UX point of view, w.r.t not bothering about a parameter named `alignment` when all they want is HTML translation. 

2a will improve the at model construction UX. 

Requesting stakeholders to comment, offer any alternate suggestions.

jerinphilip:
There is also a very specific solution for 2a. The string-based constructor is only used by WebAssembly. We can keep a function that always sets the returning option to have `alignment`: `soft`.

https://github.com/browsermt/bergamot-translator/blob/e8fd01e9f4c28d3acdd49485cd4b9395b87aa631/src/translator/translation_model.h#L49-L50

abhi-agg:
[...]

PR #296, opened 2022-01-02
Have alignments placed if HTML is on

jerinphilip:
Fixes one half of #270. 

https://github.com/browsermt/bergamot-translator/pull/271 tries to provide a `sanitizeOptions`, which is perhaps more extensible. But such a function will have to be called at several sites increasing verbosity is my conclusion after experimenting in https://github.com/jerinphilip/bergamot-translator/pull/28.

Instead, we just pick-up `ResponseOptions.HTML` and place alignments with an `or` switch if `HTML` is on, irrespective of whatever value was given for `Alignment`s.

This way a client can go from:

```
$ cat | bergamot translate -m en-es-tiny --html=1 --alignment=1 
<a href="Tintin">Tintin</a>'s dog is a good dog.
El perro de <a href="Tintin">Tintín</a> es un buen perro.
```

to

```
$ cat | bergamot translate -m en-es-tiny --html=1 
<a href="Tintin">Tintin</a>'s dog is a good dog.
El perro de <a href="Tintin">Tintín</a> es un buen perro.
```

(This is tested with python-cmdline).

Should work similarly with `ResponseOptions` for WebAssembly as well.

### DIFF
--- a/src/translator/response_builder.h
+++ b/src/translator/response_builder.h
@@ -61,7 +61,7 @@ class ResponseBuilder {
       buildQualityScores(histories, response);
     }
 
-    if (responseOptions_.alignment) {
+    if (responseOptions_.alignment or responseOptions_.HTML) {
       buildAlignments(histories, response);
     }
     html_.restore(response);
--- a/src/translator/response_builder.h
+++ b/src/translator/response_builder.h
@@ -61,7 +61,7 @@ class ResponseBuilder {
       buildQualityScores(histories, response);
     }
 
-    if (responseOptions_.alignment or responseOptions_.HTML) {
+    if (responseOptions_.alignment || responseOptions_.HTML) {
       buildAlignments(histories, response);
     }
     html_.restore(response);


jelmervdl:
I prefer this solution to #271. It might be worth to nuke response.alignments after HTML has been restored if `options.alignments == false` to reduce the amount of data that crosses the wasm -> page context bridge.

jerinphilip:
This simplifies existing `ResponseOptions` and any future edits can be done without changing external API. Supported behaviour is HTML means alignments will be computed under the hood for HTML. Merging therefore.

Caveat emptor, especially if the user intends on using alignment-matrix through only the HTML options in Response object.

jelmervdl:
> Caveat emptor, especially if the user intends on using alignment-matrix through only the HTML options in Response object.

This is why I would prefer to clear `response.alignments` if `!options.alignments` so that this is only available if `options.alignments == true`, regardless of whether `options.html` is true or false.

PR #297, opened 2022-01-04
Add API to trigger fast shutdown of AsyncService

jelmervdl:
Add a way to AsyncService to shut down without finishing the full queue through `AsyncService::clear()`. The default behaviour is still that `AsyncService::~AsyncService()` will wait for any pending translation requests to finish.

<del>You can either call `AsyncService::terminate()` or `AsyncService::join()` if you want to manually trigger the shutdown of AsyncService. Otherwise its destructor will call `join()` for you. It is safe to call `join()` (or `terminate()` for that matter) multiple times, they effectively become no-ops after the first call.</del>

You can call `AsyncService::clear()` before the destructor is called to make sure there is no work for the service to finish before the workers can stop and join. (Marian batches that are already in progress won't be stopped, we're not trying to cause interrupts in threads or something that complex. But these single batches often don't take that long to complete anyway.)

Todo:
- [x] <del>Add `terminate()` to AsyncService</del>
- [x] <ins>Add `clear()` to AsyncService</ins>
- [x] Add `clear()` to BatchingPool
- [x] Document
- [ ] Add tests
- [x] Implement usage in TranslateLocally: https://github.com/XapaJIaMnu/translateLocally/pull/80


src/translator/batching_pool.cpp
```diff
@@ -59,5 +59,11 @@ size_t BatchingPool::enqueueRequest(Ptr<Request> request) {
   return toBeFreshlyTranslated;
 }
 
+void BatchingPool::clear() {
+  for (size_t length = 0; length < bucket_.size(); length++) {
+    bucket_[length].clear();
```
XapaJIaMnu, line 64: What happens if clear is called while someone wants to consume from the bucket? (I don't know the code here, just asking if it is guaranteed to not happen?)

jelmervdl: BatchingPool is wrapped by ThreadsafeBatchingPool, which has a lock to make sure generateBatch() and clear() are not called concurrently.

jelmervdl:
This API is of little use** to browser extensions since there it is likely that only a couple of translation requests (as opposed to all requests) need to be cancelled. E.g. only a translations for a single tab while the extension serves all tabs. The fast shutdown scenario (waiting for the application to quit because translations are still in progress) also doesn't apply: you can just terminate the WASM vm at any point.

**: Unless you want to give each webpage or each origin its own AsyncService instance with its own cache. Then you can more easily terminate that one without stopping the vm and prevent cache timing attacks at the same time!

XapaJIaMnu:
Is this ready for review?

src/translator/service.cpp
```diff
@@ -61,17 +61,26 @@ AsyncService::AsyncService(const AsyncService::Config &config)
...
-AsyncService::~AsyncService() {
+void AsyncService::terminate() {
+  safeBatchingPool_.clear();
+  join();
+}
+
+void AsyncService::join() {
   safeBatchingPool_.shutdown();
   for (std::thread &worker : workers_) {
     assert(worker.joinable());
```
XapaJIaMnu, line 72: We compile with DNDEBUG, so this doesn't do anything in our release build. Do we assume threads are always joinable (Which I think is reasonable)

XapaJIaMnu: Should we make some mechanism to ensure that this isn't called twice so we don't try to join already joined threads?

src/translator/service.cpp
```diff
@@ -61,17 +61,26 @@ AsyncService::AsyncService(const AsyncService::Config &config)
...
     worker.join();
   }
+  workers_.clear();
 }
 
+AsyncService::~AsyncService() { join(); }
```
XapaJIaMnu, line 78: This feels wrong. Join() and ~AsyncService() both do the same (finishing the translation and destroying all workers). Maybe at least `join()` should be private?

XapaJIaMnu (approved):
Some small comments. Address (or don't) at will.

src/translator/service.cpp
jelmervdl: Last line of `join()` is `workers.clear()` which will make sure that a second call to `join()` won't wait for already joined workers.

There could only be non-joinable threads in there if `join()` didn't clear `workers`, or `workers` somehow contained default-constructed threads. Both would we a programming error, not something that should be able to occur in runtime. I think using `assert()` is fine here (and otherwise it will just be `thread.join()` throwing an `invalid_argument` exception.)

src/translator/service.cpp
jelmervdl: I made it public so you could more easily write something that executes after all requests are processed without relying on the scope of the AsyncService object:
```c++
int main() {
  AsynService service;
  std::cerr << "Service started\n";

  std::thread reader([&]() {
    // reads lines and submits them as requests to service until stdin is empty
  });
  reader.join();
  std::cerr << "All requests submitted\n";

  service.join();
  std::cerr << "All requests handled\n";
  return 0;
}
```

XapaJIaMnu: I guess... I prefer explicitly calling `delete service` at that point. As far as I can tell, once you call join, you can't "resurrect" the service, right?

src/translator/service.cpp
XapaJIaMnu: join doesn't hold a mutex so one could technically call it from two different threads before `workers_.clear()` is called. Do we assume this isn't ever going to happen?

src/translator/service.cpp
jelmervdl: You can use delete if it would be heap-allocated. But if it is stack-allocated the only ways I know are more parentheses or wrapping it in a function to force it to go out of scope.

src/translator/service.cpp
jelmervdl: Hmm, that's right. I would expect any sane program to only ever call join or terminate from the same thread that created (and owns) the AsyncService instance. In that way it's similar to `std::thread::join`: *No synchronization is performed on \*this itself. Concurrently calling join() on the same thread object from multiple threads constitutes a data race that results in undefined behavior.*

src/translator/service.cpp
XapaJIaMnu: oh, yeah. That. True.

jerinphilip (commented):
May I ask what kind-of usages are expected at client code? I'm afraid this might create more paths that will be hard to remove later on if code using it propagates.

Both `join()` and `terminate()` methods render the instance in an unusable state. Shouldn't we just be putting cleanup in the destructor here? All that's additionally needed is a flag that changes a default of finishing up the work items into exiting immediately and let the destructor do the rest of the job?

jelmervdl:
> May I ask what kind-of usages are expected at client code? I'm afraid this might create more paths that will be hard to remove later on if code using it propagates.

- https://github.com/browsermt/bergamot-translator/pull/297#discussion_r782979628
- https://github.com/XapaJIaMnu/translateLocally/pull/80/files#diff-743d75c1134a4d25dad5e54d86b998ef1c8d81fb78df2b0e30fd704e990d5f89R145-R149

> Both join() and terminate() methods render the instance in an unusable state.

Unfortunately yes, but I think this is expected behaviour in the context of it being a (very specific) thread pool. Most of these have a `close()` or something to stop submissions and tell waiting workers it is safe to shut down.

> Shouldn't we just be putting cleanup in the destructor here?

By default it does, but `join()` and `terminate()` are there to control [when the waiting is happening](https://github.com/browsermt/bergamot-translator/pull/297#discussion_r782979628). Without it, you can only wait for finishing through a side-effect of the destructor. If you have a stack-allocated AsyncService, that would mean you need to use syntax tricks to get there. And it is harder to document. So I mirrored a familiar API here.

> All that's additionally needed is a flag that changes a default of finishing up the work items into exiting immediately and let the destructor do the rest of the job?

That would limit you to having to make this decision when AsyncService is created. 

Note that join() and terminate() are both completely optional. You can use them to control when you block. But if you don't use them, the behaviour is still safe and predictable. Again, it mirrors the behaviour of std::thread.

jelmervdl:
Alternative implementation that just adds a `clear()` method to AsyncService: https://github.com/browsermt/bergamot-translator/tree/clearable-asyncservice

After talking with Jerin, I'm starting to prefer that implementation. It allows you to re-use a cleared service: service is not in this weird unusable state, which it would be after join() or terminate() are called.

jerinphilip (approved):
Looks good to me.

PR #319, opened 2022-01-28
Make mozilla models available to python

jerinphilip:
WIP. Not how this should be done ideally, but should be able to automate sanity check of Mozilla models at least those used in extension. 

https://github.com/mozilla/firefox-translations/issues/58


jerinphilip:
Abandoning this. Mozilla models are expected to meet the [translateLocally specification](https://github.com/XapaJIaMnu/translateLocally#importing-custom-models) to harmonize collaboration with translateLocally as well. 

There is a python script in https://gist.github.com/jerinphilip/f9bc31d70e4201771b7734ab1cfd1205 which allows 3rd parties to load from extension JS if we want to consider transforming and using the models in a translateLocally (also python) compatible way. Ideally, Mozilla is expected to keep the model open for easy pull into native testing mechanisms.

Issue #316, opened 2022-01-27
Translation-specific errors should not crash the service


jelmervdl:
From the last call:
> Should we be dying on HTML parse errors and such?  These are query-level errors that should be handled better with an error response.

I proposed to add an `ok` bool to response, similar to `fetch()`'s API: https://developer.mozilla.org/en-US/docs/Web/API/Response/ok

I tried something like this, e.g.:
```cpp
struct Response {
 std::string error;
  bool ok() const { return error.empty(); }
};

// ...

EMSCRIPTEN_BINDINGS(response) {
  class_<Response>("Response")
     // ...
      .property("ok", &Response::ok)
      .property("error", &Response::error);
}
```

But implementing it in such a way that exceptions get caught and their `what()` gets put into the correct `response.error`, especially around the constructor of `HTML`, is a bit more complicated. All methods that interact with the `response` objects now need to check whether it is valid by checking its `ok()`.

Alternative names:
- C++ also uses `good()` (e.g. `iostream::good()`)
- or `operator bool()` of course, but you can't translate that 1-to-1 to Javascript

kpu:
I thought wasm at least was living in `-fno-exceptions` territory where we can't throw and catch at all?  (And even native code stuffed into the browser is compiled `-fno-exceptions`).

jelmervdl:
That is a good point: https://emscripten.org/docs/optimizing/Optimizing-Code.html#c-exceptions

In that case implementing this is a bit more more complicated, since things like the HTML class need to be altered a bit have a separate parse method that can somehow report error messages.

Things like model loading also reference ABORT_IF. Not sure how we should handle that. Preferably also in a recoverable way (no need to crash the whole service if one model failed to load, right?). Then, how do we report the error? Some sort of rust-like `Result<RetType,ErrType>` return value? Or error messages by reference with some extra logic in the emscripten bindings to translate that to something Javascript understands?

kpu:
Some of the `ABORT_IF` we can just keep IMHO.  The biggest one is where input breaks stuff like an edge case in HTML.

kpu:
Now low priority.  This works fine in the C++ API as is, and isn't blocking WASM prototype.

kpu:
To be clear, this was only meant to return an error in WASM when provided ill-formed HTML by the extension.  The extension should never do that because it should properly use text mode or pass `innerHTML` which will be correctly formed.

PR #322, opened 2022-02-01
Response error fields

jelmervdl:
Possible implementation that solves #316. I opened it as draft for feedback.

Problem: WASM bindings compiles Marian and bergamot-translator without exception support.

This pull request removes the exceptions thrown by HTML and replaces them with error return values. It also changes the services implementations a bit to translate these HTML errors into "error responses". 

Todo
- [x] Remove exceptions from HTML
- [x] Let {Blocking,Async}Service return an error response when HTML parsing fails
- [ ] Check for any other recoverable runtime exceptions we should handle this way
- [ ] Update cli utilities that use {Async,Blocking}Service for tests to handle not-ok responses
- [x] Expose response.ok and response.error to WASM
- [x] Look for response.ok in wasm demo page
- [ ] Expose response.ok and response.error to Python

### DIFF
--- a/src/tests/common-impl.cpp
+++ b/src/tests/common-impl.cpp
@@ -9,7 +9,7 @@ Response Bridge<BlockingService>::translate(BlockingService &service, std::share
   // vector<Response>, return.
   std::vector<std::string> sources = {source};
   Response response = std::move(service.translateMultiple(model, std::move(sources), responseOptions).front());
-  ABORT_IF(!response.ok(), "Error in response:", response.error);
+  ABORT_IF(!response.ok(), "Error in response: {}", response.error);
   return response;
 }
 
@@ -25,7 +25,7 @@ Response Bridge<AsyncService>::translate(AsyncService &service, std::shared_ptr<
   responseFuture.wait();
 
   Response response = responseFuture.get();
-  ABORT_IF(!response.ok(), "Error in response:", response.error);
+  ABORT_IF(!response.ok(), "Error in response: {}", response.error);
   return response;
 }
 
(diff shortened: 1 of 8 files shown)


jelmervdl:
This API assumes you'd want other requests to pass even when one request contains bad HTML.

If that is not something we need, I'd suggest we can replace all `BadHTML` exceptions thrown in HTML with `ABORT_IF` macros. These can then be either caught if exceptions are used, or will `abort()` and force downstream to validate their input better. By using `ABORT` we get error messages (if the log level allows for it…) for cases we now don't when compiled with `-fno-exceptions`.

jerinphilip:
The Python tagtransfer mechanism hasn't gotten any `BadHTML` provided I use tidy. From a design standpoint cleaning should be separated from a strict HTML5 or something translator. Right now these concepts are interspersed in the source if I understand correctly - this is not ideal, but can be turned a blind eye towards. Another concern here is that the returns signatures are active on plaintext path as well where it is unsuitable (this is a `BadHTML` only construct, correct?).

> (a) This API assumes you'd want other requests to pass even when one request contains bad HTML.
> (b) If that is not something we need, I'd suggest we can replace all BadHTML exceptions thrown in HTML with ABORT_IF macros.

I believe you have more experience using this in the wild and some instances encountered motivates this PR.  To be able to better discuss the "if" in (b), could you shed some insights into how often there is value in partial translations when a part is malformed and causing whole to fail?  Can you provide a few example pages on the web? Is Firefox's error correction and innerHTML following not enough here?

jelmervdl:
There is currently no need for this after #343.

PR #325, opened 2022-02-02
emscripten: ccache and artefact upload

jerinphilip:
Enables ccache for emscripten. The configuration uses pyiodide for a
reference (https://github.com/pyodide/pyodide/pull/1805).

Two workflows to run on macOS and Ubuntu, reduced to one on Ubuntu. As
emscripten and the target is cross-platform, also macOS runners being
limited - it makes sense to have this removed.

Upload artefact enabled in preparation for a release action to be
scheduled which will upload the bergamot*.wasm and bergamot*.js for
consumption.


jerinphilip:
ccache is tested to be working at https://github.com/jerinphilip/bergamot-translator/pull/31. 
 
The immediate motivation (at least for me) is to get bump the version at which point python packages start getting published with permanent URLs for more permanent notebook (Colab) demonstrations. I'm simplifying existing workflows and enriching them for fast turnarounds, in preparation for a move into a single YAML to sequence a release job. See translateLocally [build.yml](https://github.com/XapaJIaMnu/translateLocally/blob/master/.github/workflows/build.yml) which packages windows, ubuntu and mac packages built through a single job at the end.

The `BERGAMOT_VERSION` file seems to be tied to Mozilla automation from https://github.com/browsermt/bergamot-translator/pull/287, but however, appears to be restricted to the CircleCI Mozilla is using which is left intact by this PR.  There is no proper visibility into Mozilla automation, yet these seem to appear in internal messaging rendering participants clueless. In any case, the ubuntu/mac GitHub workflows as I understand are orphaned and I hope there are no objections to editing this to consolidate release here and suit our needs. I remember something about GitHub limiting storage being a problem in prior exchanges, but we'll solve it when we get to that, I suppose. 

Uploads from the WebAssembly side are eventually expected to aid WebAssembly experimentation (and potentially deployment in websites like https://translate.ikhoefgeen.nl/ if all works out) @jelmervdl is currently doing in https://github.com/browsermt/bergamot-translator/tree/html-testbench which I plan to get involved and assist in at times also making this a convenience for me.

.github/workflows/wasm.yml
```diff
@@ -0,0 +1,129 @@
...
+          else
+            echo "Failure: Artifacts Not Present"
+            exit 1
+          fi
+
+      - name: Upload wasm artifact
+        uses: actions/upload-artifact@v2
+        with:
+          name: bergamot-translator-worker
+          path: ${{github.workspace}}/build-wasm/bergamot-translator-worker.*
```
jelmervdl, line 128: Wouldn't it make sense to have the "Check Artifacts" step check the same paths as the ones you upload? Or better, you can skip the "Check artifacts" step and [make the upload artifacts step fail if no files are found](https://github.com/actions/upload-artifact#customization-if-no-files-are-found).

.github/workflows/wasm.yml
```diff
@@ -0,0 +1,129 @@
...
+
+      - name: Cache-op for build-cache through ccache
+        uses: actions/cache@v2
+        with:
+          path: ${{ env.ccache_dir }}
+          key: ccache-${{ job.id }}-${{ steps.ccache_vars.outputs.hash }}-${{ github.ref }}-${{ steps.ccache_vars.outputs.timestamp }}
+          restore-keys: |-
+            ccache-${{ job.id }}-${{ steps.ccache_vars.outputs.hash }}-${{ github.ref }}
+            ccache-${{ job.id }}-${{ steps.ccache_vars.outputs.hash }}
+            ccache-${{ job.id }}
```
jelmervdl, line 82: I think I would also put the emscripten version number in the key. There are some benefits to bumping it up (but right now we can't because cmake is not configured correctly, and we're passing emscripten flags in all the wrong places which is an error condition since 2.0.16 I think?)

.github/workflows/wasm.yml
```diff
@@ -0,0 +1,129 @@
...
+      - name: Setup Paths
+        run: |
+            # Following two variables are set by emsdk activated otherwise
+            echo "_EMCC_CCACHE=1" >> $GITHUB_ENV
+            echo "EM_CONFIG=$(realpath emsdk/.emscripten)" >> $GITHUB_ENV
+
+            # Configure GitHub Paths
+            echo "${{ github.workspace }}/emsdk" >> $GITHUB_PATH
+            echo "${{ github.workspace }}/emsdk/upstream/emscripten" >> $GITHUB_PATH
+            echo "${{ github.workspace }}/emsdk/node/14.18.2_64bit/bin" >> $GITHUB_PATH
```
jelmervdl, line 46: I expect that path to break at some point when emscripten updates their stuff. In theory node could be one of the tools that makes up an SDK, and so be of a fixed version. But I doubt it is for some reason…

The action I'm using [uses `emsdk construct_env` to print all env variables that the SDK needs, and then exports those selectively to the Github environment](https://github.com/mymindstorm/setup-emsdk/blob/29ba4851d6da084ffdc1e0fc390efadbd75df9d1/src/main.ts#L80-L95).

.github/workflows/wasm.yml
jerinphilip: I need to get it to `$GITHUB_PATH` (manually) or use source script in every shell, I think given the existing setting? I like the idea. The example pointed to is JS for consumption in actions, unsure how to wire the concept in here. Another angle is I can watch pyiodide (which seems to be gaining some traction) and replicate what they're doing.

.github/workflows/wasm.yml
jerinphilip: (oops, mixed up the reply fields):

I need to get it to `$GITHUB_PATH` (manually) or use source script in every shell, I think given the existing setting? I like the idea. The example pointed to is JS for consumption in actions which appears to have access to the environment, unsure how to wire the concept in here. Another angle is I can watch pyiodide (which seems to be gaining some traction) and replicate what they're doing. Let me know if you can point me to a concrete way to solve this, happy to pursue it.

On the plus side, such a failure is easily detectable and fixable at which point we can take a second look.

jerinphilip: I came up with this monstrosity:

```
GITHUB_ENV=github.env
GITHUB_PATH=github.path

rm $GITHUB_ENV $GITHUB_PATH

eval $(./emsdk/emsdk construct_env 2>&1 \
  | grep -v "\+\=" \
  | grep "="  \
  | sed 's/export PATH=\(.*\);/echo \1 >> $GITHUB_PATH;/' \
  | sed 's/export \(.*\);/echo \1 >> $GITHUB_ENV;/' \
  | grep "echo")

tail -v $GITHUB_ENV $GITHUB_PATH
```

I had to additionally add ccache. Unsure how to obtain the entry via `construct_env`. The ccache path looks not pinned to a version, also node-path is now fixed.

.github/workflows/wasm.yml
```diff
@@ -0,0 +1,123 @@
...
+        run: |
+            (cd emsdk && ./emsdk install ${{ env.emsdk_version }} ccache-git-emscripten-64bit)
+            (cd emsdk && ./emsdk activate ${{ env.emsdk_version }} ccache-git-emscripten-64bit)
+            # mtime of this file is checked by ccache, we set it to avoid cache misses.
+            touch -m -d '1 Jan 2021 12:00' emsdk/.emscripten
+
+            # These needs to be done in the activated shell.
+            eval $(./emsdk/emsdk construct_env \
+                | sed 's/export PATH=\(.*\);/echo \1 >> $GITHUB_PATH;/' \
+                | sed 's/export \(.*\);/echo \1 >> $GITHUB_ENV;/' );
```
jelmervdl, line 68: Why does this need to be wrapped in `eval $(…)`?

jerinphilip:
```bash
#!/bin/bash
# script.sh

GITHUB_ENV=github.env
GITHUB_PATH=github.path

rm $GITHUB_ENV $GITHUB_PATH

./emsdk/emsdk activate ccache-git-emscripten-64bit 2.0.9 &> /dev/null || echo "Failed activation"

$1 $(./emsdk/emsdk construct_env  \
  | sed 's/export PATH=\(.*\);/echo \1 >> $GITHUB_PATH;/' \
  | sed 's/export \(.*\);/echo \1 >> $GITHUB_ENV;/')
```

```bash
$ bash script.sh 
tail: cannot open 'github.env' for reading: No such file or directory
tail: cannot open 'github.path' for reading: No such file or directory

$ bash script.sh eval
==> github.env <==
EMSDK=/home/jerin/versioned/sandbox/emsdk
EM_CONFIG=/home/jerin/versioned/sandbox/emsdk/.emscripten
EMSDK_NODE=/home/jerin/versioned/sandbox/emsdk/node/14.18.2_64bit/bin/node

==> github.path <==
/home/jerin/versioned/sandbox/emsdk:/home/jerin/versioned/sandbox/emsdk/upstream/emscripten:/home/jerin/versioned/sandbox/emsdk/node/14.18.2_64bit/bin:/home/jerin/.local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/usr/lib/wsl/lib:/mnt/c/Program Files/Alacritty/:/mnt/c/WINDOWS/system32:/mnt/c/WINDOWS:/mnt/c/WINDOWS/System32/Wbem:/mnt/c/WINDOWS/System32/WindowsPowerShell/v1.0/:/mnt/c/WINDOWS/System32/OpenSSH/:/mnt/c/Program Files/CMake/bin:/mnt/c/Users/jerin/AppData/Local/Microsoft/WindowsApps:/mnt/c/Users/jerin/AppData/Local/Programs/Microsoft VS Code/bin:/mnt/c/Program Files/Git/cmd:/snap/bin
```

PR #329, opened 2022-02-02
Consolidate release artefacts

jerinphilip:
Brings in the previously wasm.yml into python.yml and the resulting file is
renamed as build.yml.

python.yml already had a version and pre-release jobs. These jobs
downloaded artefacts from prior ran jobs (python wheel builds). The
newly attached emscripten build now uploads artefacts of a WebAssembly
binary and javascript files which are fed into the release and pre-release 
jobs in addition to the existing python build.

image


jerinphilip:
The version part and latest part are tested at fork: 

- https://github.com/jerinphilip/bergamot-translator/releases/tag/v0.4.1
- https://github.com/jerinphilip/bergamot-translator/releases/tag/latest

Respectively, with some GitHub Actions first time stupidity in the process 🤦.

.github/workflows/build.yml
```diff
@@ -230,11 +231,117 @@ jobs:
...
+          run: |
+            ccache -s # Print current cache stats
+
+        - name: Instantiate simd wormhole
+          working-directory: build-wasm
+          run: bash ../wasm/patch-artifacts-enable-wormhole.sh
+
+        - name: Import GEMM library from a separate wasm module
+          working-directory: build-wasm
+          run: bash ../wasm/patch-artifacts-import-gemm-module.sh
```
jerinphilip, line 328: It appears the patch is done here but is not enough. 

[CircleCI](https://github.com/browsermt/bergamot-translator/blob/main/.circleci/config.yml) appears to be doing two separate builds. Unsure what exactly is going on. I can't get the built js/wasm to work with chrome but it appears to be working in Firefox (Nightly).

.github/workflows/build.yml
```diff
@@ -230,11 +231,117 @@ jobs:
...
+          run: |
+              emcc --version
+              emcmake cmake --version
+              emmake make --version
+
+        - name: Configure builds
+          run: |
+            mkdir -p build-wasm
+            cd build-wasm
+            emcmake cmake -DCOMPILE_WASM=on ..
```
jerinphilip, line 306: -DWORMHOLE=on, per https://github.com/browsermt/bergamot-translator#-compile

jerinphilip (commented):
Hm.. Do I want this js/wasm to work with browsers other than nightly?

jerinphilip:
Both wormhole/non-wormhole are enabled for now and tested to be working at my fork for both main -> latest and tagged releases. 

* https://github.com/jerinphilip/bergamot-translator/releases/tag/v0.4.2
* https://github.com/jerinphilip/bergamot-translator/releases/tag/latest

Issue #332, opened 2022-02-05
Make default throwExceptionOnAbort


jerinphilip:
Marian is configured to exit on abort. translateLocally switches to `throwOnAbort...`. This is already mentioned in https://github.com/browsermt/bergamot-translator/issues/221, but this issue is just meant to track this specifically.

https://github.com/XapaJIaMnu/translateLocally/blob/6f40bceedb4909dfead81441cfb606582c43dcaf/src/main.cpp#L13

```cpp
// Set marian to throw exceptions instead of std::abort()
marian::setThrowExceptionOnAbort(true);
```

From experience, the python tooling should be able to gracefully do exceptions. I instead get random crashes on Colab. I should be able to get meaningful errors instead.

```
(env) [jerin@eltbk hf-issue-15109]$ ls -l
total 1777300
-rw-rw---- 1 jerin users       423 Dec  3 22:11 config-bt.mk
-rw-rw-r-- 1 jerin users       305 Dec  8 09:14 decoder.yml
-rw-r----- 1 jerin users     18652 Dec  8 09:14 LICENSE
-rw-r--r-- 1 jerin users 874629520 Dec 10 10:24 opusTCv20210807+bt-2021-12-08.zip
-rw-rw-r-- 1 jerin users 942581890 Dec  7 03:48 opusTCv20210807+bt.spm32k-spm32k.transformer-big.model1.npz.best-perplexity.npz
-rw-rw-r-- 1 jerin users      2121 Dec  8 09:14 opusTCv20210807+bt.spm32k-spm32k.transformer-big.model1.npz.best-perplexity.yml
-rw-rw-r-- 1 jerin users     66772 Dec  7 20:04 opusTCv20210807+bt.spm32k-spm32k.transformer-big.train1.log
-rw-rw-r-- 1 jerin users      3926 Dec  7 20:04 opusTCv20210807+bt.spm32k-spm32k.transformer-big.valid1.log
-rw-rw---- 1 jerin users    993770 Nov  3 22:11 opusTCv20210807+bt.spm32k-spm32k.vocab.yml
-rwxrwx--- 1 jerin users        80 Dec  8 09:14 postprocess.sh
-rwxrwx--x 1 jerin users      1413 Dec  8 09:14 preprocess.sh
-rw-rw---- 1 jerin users      1610 Dec  8 09:14 README.md
-rw-rw---- 1 jerin users    831865 Dec  8 09:14 source.spm
[...]

PR #333, opened 2022-02-05
Make default throw exception on abort for python

jerinphilip:
As a library, this is the more sane default. This also allows the
mechanism to convert into runtime errors in python, providing
informative messages to the user via pybind11 existing tooling.

### DIFF
--- a/bindings/python/bergamot.cpp
+++ b/bindings/python/bergamot.cpp
@@ -33,7 +33,10 @@ PYBIND11_MAKE_OPAQUE(Alignments);
 
 class ServicePyAdapter {
  public:
-  ServicePyAdapter(const Service::Config &config) : service_(make_service(config)) {}
+  ServicePyAdapter(const Service::Config &config) : service_(make_service(config)) {
+    // Set marian to throw exceptions instead of std::abort()
+    marian::setThrowExceptionOnAbort(true);
+  }
 
   std::shared_ptr<_Model> modelFromConfig(const std::string &config) {
     auto parsedConfig = marian::bergamot::parseOptionsFromString(config);
--- a/src/translator/service.cpp
+++ b/src/translator/service.cpp
@@ -37,10 +37,7 @@ BlockingService::BlockingService(const BlockingService::Config &config)
       requestId_(0),
       batchingPool_(),
       cache_(config.cacheSize, /*mutexBuckets=*/1),
-      logger_(config.logger) {
-  // Set marian to throw exceptions instead of std::abort()
-  marian::setThrowExceptionOnAbort(true);
-}
+      logger_(config.logger) {}
 
 std::vector<Response> BlockingService::translateMultiple(std::shared_ptr<TranslationModel> translationModel,
                                                          std::vector<std::string> &&sources,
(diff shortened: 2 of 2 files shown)


jerinphilip:
Works for python:
```
(env) [jerin@eltbk hf-issue-15109]$ python3 run.py decoder.yml 
Traceback (most recent call last):
  File "/home/jerin/sandbox/hf-issue-15109/run.py", line 16, in <module>
    model = service.modelFromConfigPath(bergamot_config_path)
RuntimeError: The file of binary model should end with .bin

```

kpu:
Won't this break most error reporting for webassembly?

jerinphilip:
> Won't this break most error reporting for webassembly?

Unsure what Webssembly is doing with exceptions exactly. Easy test for me is to get the artefacts uploaded here - I have automated this via GitHub and use it to poke and see what's happening. I will also try to look into the marian-dev source meanwhile. 

@abhi-agg Could you please take a look and let know how this interacts with WebAssembly workarounds for exceptions?

kpu:
We know what happens in WebAssembly.  It dies with aborted() and no useful error message.  This PR will make things harder to debug.

jerinphilip:
> This PR will make things harder to debug.

It is possible to unset this for `BlockingService`, toggle at a WASM Bindings constructor, or just move the thing to python bindings (since the consumer is python). translateLocally is doing it independently anyway. 

If there are serious conflicts, I can move this to the python adaptation constructor. My use case is pretty much python.

jelmervdl:
I'd be in favour of this if you find a way to make it the default for all builds except for the WASM build, because of the issue Kenneth mentioned. As far as I know there is no way to print the exception's error message when effectively `-fno-exceptions` is used during compilation :(

The WASM build would also benefit from having the default logging level set to `critical` instead of `off` so that error messages from `ABORT_*` macros are actually printed to the javascript console. So having some special init code for that build would gives us (and WASM users) more benefits.

If that's all there, then `ABORT_*` is probably the better way of throwing exceptions, also in the HTML class, which throws its own exceptions right now. For non-WASM it would function the same (but with formatting, yay!) and for WASM it would not fail silently! Only downside would be that it always throws a marian::MarianRuntimeException so it's a bit harder to catch specific classes of exceptions.

kpu:
In general the default could be exceptions if exception are on and abort if exceptions are off.  Getting this to work was a bit painful but we've already done it:

https://github.com/kpu/intgemm/blob/1b8cbd6f611c21011325cfe0312940f0635dea33/intgemm/aligned.h#L27-L31

jerinphilip:
> I'd be in favour of this if you find a way to make it the default for all builds except for the WASM build, because of the issue Kenneth mentioned. 

I have https://github.com/jerinphilip/lemonade/wiki/Setting-Up-iBus#i-dont-have-control-over-ibus-no-idea-what-is-happening at lemonade where exceptions could help as well.

> The WASM build would also benefit from having the default logging level set to critical instead of off so that error messages from ABORT_* macros are actually printed to the javascript console. So having some special init code for that build would give us (and WASM users) more benefits.

I have already mentioned this before - modifications to logging and perhaps the abort can be achieved by means of a factory-method instead of relying on the constructor at: https://github.com/browsermt/bergamot-translator/blob/b1e5a48f1a3e3a6d8a6d553dafe32d881eb3aabe/wasm/bindings/service_bindings.cpp#L78-L84

There is also https://emscripten.org/docs/porting/exceptions.html. Human developer hours are the bottleneck, making it faster in production can perhaps be done at a later stage feels to me like a wiser choice. It appears to be available in Nightly. In general, I prefer weird platform ifdefs to not be present simplifying the mental model of the code here.

I'm merging the current PR for ease of usage ahead at Python where I expect it to have immediate benefits. Request to please continue the discussion at https://github.com/browsermt/bergamot-translator/issues/332. I will try to bring about the WebAssembly related issues after reading online resources into a new issue and link to discussion here in addition.

kpu:
And I have just reverted it.  We had a solution that doesn't break things, don't cowboy merge breaking changes like that.

jerinphilip:
> We had a solution that doesn't break things, don't cowboy merge breaking changes like that.

I didn't break anything, the changes are restricted to the python cpp.

PR #342, opened 2022-02-08
Add ability to load `.npz` models 

### LINKED ISSUE #341: Support model-loading from .npz files
https://github.com/XapaJIaMnu/translateLocally/issues/89

Marian supports this path. It may be possible for us to support this - but this shouldn't be at the cost of making `TranslationModel`complicated. Current path is:

```cpp

// (1) For WebAssembly: string config, legal MemoryBundle, missing entires corresponding to bundle. Thin.
TranslationModel(const std::string& config, MemoryBundle&& memory, size_t replicas = 1)
  : TranslationModel(parseOptionsFromString(config, /*validate=*/false), std::move(memory), replicas){};
  
// (2) Constructor converting options based load into a fast-load by loading MemoryBundle from Options
TranslationModel(const Config& options, size_t replicas = 1)
  : TranslationModel(options, getMemoryBundleFromConfig(options), replicas) {}

// (3) The constructor which does the heavylifting. Chooses mmap based fastpath if memory is non-empty. 
// Falls back to options based FS loads one otherwise.
TranslationModel(const Config& options, MemoryBundle&& memory, size_t replicas = 1);
```

We have already expressed intention to clean a bit of `MemoryBundle` (affecting `getMemoryBundleFromConfig`) in #338. If we can find a way to (a) not load into MemoryBundle from `.npz`, or alternatively, (b) load into `MemoryBundle` from `.npz` we may be able to achieve what we want.

We'd prefer (b) for when #257 is solved?

Marian checks based on extension  https://github.com/marian-nmt/marian-dev/blob/b29cc07a95f49df7825f3a92e860bd642db0e812/src/common/io.cpp#L13-L21), which we can use.

@XapaJIaMnu sounds like the sketch of a plan?

jerinphilip:
Changes `ABORT` on non `.bin` model to an additional check for a `.npz` 
extension. If `.bin`, the fast load path is activated by returning `AlignedMemory`. 
Otherwise, the return of empty `AlignedMemory` causes fallback to
filesystem-based loads.

BRT: A test that checks if translation using `.npz` is approximately similar to 
that of default CLI translation is checked in to ensure stability going ahead.

Previously, we only supported `.bin` models' loading via a fast mmap 
path. While we had the underlying capability to load non `.bin` models, this 
was not exposed, encouraging fast loads. Loading `.npz` models are helpful 
for quick debugging and broader coverage of models available, which will 
enhance user experience at translateLocally and python bindings. 

Fixes #341.
See also: XapaJIaMnu/translateLocally#89

### DIFF
--- a/src/translator/byte_array_util.cpp
+++ b/src/translator/byte_array_util.cpp
@@ -1,4 +1,5 @@
 #include "byte_array_util.h"
+#include "common/io.h"
 
 #include <cstdlib>
 #include <memory>
@@ -93,10 +94,15 @@ AlignedMemory loadFileToMemory(const std::string& path, size_t alignment) {
 AlignedMemory getModelMemoryFromConfig(marian::Ptr<marian::Options> options) {
   auto models = options->get<std::vector<std::string>>("models");
   ABORT_IF(models.size() != 1, "Loading multiple binary models is not supported for now as it is not necessary.");
-  marian::filesystem::Path modelPath(models[0]);
-  ABORT_IF(modelPath.extension() != marian::filesystem::Path(".bin"), "The file of binary model should end with .bin");
-  AlignedMemory alignedMemory = loadFileToMemory(models[0], 256);
-  return alignedMemory;
+  if (marian::io::isBin(models[0])) {
+    AlignedMemory alignedMemory = loadFileToMemory(models[0], 256);
+    return alignedMemory;
+  } else if (marian::io::isNpz(models[0])) {
+    return AlignedMemory();
+  } else {
+    ABORT("Unknown extension for model: {}", models[0]);
+  }
+  return AlignedMemory();
 }
 
 AlignedMemory getShortlistMemoryFromConfig(marian::Ptr<marian::Options> options) {
--- a/src/translator/byte_array_util.cpp
+++ b/src/translator/byte_array_util.cpp
@@ -1,9 +1,9 @@
 #include "byte_array_util.h"
-#include "common/io.h"
 
 #include <cstdlib>
 #include <memory>
 
+#include "common/io.h"
 #include "data/shortlist.h"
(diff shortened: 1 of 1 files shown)


jerinphilip:
Using `csen.base` from https://github.com/browsermt/students/blob/master/csen/download-models.sh#L9-L10. 

<details>
<summary>decoder.yml </summary>

```yaml
relative-paths: true
models:
  - model.npz
vocabs:
  - vocab.csen.spm
  - vocab.csen.spm
beam-size: 6
normalize: 1
word-penalty: 0
mini-batch: 1
maxi-batch: 1
maxi-batch-sort: src
```

</details>

<details>
<summary>run.py</summary>

```py
from bergamot import ResponseOptions, Service, ServiceConfig, VectorString
from bergamot.utils import patch_marian_for_bergamot
import os
import sys

if __name__ == "__main__":
    config = ServiceConfig(numWorkers=1, logLevel="off")
    service = Service(config)

    marian_config_path = sys.argv[1]
    bergamot_config_path = sys.argv[1] + ".bergamot.yml"
    patch_marian_for_bergamot(
        marian_config_path=marian_config_path, bergamot_config_path=bergamot_config_path
    )

    model = service.modelFromConfigPath(bergamot_config_path)

    options = ResponseOptions(alignment=True, qualityScores=True, HTML=False)

    source = "Rychlá hnědá liška přeskakuje líného psa."
    responses = service.translate(model, VectorString([source]), options)
    for response in responses:
        print(response.target.text)
```

</details>

Output: 
```
(env) [jerin@eltbk csen.student.base]$ python3 run.py decoder.yml 
A quick brown fox leaps over a lazy dog.
```

XapaJIaMnu (dismissed):
Merge when you are convinced it works.

PR #346, opened 2022-02-10
Allow per-input options

### LINKED ISSUE #345: HTML/Non-HTML translation per item of Batch Translation API
Based on the discussions in https://github.com/mozilla/firefox-translations/issues/51, the translation API will require some changes to be able to specify whether each entry in the batch of text to be translated is HTML or not. This could mean (as @jelmervdl also pointed in https://github.com/mozilla/firefox-translations/issues/51#issuecomment-1024989397) accepting a `ResponseOptions` object per entry of the input batch.

Opening this issue so that it can be discussed here and appropriate changes can be done. Taking [BlockingService::translateMultiple](https://github.com/browsermt/bergamot-translator/blob/main/src/translator/service.h#L62-L63) API as example, the new API could look like:

```
std::vector<Response> translateMultiple(std::shared_ptr<TranslationModel> translationModel,
                                          std::vector<std::string> &&source, const std::vector<ResponseOptions> &responseOptions);
```

Please note that this means all [ResponseOptions](https://github.com/browsermt/bergamot-translator/blob/main/src/translator/response_options.h#L18-L30) could be different for each entry of `source` and therefore, each `Response` will have to be constructed accordingly.

cc @jerinphilip @kpu @andrenatal

jerinphilip:
Changes signature of `BlockingService::{translate,pivot}Multiple` 
functions to take per input options, so a mix of HTML and plaintext
can be sent from the extension. Templating over testing is adjusted 
to allow for continuous evaluations by modifying the test code. 

Updates WebAssembly bindings to reflect the change in signature 
and the javascript test-page to work with the new bindings.

This change lacks an accompanying test specific to the mixed HTML 
and plaintext inputs.

Fixes: #345 
See also: https://github.com/mozilla/firefox-translations/issues/94

### DIFF
--- a/src/tests/wasm.cpp
+++ b/src/tests/wasm.cpp
@@ -2,21 +2,32 @@
 using namespace marian::bergamot;
 
 void wasm(BlockingService &service, std::shared_ptr<TranslationModel> &model) {
-  std::vector<ResponseOptions> responseOptions;
+  std::vector<std::string> plainTexts = {
+      "Hello World!",                                 //
+      "The quick brown fox jumps over the lazy dog."  //
+  };
+
+  std::vector<std::string> htmlTexts = {
+      "<a href=\"#\">Hello</a> world.",                                                   //
+      "The quick brown <b id=\"fox\">fox</b> jumps over the lazy <i id=\"dog\">dog</i>."  //
+  };
+
   std::vector<std::string> texts;
-
-  // WASM always requires HTML and alignment.
-  // TODO(jerinphilip): Fix this, bring in actual tests.
-  // responseOptions.HTML = true;
-  // responseOptions.alignment = true;  // Necessary for HTML
-
-  // Hide the translateMultiple operation
-  for (std::string line; std::getline(std::cin, line);) {
-    texts.emplace_back(line);
-    responseOptions.emplace_back();
+  std::vector<ResponseOptions> options;
+  for (auto &plainText : plainTexts) {
+    texts.push_back(plainText);
+    ResponseOptions opt;
+    options.push_back(opt);
   }
 
-  auto results = service.translateMultiple(model, std::move(texts), responseOptions);
+  for (auto &htmlText : htmlTexts) {
+    texts.push_back(htmlText);
+    ResponseOptions opt;
+    opt.HTML = true;
+    options.push_back(opt);
+  }
+
+  auto results = service.translateMultiple(model, std::move(texts), options);
 
   for (auto &result : results) {
     std::cout << result.getTranslatedText() << std::endl;
--- a/src/tests/wasm.cpp
+++ b/src/tests/wasm.cpp
@@ -2,32 +2,21 @@
 using namespace marian::bergamot;
 
 void wasm(BlockingService &service, std::shared_ptr<TranslationModel> &model) {
-  std::vector<std::string> plainTexts = {
-      "Hello World!",                                 //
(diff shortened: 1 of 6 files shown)


src/translator/service.h
```diff
@@ -57,27 +57,30 @@ class BlockingService {
...
+  /// @param [in] responseOptions: ResponseOptions per source-item indicating whether or not to include some member in
+  /// the Response, also specify any additional configurable parameters.
   std::vector<Response> translateMultiple(std::shared_ptr<TranslationModel> translationModel,
-                                          std::vector<std::string> &&source, const ResponseOptions &responseOptions);
+                                          std::vector<std::string> &&source,
+                                          const std::vector<ResponseOptions> &responseOptions);
 
   std::vector<Response> translateMultipleRaw(std::shared_ptr<TranslationModel> translationModel,
-                                             std::vector<std::string> &&source, const ResponseOptions &responseOptions);
+                                             std::vector<std::string> &&source,
```
abhi-agg, line 67: Was `translateMultipleRaw` intended to be a private function and was left public by mistake? If so, can we please make it private?

abhi-agg:
@jerinphilip Do you intend to make a similar change to Async service as well?

src/tests/wasm.cpp
```diff
@@ -2,21 +2,32 @@
...
+    options.push_back(opt);
+  }
 
-  // Hide the translateMultiple operation
-  for (std::string line; std::getline(std::cin, line);) {
-    texts.emplace_back(line);
-    responseOptions.emplace_back();
+  for (auto &htmlText : htmlTexts) {
+    texts.push_back(htmlText);
+    ResponseOptions opt;
```
abhi-agg: @jerinphilip I believe `opt.HTML` needs to be set to `true` here because these are html texts 👍🏾

abhi-agg: > Check [`c5cea4b` (#346)](https://github.com/browsermt/bergamot-translator/pull/346/commits/c5cea4b21170cac793d9f9d93b35fa5687e28dfc) for a quick confirmation this feature works. In the interest of speeding up Mozilla, I'm proposing to merge this at the earliest.
> 

Oh. I see this commit is reverted.

jerinphilip:
> Do you intend to make a similar change to Async service as well?

Not sure how this is relevant to Mozilla efforts, but I will try to answer. AsyncService was meant to operate at one input which allows one option setting, not wait and go forward. Batching will happen by queueing and not blocking. Also figured it's easier to provide cancellation handles per request than for a batch of requests. Maybe we will change it later if we need the concept of a "BatchJob", but not right now. Where such a change relevant is perhaps python, but I don't see an immediate need to do it unless there's a client.

jerinphilip:
@jelmervdl points out (in internal messaging) without `VectorResponseOptions` at WASM bindings the code is broken for use in JS, consequently, JS is broken as well. 

@abhi-agg Please note that I expect you to experiment with and bring changes to `worker.js` and bindings necessary as you would usually do in the verification step.

jerinphilip:
@jelmervdl has been kind enough to edit the WASM/HTML bindings as well. Requesting @abhi-agg to review now.

jelmervdl (approved):

abhi-agg (approved):
> This change lacks an accompanying test specific to the mixed HTML
> and plaintext inputs.

Approving this change per agreement that the corresponding test cases would be added in a separate PR soon.

Issue #317, opened 2022-01-27
Cache improvements


jerinphilip:
We have some feedback about cache usage from translateLocally (https://github.com/XapaJIaMnu/translateLocally/pull/75). 

@jerinphilip expects to take on https://github.com/jerinphilip/tagtransfer/issues/3 at some point to see how useful the cache in a simulated link-surfing browsing workflow, with a bit of back and forth if there is time eventually.

The following was discussed at python merge and will need addressing at some point.

1. Some workloads don't want cache on for the entire duration of work being provided. This can be managed through a Ptr, or more suitably after C++17 - `std::optional`. 
2. Per request/response cache configuration is possible and can be added. However I think it will be just enough to configure at start for whole of request/response. 
3. Certain configuration items are not ideal to be exposed outside, but rather be configured here as a function of other config items (eg: `cache-mutex-buckets`).
4. Defaults driven by the downstream uses that make most sense, for python at least.

There some repetition eg: ternary operator used similar to https://github.com/browsermt/bergamot-translator/blob/c0f311a8c067372057a6f301c42b40bbe30a9c1a/src/translator/service.cpp#L65 in several places can be changed into an inline function.

PR #347, opened 2022-02-12
Improve cache

jerinphilip:
Hide `cache-mutex-buckets` from the user. Now configured to be equal to number
of workers. Python bindings which had exposed these are modified to reflect
the API change. `std::optional` enabled on cache, constructed only if enabled.
Pointers used are replaced with an equivalent `std::optional.`

Fixes: #317

### DIFF
--- a/src/translator/service.cpp
+++ b/src/translator/service.cpp
@@ -30,7 +30,7 @@ Response combine(Response &&first, Response &&second) {
   return combined;
 }
 
-std::optional<TranslationCache> make_optional_cache(bool enabled, size_t size, size_t mutexBuckets) {
+std::optional<TranslationCache> makeOptionalCache(bool enabled, size_t size, size_t mutexBuckets) {
   return enabled ? std::make_optional<TranslationCache>(size, mutexBuckets) : std::nullopt;
 }
 
@@ -40,7 +40,7 @@ BlockingService::BlockingService(const BlockingService::Config &config)
     : config_(config),
       requestId_(0),
       batchingPool_(),
-      cache_(make_optional_cache(config.cacheEnabled, config.cacheSize, /*mutexBuckets = */ 1)),
+      cache_(makeOptionalCache(config.cacheEnabled, config.cacheSize, /*mutexBuckets = */ 1)),
       logger_(config.logger) {}
 
 std::vector<Response> BlockingService::translateMultiple(std::shared_ptr<TranslationModel> translationModel,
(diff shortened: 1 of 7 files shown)


src/translator/service.h
```diff
@@ -110,18 +110,13 @@ class AsyncService {
     bool cacheEnabled{false};  ///< Whether to enable cache or not.
     size_t cacheSize{2000};    ///< Size in History items to be stored in the cache. Loosely corresponds to sentences to
```
jelmervdl, line 111: I doubt this comment. This is like theoretical max number of sentences. The chance 2000 sentences all have a different `hash % 2000` is pretty slim. In TranslateLocally I set it to 65536, and on Crime & Punishment it has about 1600 collisions on 15000 sentences. So occupancy is about 22%? Or see my earlier not very scientific exploration, https://github.com/XapaJIaMnu/translateLocally/pull/75#issuecomment-1004226518 

Edit: I have more of a problem with it in the description of the `--cache-size` option.

jelmervdl (approved):
Tested it in translateLocally, and works as advertised.

src/translator/service.h
XapaJIaMnu: We need a better hash function..... murmurhash perhpas

jelmervdl: That was my intuition as well. I tested it with wrapping the vocab offset size_t with `std::hash` to spread the bytes a bit and hopefully get fewer colissions, but it didn't have a real impact on the numbers I just mentioned. No idea whether that is because of the input I tried, or inherently the limit of how much of the cache can be reached by `N(0..vocab-size) % 65536` while being mixed up through a couple of hash combine functions, or something completely different. I think this (this seems to be my favourite closing line this evening) needs its own issue to explore it a bit further.

(I tested both std::hash<size_t> and murmurhash a couple of months ago when I was working on landing this in TranslateLocally. I didn't find any improvements back then, otherwise I would have opened an issue about it.)

jelmervdl: My problem here with this comment though is that 2000 seems to be some theoretical limit that is never reached. I'm afraid that downstream will read this and think "yeah no we're never going to have more than 2000 sentences that occur multiple times and are worth caching" and they end up with like 450 sentences in a cache and no guarantee that the more often occurring sentences are not constantly pushed out of the cache by collisions.

XapaJIaMnu: We should definitely open an issue lest this goes forgotten.

jerinphilip: So there are two things here:

1. What does this parameter control: Strong correspondence is to History items. Weak correspondence is the number of sentences to cache. That's all that's meant by the comment. I see that random replacement will lead to troubling interpretations. We can expand the comment to be clearer about the nuances.
2. Reasonable default: 2000 vs 65536 - I haven't been able to experiment, I kind of wanted to expose cache hit/misses from Service and get this from somewhere. https://github.com/jerinphilip/tagtransfer/issues/3 is kept for the same.

src/translator/service.cpp
```diff
@@ -30,13 +30,17 @@ Response combine(Response &&first, Response &&second) {
   return combined;
 }
 
+std::optional<TranslationCache> make_optional_cache(bool enabled, size_t size, size_t mutexBuckets) {
```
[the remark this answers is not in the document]
jerinphilip: Will fix this in this PR itself.

PR #348, opened 2022-02-14
JS: Passing ResponseOptions for every item of translation batch API

abhi-agg:
Updated wasm test page:
 - Now translate() JS API accepts ResponseOptions per batch item (courtesy of https://github.com/browsermt/bergamot-translator/pull/346)
 - Fixed the logic to create `vector<ResponseOption>` that was introduced in https://github.com/browsermt/bergamot-translator/pull/346 and broke the test page


abhi-agg:
Merging this one as it is a change in test page that doesn't impact anyone.

PR #357, opened 2022-02-17
Store subword indices for quality scores in Response

### LINKED ISSUE #355: QE for HTML input doesn't work
A quick experiment with wasm test page for html text translation shows weird byte ranges for words.

An example:

```
Sentence: "How are  <b>you</b>  doing?"
Translated sentence: "Wie geht <b>es dir</b>?"

"words":["Wie", "geht", "<b", "es d"]
"wordByteRanges":[{"begin":0,"end":3}, {"begin":4,"end":8}, {"begin":9,"end":11}, {"begin":12,"end":16}]

```

Looks like, QE ignores tags completely (treating them as if they are non-existent) in the translated sentences and compute the byte ranges of the words in the translated sentences. Is it something that needs to be fixed or am I doing something wrong at my end?

Attaching the image for detailed results.

Screenshot 2022-02-17 at 14 32 35

cc @kpu @jerinphilip @abarbosa94 @felipesantosk @mfomicheva

jelmervdl:
Fixes #355 

This change stores the quality score per subword range instead of byte range in `Response`. When inserting HTML into the translation output, the byte offsets change but the subword offsets don't. 

I added a type for `SubwordRange` (previously just an alias for `ByteRange`) to avoid any confusion between the two. Now you can't compare or interchange them, even though they're both just ranges.

The WASM bindings still produce a SentenceQualityScore that works exactly the same as now, i.e. with ByteRanges. In #355 I mused that Javascript could access the byteranges per word through `response.target.wordAsByteRange()` but neither `response.target` nor the type `AnnotatedText` are currently exposed through the bindings. So to not introduce much more new API at this point, I made `getQualityScores()` return the same thing it always did.

Downside of this approach is that you'll need to `delete()` that returned object in Javascript land. I've changed `worker.js` to do this.

### DIFF
--- a/src/tests/common-impl.cpp
+++ b/src/tests/common-impl.cpp
@@ -159,9 +159,7 @@ void TestSuite<Service>::qualityEstimatorWords(Ptr<TranslationModel> model) {
     const auto &sentenceQualityEstimate = response.qualityScores[sentenceIdx];
     std::cout << "[SentenceBegin]\n";
 
-    for (const auto &wordRange : sentenceQualityEstimate.wordRanges) {
-      const ByteRange wordByteRange{response.target.wordAsByteRange(sentenceIdx, wordRange.begin).begin,
-                                    response.target.wordAsByteRange(sentenceIdx, wordRange.end).begin};
+    for (const auto &wordByteRange : getWordByteRanges(response, sentenceIdx)) {
       const string_view word(response.target.text.data() + wordByteRange.begin, wordByteRange.size());
       std::cout << word << "\n";
     }
--- a/src/tests/common-impl.cpp
+++ b/src/tests/common-impl.cpp
@@ -155,10 +155,13 @@ void TestSuite<Service>::qualityEstimatorWords(Ptr<TranslationModel> model) {
   std::string source = readFromStdin();
   const Response response = bridge_.translate(service_, model, std::move(source), responseOptions);
 
-  for (const auto &sentenceQualityEstimate : response.qualityScores) {
+  for (size_t sentenceIdx = 0; sentenceIdx < response.qualityScores.size(); ++sentenceIdx) {
+    const auto &sentenceQualityEstimate = response.qualityScores[sentenceIdx];
     std::cout << "[SentenceBegin]\n";
 
-    for (const auto &wordByteRange : sentenceQualityEstimate.wordByteRanges) {
+    for (const auto &wordRange : sentenceQualityEstimate.wordRanges) {
+      const ByteRange wordByteRange{response.target.wordAsByteRange(sentenceIdx, wordRange.begin).begin,
+                                    response.target.wordAsByteRange(sentenceIdx, wordRange.end).begin};
       const string_view word(response.target.text.data() + wordByteRange.begin, wordByteRange.size());
       std::cout << word << "\n";
     }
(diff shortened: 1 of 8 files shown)


src/tests/common-impl.cpp
```diff
@@ -155,10 +155,13 @@ void TestSuite<Service>::qualityEstimatorWords(Ptr<TranslationModel> model) {
...
 
-  for (const auto &sentenceQualityEstimate : response.qualityScores) {
+  for (size_t sentenceIdx = 0; sentenceIdx < response.qualityScores.size(); ++sentenceIdx) {
+    const auto &sentenceQualityEstimate = response.qualityScores[sentenceIdx];
     std::cout << "[SentenceBegin]\n";
 
-    for (const auto &wordByteRange : sentenceQualityEstimate.wordByteRanges) {
+    for (const auto &wordRange : sentenceQualityEstimate.wordRanges) {
+      const ByteRange wordByteRange{response.target.wordAsByteRange(sentenceIdx, wordRange.begin).begin,
+                                    response.target.wordAsByteRange(sentenceIdx, wordRange.end).begin};
```
jerinphilip, line 164: Is there a possibility `wordByteRange` ends up including an HTML opening/closing tag if two subwords end up with an HTML token in between?

src/translator/quality_estimator.cpp
```diff
@@ -267,22 +267,4 @@ std::vector<SubwordRange> mapWords(const std::vector<float>& logProbs, const Ann
...
-
-    if (isspace(target.text.at(wordBegin))) {
-      ++wordBegin;
-    }
-
-    words.emplace_back(ByteRange{wordBegin, wordEnd});
-  }
-
-  return words;
-}
```
jerinphilip, line 286: What's happening here? There was a "words as meaningful QE units" notion before, are we losing this with the HTML integrated QE?

jerinphilip (commented):
Leaving a few queries below. Mozilla and QE have negotiated some properties (meaningful units of QE = word, etc.). There was some threshold based operation at extension discussion as well. This was before HTML, the assumptions were plaintext QE - so we may revisit. 

Requesting @mfomicheva, @abarbosa94 and @felipesantosk to take a look, since we're modifying the parts.

src/tests/common-impl.cpp
kpu: Yes.  They are completely unsynchronized and they have to be for better or worse.  I imagine HTML rendering will use something like #358 by copying the QE score tags.

src/translator/quality_estimator.cpp
jelmervdl: The QualityEstimator class still works with this assumption, subwords are grouped into words in `mapWords`: https://github.com/browsermt/bergamot-translator/blob/bce30cfcb3079e3d9cf82ef005815e1ab467517f/src/translator/quality_estimator.cpp#L241-L268

The bit of code I removed here turned those groupings into a single byterange. That has been moved to the WASM bindings bit since that's the last possible moment to do it.

I think most clients will find it useful to do this themselves by accessing the AnnotatedText's data. But since we don't expose those to Javascript, and I didn't want to start a half-assed attempt at reworking the Javascript API, I moved the functionality above to here:
https://github.com/browsermt/bergamot-translator/blob/bce30cfcb3079e3d9cf82ef005815e1ab467517f/wasm/bindings/response_bindings.cpp#L33-L52

jelmervdl:
Closing this in favour of #358 (which also includes part of this pull request)

Issue #350, opened 2022-02-14
Deprecate unnecessary WebAssembly functions in Response


jerinphilip:
The C++ API uses `Response` as data. `source` and `target` are symmetric in structure and this has benefits. 

[Python bindings](https://github.com/browsermt/bergamot-translator/blob/main/bindings/python/bergamot.cpp) almost entirely rely on pybind11 metaprogramming and composition to obtain an easy to maintain, binding friendly value-type.

https://github.com/browsermt/bergamot-translator/blob/a94725b20da9e2f3b1f5303c462d2c6e129c92f9/bindings/python/bergamot.cpp#L166-L170

It is also easier from a maintenance and development perspective that the accessors be consistent across bindings exported.

With a JSON counterpart also coming for Native Messaging (https://github.com/XapaJIaMnu/translateLocally/pull/93), it is important to make the access consistent across Python, C++ and JavaScript, which the developers use.

WebAssembly is also, just bindings. However, WebAssembly appears to planted methods some of which are dead code as of now, creating another means to access the same thing.

https://github.com/browsermt/bergamot-translator/blob/a94725b20da9e2f3b1f5303c462d2c6e129c92f9/src/translator/response.h#L65-L75

https://github.com/browsermt/bergamot-translator/blob/a94725b20da9e2f3b1f5303c462d2c6e129c92f9/wasm/bindings/response_bindings.cpp#L25-L33

This issue proposes to deprecate the WebAssembly added extra functions and constructing bindings via composition to make access to source/target identical and consistent with usages elsewhere. 

The changes proposed in this issue will break the WebAssembly test page, but can be fixed quite easily. Also some stuff at https://github.com/mozilla/firefox-translations/blob/6892ad64164787adf1e284141cbd0d10e71fb40d/extension/controller/translation/translationWorker.js at some point, but assuming worker draws from test-page it's solved already by fixing the test-page. It is easier to incur the cost now than later and have all exports of at least `Response` evolve together. 

I am not aware of performance hits we will incur with the convenience mentioned above, but a consistent API expecting the platform (emscripten/WebAssembly) to fix the performance hits in future is what I propose we should still go for.

jerinphilip:
Replacing `.function` with `.property` providing only a getter (no setter) is a potential way to go - https://emscripten.org/docs/api_reference/bind.h.html#_CPPv4NK6class_8propertyEPKcM9ClassType9FieldType. 

[...]

PR #358, opened 2022-02-17
Embed quality-scores as HTML tag attributes

jelmervdl:
Extended fix for #355.

This is #357 but with the quality scores also exposed as `<font x-bergamot-sentence-score="">` and `<font x-bergamot-word-score="">` tags in the HTML output. It makes the HTML output itself look pretty horrific, but the resulting rendered HTML can easily be styled to show the scores.

Example: `Have you seen my <strong>car</strong> recently? I lost it.`

Italian: `<font x-bergamot-sentence-score="-0.112859"><font x-bergamot-word-score="-0.273196">Hai</font> <font x-bergamot-word-score="-0.0245442">visto</font> <font x-bergamot-word-score="-0.193084">la</font> <font x-bergamot-word-score="-0.00968757">mia</font></font> <strong><font x-bergamot-sentence-score="-0.112859"><font x-bergamot-word-score="-0.131326">auto</font></font></strong> <font x-bergamot-sentence-score="-0.112859"><font x-bergamot-word-score="-0.157974">di</font> <font x-bergamot-word-score="-0.000199795">recente?</font></font> <font x-bergamot-sentence-score="-0.18783"><font x-bergamot-word-score="-0.224375">L&apos;ho</font> <font x-bergamot-word-score="-0.151286">perso.</font></font>`

With Javascript or CSS you could pretty easily have some interface based on those extra attributes. Or we could change the attributes to something like `style="--x-bergamot-word-score: -0.4"` and you can use CSS to directly have it control a background colour or something. But that would be a bit harder to parse with Javascript.

Example:
image


jelmervdl:
I'm using `<font>` here as an element (even though `span` would be the semantically correct one) because I'd expect nobody in their right mind would have added rules to their stylesheet to style those by default. But I'm open for better suggestions. Also regarding the attribute names.

src/translator/html.cpp
```diff
@@ -580,6 +589,54 @@ void HTML::copyTaint(Response const &response, std::vector<std::vector<size_t>>
...
+      for (size_t wordIdx = 0; wordIdx < sentenceQuality.wordRanges.size(); ++wordIdx) {
+        Tag *wordTag = makeTag({Tag::ELEMENT, "font"});
+        wordTag->attributes += format(" x-bergamot-word-score=\"{}\"", sentenceQuality.wordScores[wordIdx]);
+        auto const &range = sentenceQuality.wordRanges[wordIdx];
+        for (size_t tokenIdx = range.begin; tokenIdx < range.end; ++tokenIdx) {
+          targetTokenTags[tagOffset + tokenIdx].push_back(wordTag);
+        }
+      }
+    }
+  }
```
jerinphilip, line 631: What are wordRanges? There are two ways I see this happening for an example of two subwords constituting a word.

(1) `<sw>A</sw></tag><sw>a</sw> -> <w score='s'>A</tag>a</w>` (this is troublesome and can break HTML)
(2)  `<sw>A</sw></tag><sw>a</sw> -> <w score='s'>A</w></tag><w score='s'>a</w>` (this is acceptable if the same value is used across both and rendering (if color coding) appears similar. 

Given `wordIdx` appears to index `sentenceQuality.wordScores` is (1) a possibility?

jelmervdl: No, (1) not possible thanks to how HTML insertion works. It will automatically generate (2).

There is an additional guarantee right now: the added tag will always be the leaf node. So instead of

(A) `<p><w>He<tag>le</tag>boel</w></p>`  you will have
(B) `<p><w>He</w><tag><w>le</w></tag><w>boel</w></p>` (where those `<w>` tags will have the same attributes again).

This is quite useful since now it wont break CSS selectors like `p > tag { text-decoration: underline; }`.

jerinphilip: > not possible thanks to how HTML insertion works. It will automatically generate (2).

This is helpful. Thanks. If we are controlling HTML we will need inputs from Mozilla on what their UI plans are, right now we're guessing. Let's carry over this part in #355.

jelmervdl:
I was experimenting a bit with this. If you're planning on making a popup highlight the full word or sentence on hover, i.e. like Google Translate does, it's difficult to find all `<font>` tags that together form a word or sentence. They have the same score, but there is no unique ID or anything to find them. That's something that can be added though.

It's not an issue if you just want to highlight good or sentences or words, non-interactively. Then just the scores are sufficient information. All the mock-ups I've seen show this type of output.

src/translator/html.cpp
```diff
@@ -31,12 +31,12 @@ void encodeEntities(marian::string_view const &input, std::string &output) {
...
-        break;
-      case '\'':
-        output.append("&apos;");
-        break;
+      // case '"':
+      //   output.append("&quot;");
+      //   break;
+      // case '\'':
+      //   output.append("&apos;");
+      //   break;
```
jerinphilip, line 39: What's happening here? Is HTML subword splitting QE influencing/motivating HTML render?

jelmervdl: I initially started escaping the single and double quote to allow for the text to work with html attributes. But I'm not adding text to attributes any more, so there is no need to support those in the escaping routine.

abhi-agg:
> Extended fix for #355.
> 
> This is #357 but with the quality scores also exposed as `<font x-bergamot-sentence-score="">` and `<font x-bergamot-word-score="">` tags in the HTML output.

@jelmervdl Just confirming as I had missed it before. Does this PR build on top of https://github.com/browsermt/bergamot-translator/pull/357? Meaning, apart from exposing quality scores for "word" byte ranges for html text, this PR additionally provides quality scores embedded into the translated result (using font attribute) as well?

jelmervdl:
@abhi-agg yes, when processing a translation request in HTML mode, the scores are available both in the HTML itself as well as through byte offsets. (And when _not_ in HTML mode byte offsets are the only way they're available.)

abhi-agg:
Thanks for the clarification 👍🏾

wasm/bindings/response_bindings.cpp
```diff
@@ -10,17 +10,38 @@
...
+  scores.reserve(response.qualityScores.size());
+
+  for (size_t sentenceIdx = 0; sentenceIdx < response.qualityScores.size(); ++sentenceIdx) {
+    scores.emplace_back(SentenceQualityScore{response.qualityScores[sentenceIdx].wordScores,
+                                             marian::bergamot::getWordByteRanges(response, sentenceIdx),
+                                             response.qualityScores[sentenceIdx].sentenceScore});
+  }
+
+  return scores;
+}
```
jerinphilip, line 44: All features expected to be happening at the extension by using `ByteRange` instances (words, sentences) on plain-text translations are now using HTML as the delivery mechanism in proof of concepts:

 * https://github.com/jelmervdl/firefox-translations/pull/11 
 * https://github.com/jelmervdl/firefox-translations/issues/5 
 
 Providing an alternate means of access incurs a maintenance overhead and is impractical to be used at the extension due to part tags being present in these `ByteRange`s potentially.

wasm/test_page/js/worker.js
```diff
@@ -326,6 +326,7 @@ const _parseTranslatedTextSentenceQualityScores = (vectorResponse) => {
       sentenceQualityScores.push(sentenceQualityScore);
     }
     result.push(sentenceQualityScores);
+    vectorSentenceQualityScore.delete();
```
jerinphilip, line 329: Can be removed as well?

src/translator/response.h
```diff
@@ -77,6 +77,8 @@ struct Response {
 
 std::vector<Alignment> remapAlignments(const Response &first, const Response &second);
 
+std::vector<ByteRange> getWordByteRanges(Response const &response, size_t sentenceIdx);
```
jerinphilip, line 80: There are no plaintext clients to QE after outbound translation out of scope. Delivery for HTML is via HTML attributes. Can remove this?

src/translator/response.cpp
```diff
@@ -142,4 +142,22 @@ std::vector<Alignment> remapAlignments(const Response &first, const Response &se
...
+
+    if (std::isspace(response.target.text.at(wordBegin))) {
+      ++wordBegin;
+    }
+
+    wordByteRanges.emplace_back(ByteRange{wordBegin, wordEnd});
+  }
+
+  return wordByteRanges;
+}
```
jerinphilip, line 161: I think we require this for tests not to fail. Perhaps move within one of the QE source files or within `common-impl.cpp`.

src/translator/quality_estimator.cpp
```diff
@@ -160,7 +160,7 @@ Response::SentenceQualityScore LogisticRegressorQualityEstimator::computeSentenc
   const float sentenceScore =
       std::accumulate(std::begin(wordScores), std::end(wordScores), float(0.0)) / wordScores.size();
 
-  return {wordScores, subwordToWords(wordIndices, target, sentenceIdx), sentenceScore};
+  return {wordScores, wordIndices, sentenceScore};
```
jerinphilip, line 163: > #357 fixes this by making quality estimates code store references to the text in the same way as alignment etc. does: metadata always attached at the subword level.

@jelmervdl I assume you're talking about this. I agree instances of the above are an improvement.

jerinphilip (commented):
As was the general agreement in the plenary, I believe the current proposal (based on HTML) is different but solves the problem better reducing work at the extension.

Leaving the below review based on what I understood from the plenary. These are towards not creating an unnecessary failure mode (broken HTML as byte-range).

src/translator/response.cpp
jelmervdl: It is being used by `common-impl.cpp` in `TestSuite<Service>::qualityEstimatorWords`. Feel free move it there.

src/translator/response.h
jelmervdl: I'm in favour.

wasm/bindings/response_bindings.cpp
jelmervdl: If we ever need these in WASM, we should expose them through the same interface as the other APIs use. For now I've removed these functions.

abhi-agg:
> As was the general agreement in the plenary, I believe the current proposal (based on HTML) is different but solves the problem better reducing work at the extension.

Correction. There was no final agreement from Mozilla side in the Plenary regarding which of the 2 approaches (https://github.com/browsermt/bergamot-translator/pull/357 or https://github.com/browsermt/bergamot-translator/pull/358) will the extension use. We did say that we are evaluating https://github.com/browsermt/bergamot-translator/pull/358 option as this looks promising.

However, @andrenatal confirmed few hours later via email that the extension will use this approach.

> All features expected to be happening at the extension by using `ByteRange` instances (words, sentences) on plain-text translations are now using HTML as the delivery mechanism in proof of concepts:
> 
>     * [Quality estimation information jelmervdl/firefox-translations#11](https://github.com/jelmervdl/firefox-translations/pull/11)
> 
>     * [Show original sentence on hover jelmervdl/firefox-translations#5](https://github.com/jelmervdl/firefox-translations/issues/5)
> 
> Providing an alternate means of access incurs a maintenance overhead and is impractical to be used at the extension due to part tags being present in these `ByteRange`s potentially.

I agree about the maintenance overhead but even though we have a proof of concept available from @jelmervdl, we could have deferred this removal once QE integration is complete in the extension. However, I am not adamant about it.

@jelmervdl This PR is in draft state. Is it up for review?

jelmervdl:
@abhi-agg I kept it in a draft state waiting for your and @andrenatal's approval to take this approach. It should be ready for review now.

wasm/test_page/js/index.js
```diff
@@ -38,22 +38,48 @@ const _prepareTranslateOptions = (paragraphs) => {
...
+  return div.innerHTML;
+};
+
 const translateCall = () => {
-  const text = document.querySelector("#input").value + "  ";
+  const text = document.querySelector("#input").value;
   if (!text.trim().length) return;
 
-  const paragraphs = text.split("\n");
+  const paragraphs = [textToHTML(text)]; // escape HTML
```
abhi-agg, line 51: @jelmervdl Motivation of splitting the content with a new line separator was to create multiple entries to translate for the batch translation api. This change now will create only one entry in the batch. Right? Basically, `textToHTML` should be called for each new line separated entry of `text` object.

wasm/test_page/js/index.js
```diff
@@ -38,22 +38,48 @@ const _prepareTranslateOptions = (paragraphs) => {
...
+
 const translateCall = () => {
-  const text = document.querySelector("#input").value + "  ";
+  const text = document.querySelector("#input").value;
   if (!text.trim().length) return;
 
-  const paragraphs = text.split("\n");
+  const paragraphs = [textToHTML(text)]; // escape HTML 
   const translateOptions = _prepareTranslateOptions(paragraphs);
-  $("#output").setAttribute("disabled", true);
```
abhi-agg, line 47: Why was this removed?

wasm/test_page/js/index.js
jelmervdl: That makes sense! I've restored it (and updated the code to insert each entry in its own div in the output so that the sentence ids are at least unique inside the div. The 'mousover' highlight code now makes use of this.)

wasm/test_page/js/index.js
jelmervdl: I changed #output to be a `div` instead of a `textarea`, and div's don't have a disabled state.

src/translator/response.h
abhi-agg: So, we can just remove it. Right?

wasm/test_page/js/index.js
abhi-agg: Thanks for restoring.

abhi-agg:
@jelmervdl I did a quick test and overall it looks good to me. I can approve it after you rebase to latest main and resolve the https://github.com/browsermt/bergamot-translator/pull/358#discussion_r814760859. Thanks for the work 👍🏾

jerinphilip:
> `x-bergamot-sentence-score` and `x-bergamot-word-score` strings are too big.

Is this a variable naming nit or are there performance implications of this somewhere?

jelmervdl:
> x-bergamot-sentence-score and x-bergamot-word-score strings are too big.

What are they too long for? I intentionally made them this verbose to make sure they'd never overlap with any attribute that exists (or will exist) in a website. I think with just `sentence-score` and `word-score` that might occur. (It's a common string on Github.) Adding the `x-bergamot` makes it pretty unique. The `x-` prefix comes from https://stackoverflow.com/a/17902387 (the SO post summarises it better than the docs do)

abhi-agg:
> > `x-bergamot-sentence-score` and `x-bergamot-word-score` strings are too big.
> 
> Is this a variable naming nit or are there performance implications of this somewhere?

> > x-bergamot-sentence-score and x-bergamot-word-score strings are too big.
> 
> What are they too long for? I intentionally made them this verbose to make sure they'd never overlap with any attribute that exists (or will exist) in a website. I think with just `sentence-score` and `word-score` that might occur. (It's a common string on Github.) Adding the `x-bergamot` makes it pretty unique. The `x-` prefix comes from https://stackoverflow.com/a/17902387 (the SO post summarises it better than the docs do)

An obvious implication is the increased memory footprint which is evident in the example used in the description of this PR. Plus, all of this data will be copied from wasm to JS side during runtime. However, please feel free to ignore it as it was just a suggestion after reading https://github.com/browsermt/bergamot-translator/pull/358#issuecomment-1044274167.

I can approve the PR as soon as the rest of the comments are resolved 👍🏾

src/translator/response.h
jelmervdl: It is being used by `common-impl.h` to extract the words for the QE test cases. @jerinphilip can I leave it up to you what you want to do with it? Either move it in there, or keep it in response.h. It's not publicly exposed.

jerinphilip: I can take care of this eventually as part of https://github.com/browsermt/bergamot-translator/issues/350. Marking as resolved.

jerinphilip (approved):

abhi-agg (approved):

PR #370, opened 2022-03-03
Use right range and threshold for showing "bad" words/sentences

abhi-agg:
Higher QE scores means better quality.
Changed the threshold to ~~`-0.5`~~ `ln(0.5) => -0.6931` as per discussions in QE meetings.

@mfomicheva @abarbosa94 @felipesantosk Please let me know if any of the above is wrong 👍🏾


jelmervdl:
If these are the correct thresholds we can also lose the "these thresholds are just examples" comments :)

abhi-agg:
I just want to confirm with @mfomicheva @abarbosa94 @felipesantosk once more that the threshold of `-0.5` is a good one as a starting point **for all the language pairs** irrespective of whether the quality scores are returned using translation models or supervised QE models under the hood.

PR #377, opened 2022-03-14
JS: Reuse Model registry from firefox-translation-models for test page

abhi-agg:
- https://github.com/mozilla/firefox-translations-models/blob/main/registry.json is reused
- Removed existing registry


kpu:
Thanks by the way.

Issue #318, opened 2022-01-27
Ability to provide quality estimation models through WASM Bindings


jerinphilip:
https://github.com/browsermt/bergamot-translator/blob/cf541c68f9b43bce8c68e2292007a1573cfaa38e/wasm/bindings/service_bindings.cpp#L59-L64

Currently, the extension is missing the ability to provide QE Models through WASM absent QE models in the bundling function above.

PR #378, opened 2022-03-14
JS: Using supervised QE models for available language pairs

abhi-agg:
Fixes: https://github.com/browsermt/bergamot-translator/issues/318

Modified the wasm test page and tested it there. It works.
Supervised QE binary models are being used for 3 language pairs.

### DIFF
--- a/wasm/bindings/service_bindings.cpp
+++ b/wasm/bindings/service_bindings.cpp
@@ -45,11 +45,14 @@ std::vector<std::shared_ptr<AlignedMemory>> prepareVocabsSmartMemories(std::vect
 }
 
 MemoryBundle prepareMemoryBundle(AlignedMemory* modelMemory, AlignedMemory* shortlistMemory,
-                                 std::vector<AlignedMemory*> uniqueVocabsMemories) {
+                                 std::vector<AlignedMemory*> uniqueVocabsMemories, AlignedMemory* qeMemory) {
   MemoryBundle memoryBundle;
   memoryBundle.model = std::move(*modelMemory);
   memoryBundle.shortlist = std::move(*shortlistMemory);
   memoryBundle.vocabs = std::move(prepareVocabsSmartMemories(uniqueVocabsMemories));
+  if (qeMemory != nullptr) {
+    memoryBundle.qualityEstimatorMemory = std::move(*qeMemory);
+  }
 
   return memoryBundle;
 }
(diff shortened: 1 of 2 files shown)


abhi-agg:
@jelmervdl @jerinphilip Could one of you review the PR? It is a small one. I have to merge it today to be able to make progress on integrating supervised QE models in the extension.

jerinphilip:
> It works.

There are no tests accompanying this change. At least put a screenshot. How does the colour coding change? Does the visual output make sense? 

> I have to merge it today to be able to make progress on integrating supervised QE models in the extension.

There's nothing really stopping you from bringing the bigger PR which adds the QE feature as a whole to UI. This will additionally allow experimenting with values etc (https://github.com/browsermt/bergamot-translator/pull/370 is probably not going to work, for instance. Do we need [calibration](https://scikit-learn.org/stable/modules/calibration.html)?). I fail to understand what stops you from dogfooding your own extension UI experimentation developing in a branch and bringing changes here more complete - look at https://github.com/browsermt/bergamot-translator/pull/312 for example. Alternatively, do it properly at the test page and show application examples (screenshots, not console logs) - which makes an import into the extension UI perhaps easier. The larger context will help review here.

jelmervdl (approved):
I tested it locally by loading models that have (enes) and don't have (bgen) quality models associated with them. Both work as in I can see translations with scores attached. Scores seem to be in the range [-1.0, 0]. So not sure whether the binding changes are working, but they're not breaking anything.

Maybe the people working on QE have some reference data, e.g. an example sentence combined with one of our models that they expect a certain score for. Or some easy examples we can use to test it. E.g. I use german <-> english often to test movement. Similarly, there must be example sentences that trigger a response from the QE models we can see.

But this doesn't block this pull request imo.

abhi-agg:
> So not sure whether the binding changes are working, but they're not breaking anything.

Actually it is easy to see that bindings work because whenever supervised models are available then you will see the QE model size in the console log (which indicates that the TranslationModel construction API uses QE model aligned memory).

Attaching screenshot for reference.

Screenshot 2022-03-15 at 15 52 39

> Maybe the people working on QE have some reference data, e.g. an example sentence combined with one of our models that they expect a certain score for. Or some easy examples we can use to test it. E.g. I use german <-> english often to test movement. Similarly, there must be example sentences that trigger a response from the QE models we can see.

This sounds like an option for verification.

I am merging this one.

Issue #344, opened 2022-02-09
Bulgarian missing from test page


kpu:
https://github.com/browsermt/bergamot-translator/blob/a0cb1e4b3d2e06027a7f979b0c66f1336a6688e9/wasm/test_page/js/index.js#L10-L19
does not have Bulgarian.  

In the longer term, consider automating from https://translatelocally.com/models.json somehow, since the same will happen when Polish and French models come out.  Or we could deduplicate this with the extension repo?

PR #379, opened 2022-03-15
JS: Update languages & use Intl API for their display names

jelmervdl:
I noticed that Bulgarian wasn't on the list.

This change uses the registry.json file that's already being loaded by worker.js to populate the language dropdowns. I'm using [Intl.DisplayNames](http://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DisplayNames/DisplayNames) to show the full language names.


jerinphilip (approved):
image

Tested working. Fixes #344. Review from WebAssembly/JS CodeOwners were requested, but appears to be taking long. Merging, please carry forward anything amiss to issues.

PR #388, opened 2022-03-23
JS: Fix swap button on test-page

jerinphilip:
https://github.com/browsermt/bergamot-translator/pull/358 changed a text area to a div to show quality UI renders, causing the code that swapped content on click of the button to populate instead with `undefined`.

This change uses `innerText` on the output to populate input to fix the issue.


jerinphilip:
image

Swapped with button: 
image

jelmervdl (approved):
Thanks! I thought I fixed that, but that was in one of my own branches. Oops! 😅

PR #399, opened 2022-04-05
Rework WASM compilation options

jelmervdl:
Necessary to work with newer versions of emscripten that are more picky about which option goes to the compiler, and which to the linker. Also took the opportunity to remove the need for the patching of the bergamot-translation-worker.js file, this can now easily be done through supported apis. Furthermore, I tried to downsize the generated javascript and wasm code a bit.

Initial estimates show that bergamot-translator compiled with emscripten 3.0.0 runs at about 3x the speed of 2.0.9 (when using embedded intgemm). Speed-up when using mozIntGemm is less dramatic, I expect:

```
Benchmark 1: node ./test-nodejs.js ../build-wasm/
  Time (mean ± σ):      9.355 s ±  0.222 s    [User: 13.142 s, System: 0.368 s]
  Range (min … max):    9.119 s …  9.882 s    10 runs

Benchmark 2: node ./test-nodejs.js ../build-wasm-github/
  Time (mean ± σ):     27.380 s ±  1.004 s    [User: 30.565 s, System: 0.431 s]
  Range (min … max):   26.419 s … 29.500 s    10 runs

Summary
  'node ./test-nodejs.js ../build-wasm/' ran
    2.93 ± 0.13 times faster than 'node ./test-nodejs.js ../build-wasm-github/'
```

Filesizes are also a bit smaller:
```
.rw-r--r--   83Ki jelmer staff  5 Apr 12:27   -I build-wasm/bergamot-translator-worker.js
.rwxr-xr-x  5.8Mi jelmer staff  5 Apr 12:27   -I build-wasm/bergamot-translator-worker.wasm

.rw-r--r--@ 161Ki jelmer staff 31 Mar 11:21   -I build-wasm-github/bergamot-translator-worker.js
.rw-r--r--@ 6.6Mi jelmer staff 31 Mar 11:21   -I build-wasm-github/bergamot-translator-worker.wasm
```


abhi-agg:
> Initial estimates show that bergamot-translator compiled with emscripten 3.0.0 runs at about 3x the speed of 2.0.9 (when using embedded intgemm)

@jelmervdl Wow. This sounds awesome. Thanks for the initiative. I hope we get the same performance in browser environment as well. Irrespective of that, we should upgrade the toolchain.

Although the PR is in draft state, I am leaving few comments. I hope that is fine with you 👍🏾

abhi-agg:
> Initial estimates show that bergamot-translator compiled with emscripten 3.0.0 runs at about 3x the speed of 2.0.9 (when using embedded intgemm)

@jelmervdl Wow. This sounds awesome. Thanks for the initiative. I hope we get the same performance in browser environment as well. Irrespective of that, we should upgrade the toolchain.

Although the PR is in draft state, I am leaving few comments. I hope that is fine with you 👍🏾

CMakeLists.txt
```diff
@@ -112,9 +112,59 @@ message(STATUS "Project name: ${PROJECT_NAME}")
...
+    -mssse3
+    -msimd128
+    # -fno-exceptions # Can't do that because spdlog uses exceptions
+    -sDISABLE_EXCEPTION_CATCHING=1
+    -sSTRICT=1
+  )
+  list(APPEND WASM_LINK_FLAGS
+    --bind # TODO: I thought it was an alias for -lembind, but maybe not yet in this version of emscripten
+    -g2
+
```
abhi-agg, line 130: The optimization flag is a compile + link flag. So, we should add `-O3` in link flag as well.

abhi-agg, line 130: The optimization flag is a compile + link flag. So, we should add `-O3` in link flag as well.

jelmervdl: [You're right!](https://emscripten.org/docs/tools_reference/emcc.html#arguments). I'll fix it.

jelmervdl:
I only later noticed that marian-dev also has a wasm example itself. I think I removed too much from its CMakeLists file to still work, so I'll bring back some of that as well.

wasm/test_page/js/worker.js
```diff
@@ -23,13 +23,68 @@ const fileInfo = [
...
+    info['wasm_gemm'] = this.createFallbackGemm(Module, info);
+    console.info(`Using gemm from ${FALLBACK_GEMM} because`, e);
+  }
+
+  const binary = new Uint8Array(wasmBinary);
+  const {instance} = await WebAssembly.instantiate(binary, info);
+  
+  return instance;
+}
+
```
abhi-agg, line 75: One general comment. In my opinion, this code should still be part of the generated JS file because:

- It is common code that will be replicated for every client
- The client now needs to be aware of the optimized gemm and bears the responsibility of writing this code
- Without this code, the generated artifacts will not work out of box because even if optimized gemm is not available

If I am missing something here then please let me know. Else, I recommend to revert this specific change.

abhi-agg, line 75: One general comment. In my opinion, this code should still be part of the generated JS file because:

- It is common code that will be replicated for every client
- The client now needs to be aware of the optimized gemm and bears the responsibility of writing this code
- Without this code, the generated artifacts will not work out of box because even if optimized gemm is not available

If I am missing something here then please let me know. Else, I recommend to revert this specific change.

abhi-agg:
> I only later noticed that marian-dev also has a wasm example itself. I think I removed too much from its CMakeLists file to still work, so I'll bring back some of that as well.

I was about to write the same 👍🏾  If you want, I can also submit a PR to have similar changes in marian-dev if that is helpful.

jerinphilip:
> I only later noticed that marian-dev also has a wasm example itself. I think I removed too much from its CMakeLists file to still work, so I'll bring back some of that as well.

I propose going forward abandoning of WASM marian-decoder to keep things simple. The original attempt to port marian to WASM for inference left us with marian-decoder in WASM (https://github.com/browsermt/marian-dev/pull/6). Bergamot's WASM bindings/port have filled in the role.

jelmervdl:
> I propose going forward abandoning of WASM marian-decoder to keep things simple. The original attempt to port marian to WASM for inference left us with marian-decoder in WASM (https://github.com/browsermt/marian-dev/pull/6). Bergamot's WASM bindings/port have filled in the role.

I'm a fan. Simplifies things a lot.

> I was about to write the same 👍🏾 If you want, I can also submit a PR to have similar changes in marian-dev if that is helpful.

If we're abandoning the wasm interface in marian-dev (the one that's not actively being used or tested right now…) then just a minimal change like [this one](https://github.com/browsermt/marian-dev/compare/844800efccba6e670250caac1735ca2c8c8e508e...1675e80a4b2ca78c2a99d3f356fbaf4b1b424b43) should be sufficient. Maybe another one that removes the 'wasm' directory from marian-dev entirely.

wasm/test_page/js/worker.js
jelmervdl: I agree that we should have something that works out of the box, but I don't agree with you that the optimised gemm code needs to be included in that default code as well. It is only specifically javascript in extensions in Firefox Nightly that support optimised intgemm, and the number of clients that target that specific configuration will be pretty limited.

Ideally we have the wasm binary itself initialise the function pointers to the fallback implementations, and then have the ability to override them during initialisation. I haven't yet figured out how to do that.

abhi-agg: I am fine with maintaining FFx specific artifact at our end. However, this change can/should be done as a separate PR? If you are upgrading toolchain in this PR then let's stick to that.

abhi-agg:
> If we're abandoning the wasm interface in marian-dev (the one that's not actively being used or tested right now…) then just a minimal change like [this one](https://github.com/browsermt/marian-dev/compare/844800efccba6e670250caac1735ca2c8c8e508e...1675e80a4b2ca78c2a99d3f356fbaf4b1b424b43) should be sufficient. Maybe another one that removes the 'wasm' directory from marian-dev entirely.

I agree. So, a PR in marian-dev with the corresponding submodule update here should be sufficient. I haven't looked into the exact changes [in marian](https://github.com/browsermt/marian-dev/compare/844800efccba6e670250caac1735ca2c8c8e508e...1675e80a4b2ca78c2a99d3f356fbaf4b1b424b43) but I can go through them right away 👍🏾

jelmervdl:
@abhi-agg do you have a specific timeline for this issue? I.e. asap for code freeze or something? I was working on something else today, but can switch if necessary.

abhi-agg:
> @abhi-agg do you have a specific timeline for this issue? I.e. asap for code freeze or something? I was working on something else today, but can switch if necessary.

Yep. We want to freeze by eod today. So upgrading emscripten version if the topmost priority for me. If you don't have bandwidth today then I am more than happy to take it up.

jelmervdl:
@abhi-agg  Feel free to go ahead and take over this pull request. I feel that's probably the faster way of getting this done today.

abhi-agg:
> remove the need for the patching of the bergamot-translation-worker.js file, this can now easily be done through supported apis

Just a heads up. I will not be including this change. I also don't know right now how to have the wasm binary itself initialise the function pointers to the fallback implementations, and then have the ability to override them during initialisation.

jelmervdl:
Replaced by #414

PR #405, opened 2022-04-12
Update ssplit-cpp submodule

jerinphilip:
Checking if https://github.com/ugermann/ssplit-cpp/pull/29 is suitable for Windows, etc.


jerinphilip:
> As in a number of other recent issues, this is probably related to the fact that building some ports relies on tools from the rolling-release MSYS2 distribution, and past revisions of vcpkg do refer to binary packages which are no longer provided by MSYS2. You must update your vcpkg repo. You can use old port versions via overlay ports or manifest mode.

https://github.com/microsoft/vcpkg/issues/24028#issuecomment-1092482523

cc @abhi-agg. To be honest, I know very little `vcpkg`. I am trying an attempt at an upgrade, if it does not work out within reasonable time, Windows CI might remain broken for until someone can fix this. But I can tell that it's not ssplit causing issues right now. Even if it is we will deal with it later.

jerinphilip:
Closing this, will experiment with API breaking run-vcpkg and run-cmake in a smaller setup and bring in later.

Issue #421, opened 2022-06-08
Remove wormhole references


jerinphilip:
Wormhole is not being used after https://github.com/mozilla/firefox-translations/pull/353 by any known clients. 

[Quick search](https://github.com/browsermt/bergamot-translator/search?q=WORMHOLE) points to references in:
 - GitHub Workflows
 - Helper scripts: [wasm/patch-artifacts-enable-wormhole.sh](https://github.com/browsermt/bergamot-translator/blob/8de368c16674b378ad30135e3b3e664ff5ecf7a8/wasm/patch-artifacts-enable-wormhole.sh)
 - [CMakeLists.txt](https://github.com/browsermt/bergamot-translator/blob/e34420647ddc528dac47ec57bab13771bd05abef/CMakeLists.txt#L116) 
 - [ReadMe](https://github.com/browsermt/bergamot-translator/blob/e34420647ddc528dac47ec57bab13771bd05abef/README.md#L44)
 
 Might as well remove the complexity here and further in marian-dev?

kpu:
Removed from kpu/intgemm and browsermt/marian-dev

Issue #330, opened 2022-02-05
Enhanced JavaScript module from this repository


jerinphilip:
The current JavaScript "[example](https://github.com/browsermt/bergamot-translator/blob/b1e5a48f1a3e3a6d8a6d553dafe32d881eb3aabe/wasm/test_page/js/worker.js)"  for the test page is perhaps not the best provision from here. There are a few common functions that can be abstracted here, potentially tested headless and seen by multiple eyes. Therefore, a better solution is a JavaScript library or module, ideally with an API consistent with the C++ from which it is derived. 

The following are ideas of a few downstream applications of a JavaScript library that can do translation and is batteries included: 

1. The extension could be a consumer of this tested library eliminating human errors that come out of manual copying. (We got `alignment:true` and `responseOptions.HTML` correct here, but transfer to [firefox-translations](https://github.com/mozilla/firefox-translations/) led to inconsistent config).
2. https://github.com/jerinphilip/lemonade/issues/56.  
3. A plugin that can be embedded on to sites to provide translation similar to a google offering: https://www.dummies.com/article/technology/notable-websites/google/how-to-install-google-translates-website-translator-plugin-145074

There are parts of this library already existing in sources across a few repositories. At the moment the extra work and duplication is convenient for faster development cycles. But at some point in the future, we should look at consolidating commonly used functionalities here co-located with the bindings and continuously tested. The JavaScript module can be part of the releases from this repository. 

This issue is meant to investigate and discuss ideas, weigh the pros and cons of the approach.

jerinphilip:
OpenCV is a mature C++ library which exposes bindings to JavaScript and has JavaScript tests (https://docs.opencv.org/3.4/d4/da1/tutorial_js_setup.html) which we could potentially draw references/examples from. They also appear to be using WebAssembly SIMD and threads.

They allow variations of:

1. Manually in your browser
2. Headless with Puppeteer
3. Using Node.js.

jerinphilip:
@jelmervdl has successfully managed to run the WASM module via node.js in https://gist.github.com/jelmervdl/a4c8b6b92ad88a885e1cbd51c6ad4902. This opens up possibilities for headless testing.

PR #425, opened 2022-06-21
Basic HTML property testing for WebAssembly

jerinphilip:
Imports
https://gist.github.com/jelmervdl/a4c8b6b92ad88a885e1cbd51c6ad4902 and
attach it to CI.  NodeJS-14 is failing on trying to use the WebAssembly
binary. So we use node-16 independently setup.  This paves way for more
complicated testing for WebAssembly bindings in the future. It should be possible
to remove minimal builds in favour of closer to WebAssembly tests following this 
change.

https://github.com/mozilla/firefox-translations/pull/353 phases out wormhole. 
This change removes the `.js` and `.wasm` with wormhole from the GitHub 
release done. (This does not edit Mozilla's CircleCI or files, nor attempt to 
remove WORMHOLE completely).

This also includes a fix for a `.js` and `.wasm` release broken in #424.
GitHub CI code verified to work as intended at at fork 
(https://github.com/jerinphilip/bergamot-translator).

Related to: #421, #330

### DIFF
--- a/wasm/node-test.js
+++ b/wasm/node-test.js
@@ -1,0 +1,107 @@
+const {Blob} = require('buffer');
+const fs = require('fs');
+const https = require('https');
+
+const wasmBinary = fs.readFileSync('./bergamot-translator-worker.wasm');
+global.Module = {
+  wasmBinary,
+  onRuntimeInitialized
+};
+
+// Execute bergamot-translation-worker.js in this scope
+const js = fs.readFileSync('./bergamot-translator-worker.js', {encoding: 'utf8'});
+eval.call(global, js);
+
+/**
+ * Helper to download file into ArrayBuffer.
+ */
+function download(url) {
+  return new Promise((accept, reject) => {
+    https.get(url, (res) => {
+      const chunks = [];
+      res.on('error', reject);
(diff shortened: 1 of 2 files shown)


wasm/node-test.js
```diff
@@ -0,0 +1,107 @@
...
+
+  // Construct std::vector<ResponseOptions>
+  const options = new Module.VectorResponseOptions();
+  options.push_back({qualityScores: false, alignment: false, html: false});
+
+  // Translate our batch (of 1)
+  const output = service.translate(model, input, options);
+
+  // Get output from std::vector<Response>
+  console.log(output.get(0).getTranslatedText());
```
jelmervdl, line 101: Should it check the output to be "Hallo Welt!", to detect whether it is producing rubbish or not?

jerinphilip: I'm more in favour of property based testing. Input valid non-empty HTML, did I get valid non-empty HTML translation out, that sort. I expect such stuff to grow if we start exposing a JS library as mentioned in #330. Adding ASSERT_EQUALS will give us pain if compiler etc changes, which often happened in BRT.

jelmervdl (dismissed):
LGTM

PR #432, opened 2022-06-29
MacOS Wheels

graemenail:
Adds extra python wheels for MacOS

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -413,7 +413,7 @@ jobs:
           working-directory: ./doc
           run: |
             python3 -m pip install -r requirements.txt
-            python3 -m pip install --no-index --find-links=${{github.workspace}}/wheels bergamot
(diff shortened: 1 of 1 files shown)


jerinphilip (approved):
@graemenail Thank you for taking the time out to test this on MacOS and the improvements!

I see some docs failure, may I know what exactly happened here - did something fail? I notice the updated commands are possibly version/OS proof, which looks like an improvement and a TIL. 

@jelmervdl May I trouble you for a quick test if the wheels are functional on your Mac machine as well, once they're available in releases?

graemenail:
Some portion of the docs requires the bergamot package to be installed. The previous method installed from the wheel directly from `cp37-*`. 

Prior to this PR that was fine, because only the manylinux wheel matched this glob. Now we have a wheel for MacOS matching it too. 

I tried to modify the install such that it would pickup the correct wheel automatically. The last commit here removed `--no-index` because we need to search the index for bergamot dependencies. Perhaps there's a better way..

PR #434, opened 2022-07-06
Expand the node-test.js example code with documentation

jelmervdl:
This looked like the easiest way to provide some documentation & examples for the WASM API.

I've upgraded to node 18 because that allows us to use `fetch()` which matches what you would use in the browser.

Every line of code is annotated with some documentation on why it is there and what you can do with it.

### DIFF
--- a/wasm/node-test.js
+++ b/wasm/node-test.js
@@ -1,92 +1,85 @@
 #!/usr/bin/env node
+
+/**
+ * A note upfront: the bergamot-translator API is pretty low level, and
+ * embedding it successfully requires some knowledge about the WebWorkers and
+ * WebAssembly APIs. This script tries to demonstrate the bergamot-translator
+ * API with as little of that boiler plate code as possible.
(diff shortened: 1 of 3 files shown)


wasm/node-test.js
```diff
@@ -1,124 +1,172 @@
...
-  // Get output from std::vector<Response>
-  // The following works as a simple black-box test of the API, based on
-  // properties of HTML.
-  const translation = output.get(0).getTranslatedText()
-
-  // Print raw translation for inspection.
-  console.log(translation)
-
-  const fragment = JSDOM.fragment(translation)
+  // Translate our batch (of 1). Output will be another vector of type
```
jerinphilip, line 151: of 1 or 2? There's an HTML and a non HTML sample now, isn't there?

jerinphilip (dismissed):

XapaJIaMnu:
What about this one @jelmervdl

jelmervdl:
We can merge this as well. It works as documentation for the lower level API (the one exposed directly by the WASM interop code).

PR #437, opened 2022-07-26
More portable WASM demo

jelmervdl:
Replacing the WASM demo page with code that can be copied into other projects so people can easily embed bergamot translator on their own website.

I'm tempted to also add some version of `InPageTranslation` to make this a embeddable version of the translation plugin, a bit like Google Translate for websites.

### Tasks
- [x] Recreate WASM demo page functionality
- [x] Document BergamotTranslator classes and their public API
- [ ] Auto-publish npm packages on release?


jelmervdl:
Feel free to merge this for now if the tests pass. I have some future API changes planned but that's what NPM package versions are for.

PR #448, opened 2023-05-26
Terminology

XapaJIaMnu:
Basic Python interface + installation instructions

### DIFF
--- a/3rd_party/CMakeLists.txt
+++ b/3rd_party/CMakeLists.txt
@@ -31,6 +31,6 @@ get_directory_property(CMAKE_CXX_FLAGS DIRECTORY marian-dev DEFINITION CMAKE_CXX
 set(CMAKE_C_FLAGS ${CMAKE_C_FLAGS} PARENT_SCOPE)    
 set(CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS} PARENT_SCOPE)    
 
-if(COMPILE_PYTHON)
-  add_subdirectory(pybind11)
-endif(COMPILE_PYTHON)
+#if(COMPILE_PYTHON)
+#  add_subdirectory(pybind11)
+#endif(COMPILE_PYTHON)
--- a/README.md
+++ b/README.md
@@ -85,3 +85,55 @@ A short example of how to use the APIs is provided in `app/bergamot.cpp` file.
 ### Using WASM version
 
 Please follow the `README` inside the `wasm` folder of this repository that demonstrates how to use the translator in JavaScript.
+
+### Using python API
+
+Compile and install:
+```
+export CMAKE_BUILD_PARALLEL_LEVEL=8 # Use 8 cores to compile
+pip install wheel
+pip install .
+
+# Desktop app
+% bergamot-translator --help
+bergamot-translator interfance
+
+options:
+  -h, --help            show this help message and exit
+  --config CONFIG, -c CONFIG
+                        Model YML configuration input.
+  --num-workers NUM_WORKERS, -n NUM_WORKERS
+                        Number of CPU workers.
+  --logging LOGGING, -l LOGGING
+                        Set verbosity level of logging: trace, debug, info, warn, err(or), critical, off. Default is off
+  --cache-size CACHE_SIZE
+                        Cache size. 0 for caching is disabled
+  --terminology-tsv TERMINOLOGY_TSV, -t TERMINOLOGY_TSV
+                        Path to a terminology file TSV
+  --force-terminology, -f
+                        Force terminology to appear on the target side.
+  --path-to-input PATH_TO_INPUT, -i PATH_TO_INPUT
+                        Path to input file. Uses stdin if empty
+```
+Using the python interface
+```python
+from bergamot.translator import Translator
+print(Translator.__doc__)
+Bergamot translator interfacing with the C++ code.
+    
+    Attributes:
+        num_workers        Number of parallel CPU workers.
+        cache:             Cache size. 0 to disable cache.
+        logging:           Log level: trace, debug, info, warn, err(or), critical, off. Default is off
+        terminology:       Path to a TSV terminology file
+        force_terminology  Force the terminology to appear on the target side. May affect translation quality negatively.
+        
+        _config            Translation model config
+        _model:            Translation model
+        _responseOpts      What to include in the response (alignment, html restoration, etc..)
+        _service           The translation service
+
+translator = Translator("/path/to/model.npz.best-bleu.npz.decoder.brg.yml", terminology="/path/to/terminology.tsv")
+translator.translate(["text"])
+output
+```
--- a/bindings/python/translator.py
+++ b/bindings/python/translator.py
@@ -59,9 +59,7 @@ class Translator:
         return ret
     #@TODO add async translate with futures
 
-
-
-if __name__ == '__main__':
+def main():
     parser = argparse.ArgumentParser(description="bergamot-translator interfance")
     parser.add_argument("--config", '-c', required=True, type=str, help='Model YML configuration input.')
     parser.add_argument("--num-workers", '-n', type=int, default=1, help='Number of CPU workers.')
@@ -82,3 +80,5 @@ if __name__ == '__main__':
         for line in stdin:
             print(translator.translate(line))
 
+if __name__ == '__main__':
+    main()
--- a/bindings/python/translator.py
+++ b/bindings/python/translator.py
@@ -5,54 +5,79 @@ from sys import stdin
 from typing import List
 
 class Translator:
+    """Bergamot translator interfacing with the C++ code.
+    
+    Attributes:
+        num_workers        Number of parallel CPU workers.
+        cache:             Cache size. 0 to disable cache.
+        logging:           Log level: trace, debug, info, warn, err(or), critical, off. Default is off
+        terminology:       Path to a TSV terminology file
+        force_terminology  Force the terminology to appear on the target side. May affect translation quality negatively.
+        
+        _config            Translation model config
+        _model:            Translation model
+        _responseOpts      What to include in the response (alignment, html restoration, etc..)
+        _service           The translation service
+    """
     num_workers: int
     cache: int
     logging: str
     terminology: str
     force_terminology: bool
 
-    config: bergamot.ServiceConfig
-    model: bergamot.TranslationModel
-    responseOpts: bergamot.ResponseOptions
-    service: bergamot.Service
+    _config: bergamot.ServiceConfig
+    _model: bergamot.TranslationModel
+    _responseOpts: bergamot.ResponseOptions
+    _service: bergamot.Service
 
     def __init__(self, model_conifg_path: str, num_workers: int=1, cache: int=0, \
                  logging="off", terminology: str="", force_terminology: bool=False):
+        """Initialises the translator class
+
+        :param model_conifg_path: Path to the configuration file for the translation model.
+        :param num_workers: Number of CPU workers.
+        :param cache: cache size. 0 means no cache.
+        :param logging: Log level: trace, debug, info, warn, err(or), critical, off.
+        :param terminology: Path to terminology file, TSV format
+        :param force_terminology: Force terminology to appear on the target side. May impact translation quality.
+        """
         self.num_workers = num_workers
         self.cache = cache
         self.logging = logging
         self.terminology = terminology
         self.force_terminology = force_terminology
 
-        self.config = bergamot.ServiceConfig(self.num_workers, self.cache, self.logging, self.terminology, self.force_terminology)
-        self.service = bergamot.Service(self.config)
-        self.responseOpts = bergamot.ResponseOptions() # Default false for all, if we want to enable HTML later, from here
-        self.translationModel = self.service.modelFromConfigPath(model_conifg_path)
+        self._config = bergamot.ServiceConfig(self.num_workers, self.cache, self.logging, self.terminology, self.force_terminology)
+        self._service = bergamot.Service(self._config)
+        self._responseOpts = bergamot.ResponseOptions() # Default false for all, if we want to enable HTML later, from here
+        self._model = self._service.modelFromConfigPath(model_conifg_path)
 
     def resetTerminology(self, terminology: str="", force_terminology: bool=False) -> None:
-        """Resets the terminology of the model"""
+        """Resets the terminology of the model
+        :param terminology: path to the terminology file.
+        :param force_terminology: force terminology
+        :return: None
+        """
         self.terminology = terminology
         self.force_terminology = force_terminology
-        self.config = bergamot.ServiceConfig(self.num_workers, self.cache, self.logging, self.terminology, self.force_terminology)
-        self.service = bergamot.Service(self.config)
+        self._config = bergamot.ServiceConfig(self.num_workers, self.cache, self.logging, self.terminology, self.force_terminology)
+        self._service = bergamot.Service(self._config)
 
     def resetNumWorkers(self, num_workers) -> None:
-        """Resets the number of workers"""
+        """Resets the number of workers
+        :param num_workers: number of parallel CPU threads.
+        :return: None
+        """
         self.num_workers = num_workers
-        self.config = bergamot.ServiceConfig(self.num_workers, self.cache, self.logging, self.terminology, self.force_terminology)
-        self.service = bergamot.Service(self.config)
-
-    def translate(self, sentences: str) -> str:
-        "Translates a strings"
-        responses = self.service.translate(self.translationModel, bergamot.VectorString([sentences]), self.responseOpts)
-        ret = ""
-        for response in responses:
-            ret = ret + response.target.text
-        return ret
+        self._config = bergamot.ServiceConfig(self.num_workers, self.cache, self.logging, self.terminology, self.force_terminology)
+        self._service = bergamot.Service(self._config)
     
     def translate(self, sentences: List[str]) -> str:
-        "Translates a list of strings"
-        responses = self.service.translate(self.translationModel, bergamot.VectorString(sentences), self.responseOpts)
+        """Translates a list of strings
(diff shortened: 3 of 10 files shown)


jelmervdl:
Wouldn't it be much simpler to make the terminology map a member of the ResponseOptions class, and map it to a `dict`? Then you can change it per request, and you need less plumbing and no file IO for passing on that info.

I don't know if ResponseOptions is an immutable struct at this point. I don't think we have an async API on the python side so there's no risk of Python changing the terminology during a translation. But if you do want to implement that, its something to take into account.

README.md
```diff
@@ -85,3 +85,55 @@ A short example of how to use the APIs is provided in `app/bergamot.cpp` file.
...
+
+Compile and install:
+```
+export CMAKE_BUILD_PARALLEL_LEVEL=8 # Use 8 cores to compile
+pip install wheel
+pip install .
+
+# Desktop app
+% bergamot-translator --help
+bergamot-translator interfance
```
jelmervdl, line 99: _interfance_?

README.md
```diff
@@ -85,3 +85,55 @@ A short example of how to use the APIs is provided in `app/bergamot.cpp` file.
...
+from bergamot.translator import Translator
+print(Translator.__doc__)
+Bergamot translator interfacing with the C++ code.
+    
+    Attributes:
+        num_workers        Number of parallel CPU workers.
+        cache:             Cache size. 0 to disable cache.
+        logging:           Log level: trace, debug, info, warn, err(or), critical, off. Default is off
+        terminology:       Path to a TSV terminology file
+        force_terminology  Force the terminology to appear on the target side. May affect translation quality negatively.
```
jelmervdl, line 129: I wouldn't expose these as public properties, since changing them has no effect. You need to trigger a reset of the service first.

src/translator/service.cpp
```diff
@@ -137,6 +211,43 @@ AsyncService::AsyncService(const AsyncService::Config &config)
...
+    if(!myFile.is_open()) throw std::runtime_error("Could not open file: " + config_.terminologyFile);
+    std::string line;
+    while(std::getline(myFile, line)) {
+        // Create a stringstream of the current line
+        std::stringstream ss(line);
+
+        std::string srcword;
+        std::string replacementword;
+        getline(ss, srcword, '\t');
+        getline(ss, replacementword, '\n');
```
jelmervdl, line 229: Make sure people don't use windows line endings, hehe. Incidentally, if we could do this terminology map loading in Python you'd get their line ending strippping code for free.

src/translator/service.cpp
```diff
@@ -137,6 +211,43 @@ AsyncService::AsyncService(const AsyncService::Config &config)
...
+        // @TODO it seems like removing the tags forces the model to copy which is
+        // I guess just as good and more reliable. In that case we just don't tell the model
+        // what the original source is and it just has no choice BUT to generate the target.
+        if (!config_.terminologyForce) {
+          replacementword = srcword + " <tag0> " + replacementword + " </tag0> ";
+        }
+        this->terminologyMap_.insert({srcword, replacementword});
```
jelmervdl, line 236: why the explicit `this->`?

src/translator/service.cpp
```diff
@@ -202,6 +313,11 @@ void AsyncService::pivot(std::shared_ptr<TranslationModel> first, std::shared_pt
 void AsyncService::translate(std::shared_ptr<TranslationModel> translationModel, std::string &&source,
                              CallbackType callback, const ResponseOptions &responseOptions) {
   // Producer thread, a call to this function adds new work items. If batches are available, notifies workers waiting.
+  // Tagging
+  for (auto&& teminologyPair : terminologyMap_) {
+    source = Replace(source, teminologyPair.first, teminologyPair.second);
+  }
```
jelmervdl, line 319: Wait so `terminologyForce` is exactly the same as doing it on the input yourself? I.e.

```py
def forceTerminology(text, terms):
  for term, replacement in sorted(terms.items(), reverse=True, key=lambda pair: len(pair[0])):
    text = text.replace(term, replacement)
  return text


translator = Translator(...)
translator.translate([forceTerminology(text, terminology)])
 ```

That also highlights two issues I have with your implementation:

1. I don't see any explicit sorting on the length of the to be replaced terms. In that case, if you have `house` and `houseboat` in there, and the first one occurs first in the map, `houseboat` is never matched because `house` will already have been replaced.
2. Because you're restarting every loop from the start, terminology that appears in the replacement part of one term can then be modified if it matches the search part of another term.

File: src/translator/service.cpp [hunk 6]
```diff
@@ -137,6 +211,43 @@ AsyncService::AsyncService(const AsyncService::Config &config)
...
+        // Create a stringstream of the current line
+        std::stringstream ss(line);
+
+        std::string srcword;
+        std::string replacementword;
+        getline(ss, srcword, '\t');
+        getline(ss, replacementword, '\n');
+        // @TODO it seems like removing the tags forces the model to copy which is
+        // I guess just as good and more reliable. In that case we just don't tell the model
+        // what the original source is and it just has no choice BUT to generate the target.
```
jelmervdl, line 232: Edit: ah you say it here explicitly. Copy is a copy with the assumption it won't try to translate because it doesn't know the translation. For Chinese <-> English I can imagine this working, but no way that English <-> French would accept something like that… right?

bindings/python/translator.py
```diff
@@ -0,0 +1,109 @@
...
+    def translate(self, sentences: List[str]) -> str:
+        """Translates a list of strings
+        :param sentences: A List of strings to be translated.
+        :return: Translation output.
+        """
+        responses = self._service.translate(self._model, bergamot.VectorString(sentences), self._responseOpts)
+        ret = ""
+        for response in responses:
+            ret = ret + response.target.text
+        return ret
```
jelmervdl, line 84: Why require a list of inputs, but always only return a single string as output? I'd say do either one or the other, and I don't see a reason why we can't return a list in the output, e.g. `return [response.target.text for response in responses]`. Also saves you making assumptions on how to concatenate the output. There might be a need for spaces there.

3rd_party/CMakeLists.txt
```diff
@@ -31,6 +31,6 @@ get_directory_property(CMAKE_CXX_FLAGS DIRECTORY marian-dev DEFINITION CMAKE_CXX
 set(CMAKE_C_FLAGS ${CMAKE_C_FLAGS} PARENT_SCOPE)    
 set(CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS} PARENT_SCOPE)    
 
-if(COMPILE_PYTHON)
-  add_subdirectory(pybind11)
-endif(COMPILE_PYTHON)
+#if(COMPILE_PYTHON)
+#  add_subdirectory(pybind11)
+#endif(COMPILE_PYTHON)
```
jelmervdl, line 36: Looks like you moved this to a `find_package()` call later on. I think you can safely just remove it here then, no need to just comment it out.

jelmervdl:
Still on my wish list, but not a priority for a first release:
1. Move terminology config to the ResponseOptions struct. I still want to do the same with the Cache as well some day. I briefly looked into how hard this would be to do, but the way the C++ demo app parses command line options doesn't help.
2. Do terminology replacement *after* HTML parsing. Right now it happens before, which could invalidate the HTML and will mess up things if HTML-like tags are used for terminology. I looked at moving it to after HTML parsing, but that needs a bit more effort. The HTML parsing tracks byte offsets of the text it outputs, and expects the text that is translated (and for which we get alignment info after translation) is exactly the same as the text it produced. If any terminology replacements are added after HTML, but before translation, this doesn't hold. There is a data structure, AnnotatedText, that might be extended to do this task of tracking updated mappings. Using something like this was always the plan for HTML, and having a more expendable more onion-like structure, but we never got around to refactoring it.
3. Add pre/post processing that makes sure the terminology formatting *cannot* occur in the input string by replacing it with a placeholder, or doing something like the HTML transfer to copy it to the output.

jelmervdl:
Todo:  instead of a `--terminology-form` cli option, it should go in the model yaml.

graemenail:
> Todo:  instead of a `--terminology-form` cli option, it should go in the model yaml.

Should we nest application specific augments e.g.

```yml
bergamot-translator-app:
  terminology: terms.tsv
  terminology-form: "__source__ {src} __target {trg} __done__"
```

to avoid any future name conflicts with upstream Marian

jelmervdl:
Don't know if Marian's yaml parser can handle that, but if we can do that, it might be a good future-proofing.

Ideally there would be no collisions because this would be *the marian implementation* but maybe that's slightly unrealistic.

graemenail:
I would guess the parser library supports it, but from a quick look it's not wired up on the Marian side. But ideally Marian wouldn't need to parse it, just ignore it since terminology (for now?) is just implemented in bergamot-translator.

But on the ignoring unknown args:
```
./marian -c tmp.yml 
[2023-07-10 17:59:32] Error: There are option(s) in a config file that are not expected: bergamot-translator-app
```
(and similarly directly on the cli)

XapaJIaMnu:
The python interface randomly deadlocks, and I have no idea why. I didn't have this problem before so I blame the ensemble change, but i haven't bisected, or anything. FML.

XapaJIaMnu:
This is where we are at:
```
0x00007f2a3dd192ed in syscall () from /usr/lib/libc.so.6
(gdb) up
#1  0x00007f2a1a8de76a in std::__atomic_futex_unsigned_base::_M_futex_wait_until (this=<optimized out>, __addr=0x55d7e9141070, __val=2147483648, __has_timeout=<optimized out>, __s=..., __ns=...)
    at /usr/src/debug/gcc/gcc/libstdc++-v3/src/c++11/futex.cc:122
Downloading source file /usr/src/debug/gcc/gcc/libstdc++-v3/src/c++11/futex.cc
122     /usr/src/debug/gcc/gcc/libstdc++-v3/src/c++11/futex.cc: Directory not empty.                                                                                                                                    
(gdb) up
#2  0x00007f2a384e4b71 in std::__atomic_futex_unsigned<2147483648u>::_M_load_and_test_until (this=0x55d7e9141070, __assumed=0, __operand=1, __equal=true, __mo=std::memory_order_acquire, __has_timeout=false, 
    __s=std::chrono::duration = { 0s }, __ns=std::chrono::duration = { 0ns }) at /usr/lib/gcc/x86_64-pc-linux-gnu/12.3.0/include/c++/bits/atomic_futex.h:109
109               bool __ret = _M_futex_wait_until((unsigned*)(void*)&_M_data,
(gdb) up
#3  0x00007f2a384cb2ba in std::__atomic_futex_unsigned<2147483648u>::_M_load_and_test (this=0x55d7e9141070, __assumed=0, __operand=1, __equal=true, __mo=std::memory_order_acquire)
    at /usr/lib/gcc/x86_64-pc-linux-gnu/12.3.0/include/c++/bits/atomic_futex.h:158
158           return _M_load_and_test_until(__assumed, __operand, __equal, __mo,
(gdb) up
#4  0x00007f2a384a1ef8 in std::__atomic_futex_unsigned<2147483648u>::_M_load_when_equal (__mo=std::memory_order_acquire, __val=1, this=0x55d7e9141070)
    at /usr/lib/gcc/x86_64-pc-linux-gnu/12.3.0/include/c++/bits/atomic_futex.h:212
212           _M_load_and_test(__i, __val, true, __mo);
(gdb) up
#5  std::__future_base::_State_baseV2::wait (this=0x55d7e9141060) at /usr/lib/gcc/x86_64-pc-linux-gnu/12.3.0/include/c++/future:351
351             _M_status._M_load_when_equal(_Status::__ready, memory_order_acquire);
(gdb) up
#6  0x00007f2a384b4204 in std::__basic_future<marian::bergamot::Response>::wait (this=0x55d7e8790800) at /usr/lib/gcc/x86_64-pc-linux-gnu/12.3.0/include/c++/future:714
714             _M_state->wait();
(gdb) up
#7  0x00007f2a384a31b6 in ServicePyAdapter::translate (this=0x55d7e877a1e0, model=std::shared_ptr<marian::bergamot::TranslationModel> (use count 5, weak count 0) = {...}, 
    inputs=std::vector of length 10, capacity 10 = {...}, options=...) at /home/dheart/uni_stuff/postdoc/bergamot-translator/bindings/python/bergamot.cpp:82
82            futures[i].wait();
```

The code in question is here: https://github.com/browsermt/bergamot-translator/blob/1e80e799fd91573f6dfa99524ff7f2c065294a58/bindings/python/bergamot.cpp#L116

What's happening? For some reason a future just takes forever while no computation is happening...

XapaJIaMnu:
Reverting: `git revert 4b0da8d434e5a688139255873afd177f647ef777` fixes the issue. @graemenail not sure why.

jelmervdl:
@nick what is the python code you're using, and which models?

XapaJIaMnu:
How to compile the python interface with CUDA support:
```
export COMPILE_CUDA=ON 
export CMAKE_BUILD_PARALLEL_LEVEL=16
pip install .
```

bindings/python/translator.py
```diff
@@ -0,0 +1,212 @@
...
+    _terminology_form: str
+
+    _config: bergamot.ServiceConfig
+    _model: bergamot.TranslationModel
+    _responseOpts: bergamot.ResponseOptions
+    _service: bergamot.Service
+
+    def __init__(
+        self,
+        model_conifg_path: str,
```
jelmervdl, line 42: `model_conifg_path`

src/translator/service.cpp
```diff
@@ -135,8 +158,43 @@ AsyncService::AsyncService(const AsyncService::Config &config)
...
+    std::string line;
+    std::unordered_map<std::string, std::string> tempmap;  // Read in the TSV to here
+    while (std::getline(myFile, line)) {
+      // Create a stringstream of the current line
+      std::stringstream ss(line);
+
+      std::string srcword;
+      std::string replacementword;
+      getline(ss, srcword, '\t');
+      getline(ss, replacementword, '\n');  // BEWARE of windows file ndings
```
jelmervdl, line 187: This is why I would have liked it if we could have kept the "read the tsv file" part out of the bergamot C++ codebase, and just implement it in Python/Qt/JavaScript which are much better at integrating with the local environment.

… But treat this complaint so I'll stop complaining about this from now on!

src/translator/service.cpp
```diff
@@ -150,6 +208,29 @@ AsyncService::AsyncService(const AsyncService::Config &config)
...
+  // Copy the map. Since we might be coming from python space anyways. Also take care of force here
+  for (auto const &[key, val] : terminology) {
+    if (!forceTerminology) {
+      terminologyMap_[key] = string_format(config_.format, key, val);
+    } else {
+      // @TODO it seems like removing the tags forces the model to copy which is
+      // I guess just as good and more reliable. In that case we just don't tell the model
+      // what the original source is and it just has no choice BUT to generate the target.
+      terminologyMap_[key] = val;
+    }
```
jelmervdl, line 222: Can we flip these if/else clauses? I do apologise for being pedantic.

src/translator/service.cpp
```diff
@@ -202,12 +283,14 @@ void AsyncService::pivot(std::shared_ptr<TranslationModel> first, std::shared_pt
 void AsyncService::translate(std::shared_ptr<TranslationModel> translationModel, std::string &&source,
                              CallbackType callback, const ResponseOptions &responseOptions) {
   // Producer thread, a call to this function adds new work items. If batches are available, notifies workers waiting.
+  // Tagging
+  if (!terminologyMap_.empty()) source = ReplaceTerminology(std::move(source), terminologyMap_);
```
jelmervdl, line 287: `std::string ReplaceTerminology(std::string const &str, TerminologyMap const &terminology);`. I think I made the `std::move` pointless when I changed the `ReplaceTerminilogy` implementation.

src/translator/service.h
```diff
@@ -108,15 +109,26 @@ class BlockingService {
...
     static void addOptions(App &app, Config &config) {
       app.add_option("--cpu-threads", config.numWorkers, "Workers to form translation backend");
+      app.add_option("--gpu-workers", config.gpuWorkers, "GPU workers for the translation backend.");
       app.add_option("--cache-size", config.cacheSize, "Number of entries to store in cache.");
+      app.add_option("--terminology-file", config.terminologyFile, "tsv, one term at a time terminology file.");
+      app.add_option(
+          "--force-terminology", config.terminologyForce,
+          "Force the terminology to appear on the target side. May degrade translation quality. Not recommended.");
+      app.add_option("--terminology-form", config.format,
+                     "Form for technology. Default is \"%s __target__ %s __done__ \". Change depending on the model.");
```
jelmervdl, line 131: Do we want to figure out how to store this in the model yaml instead before we merge?

src/translator/terminology.cpp
```diff
@@ -0,0 +1,50 @@
...
+namespace {
+
+struct Hit {
+  size_t pos;
+  TerminologyMap::const_iterator entry;
+};
+
+}  // namespace
+
+std::string ReplaceTerminology(std::string const &str, TerminologyMap const &terminology) {
```
jelmervdl, line 17: Once `TerminologyMap` becomes very large, this method is going to be stupidly slow. At that point we should change it into a [state machine type of search thingy](https://en.wikipedia.org/wiki/Commentz-Walter_algorithm).

src/translator/service.cpp
XapaJIaMnu: Well, functionally, we can bypass the redundant terminology implementation, by applying terminology constraints directly. Maybe we can add this to the readme. I would like to keep a C++ implementation at the very least as documentation.

src/translator/service.h
XapaJIaMnu: Pedantically, we probably should.

src/translator/terminology.cpp
XapaJIaMnu: If terminology map is very large, i feel something wrong is happening.
State machine would always be better but also what is the time this would take.

PR #450, opened 2023-07-19
Enables model ensembles

graemenail:
Adds the ability to use ensembles of models. This supports ensembles of binary- or npz-format models, as well as mixtures of both.

When all models in the ensembles are of binary format, the load from memory path is used. Otherwise, they are loaded via the file system. Enable log-level debug for output related to this.

### DIFF
--- a/src/translator/byte_array_util.h
+++ b/src/translator/byte_array_util.h
@@ -5,7 +5,7 @@ namespace marian {
 namespace bergamot {
 
 AlignedMemory loadFileToMemory(const std::string& path, size_t alignment);
-AlignedMemory getModelMemoryFromConfig(marian::Ptr<marian::Options> options);
+std::vector<std::shared_ptr<AlignedMemory>> getModelMemoryFromConfig(marian::Ptr<marian::Options> options);
 AlignedMemory getQualityEstimatorModel(const marian::Ptr<marian::Options>& options);
 AlignedMemory getQualityEstimatorModel(MemoryBundle& memoryBundle, const marian::Ptr<marian::Options>& options);
 AlignedMemory getShortlistMemoryFromConfig(marian::Ptr<marian::Options> options);
--- a/src/translator/byte_array_util.h
+++ b/src/translator/byte_array_util.h
@@ -5,7 +5,7 @@ namespace marian {
 namespace bergamot {
 
 AlignedMemory loadFileToMemory(const std::string& path, size_t alignment);
-std::vector<std::shared_ptr<AlignedMemory>> getModelMemoryFromConfig(marian::Ptr<marian::Options> options);
+std::vector<AlignedMemory> getModelMemoryFromConfig(marian::Ptr<marian::Options> options);
 AlignedMemory getQualityEstimatorModel(const marian::Ptr<marian::Options>& options);
 AlignedMemory getQualityEstimatorModel(MemoryBundle& memoryBundle, const marian::Ptr<marian::Options>& options);
 AlignedMemory getShortlistMemoryFromConfig(marian::Ptr<marian::Options> options);
--- a/src/translator/definitions.h
+++ b/src/translator/definitions.h
@@ -19,8 +19,8 @@ typedef AlignedVector<char> AlignedMemory;
 /// Memory bundle for all byte-arrays.
 /// Can be a set/subset of model, shortlist, vocabs and ssplitPrefixFile bytes.
 struct MemoryBundle {
-  std::vector<std::shared_ptr<AlignedMemory>> models{};      ///< Byte-array of model (aligned to 256)
-  AlignedMemory shortlist{};  ///< Byte-array of shortlist (aligned to 64)
+  std::vector<std::shared_ptr<AlignedMemory>> models{};  ///< Byte-array of model (aligned to 256)
+  AlignedMemory shortlist{};                             ///< Byte-array of shortlist (aligned to 64)
 
   /// Vector of vocabulary memories (aligned to 64).
   /// If two vocabularies are the same (based on the filenames), two entries (shared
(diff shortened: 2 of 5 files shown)


graemenail:
emscripten was failing because I hadn't updated the bindings for WASM. I've fixed that, but only to restore support for single models. I can look at supporting ensembles here too if needed.

src/translator/translation_model.cpp
```diff
@@ -61,24 +61,35 @@ void TranslationModel::loadBackend(size_t idx) {
...
-          nullptr) {  // If we have provided a byte array that contains the model memory, we can initialise the
-                      // model from there, as opposed to from reading in the config file
-    ABORT_IF((uintptr_t)memory_.model.begin() % 256 != 0,
-             "The provided memory is not aligned to 256 bytes and will crash when vector instructions are used on it.");
-    if (options_->get<bool>("check-bytearray", false)) {
-      ABORT_IF(!validateBinaryModel(memory_.model, memory_.model.size()),
-               "The binary file is invalid. Incomplete or corrupted download?");
-    }
-    const std::vector<const void *> container = {
-        memory_.model.begin()};  // Marian supports multiple models initialised in this manner hence std::vector.
```
XapaJIaMnu, line 76: Hmmm this comment didn't age well.

XapaJIaMnu (approved):
I think it's fine. Maybe could make the wasm mode take the first model and cry an error message if there's more than one?

src/translator/translation_model.cpp
graemenail: 😅

graemenail:
The bindings only take in a single alignedmemory, so currently it just forwards that. 

At the point where I'm passing multiple models via the bindings, then it's basically enabled for wasm too. I'll have a look

src/translator/definitions.h
```diff
@@ -19,8 +19,8 @@ typedef AlignedVector<char> AlignedMemory;
 /// Memory bundle for all byte-arrays.
 /// Can be a set/subset of model, shortlist, vocabs and ssplitPrefixFile bytes.
 struct MemoryBundle {
-  AlignedMemory model{};      ///< Byte-array of model (aligned to 256)
-  AlignedMemory shortlist{};  ///< Byte-array of shortlist (aligned to 64)
+  std::vector<std::shared_ptr<AlignedMemory>> models{};  ///< Byte-array of model (aligned to 256)
```
kpu, line 22: Was `std::vector<AlignedMemory>` not working?  It has move semantics.

graemenail: I went for the same lifetime semantics as is implemented for the vocabulary. 

I'll check if shared_ptr can be avoided

graemenail: Works without `shared_ptr`. The vocabulary use-case avoids a second allocation when using joint-vocabularies.

While a user could specify the same model twice in the ensemble, I don't think it's necessary provide a more memory efficient behaviour in that instance, because they really should be using `weights`  in their `config.yml` to achieve that anyway.

graemenail:
I've checked the code is working after the latest changes. I used the regression test apps to check functionality
```sh
blocking \
  --log-level trace \
  --bergamot-mode test-forward-backward \
  --model-config-paths \
    ensemble_enes.yml \
    ensemble_esen.yml
```

Examples of the new logging messages introduced here:

When loading any npz file:
```
[2023-07-30 16:58:12] Encountered an npz file model.esen.npz; will use file loading for 4 models
```

When loading from memory:
```
[2023-07-30 16:58:12] Loaded model 1 of 2 from memory
[2023-07-30 16:58:12] Loaded model 2 of 2 from memory
```

When loading from file:
```
[2023-07-30 16:58:13] Loaded 4 model(s) from file
```

graemenail:
> Just to confirm, you see the increased runtime and actual different translations/scores?

Yes, there are different outputs and durations.

```
Model-Single:   If you use multiple models, it should be obvious that the decoding time is increasing.
real    0m4.500s
user    0m2.982s
sys     0m1.366s

Model-Ensemble:  If you use several models, it should be obvious that the decoding time increases.
real    0m15.184s
user    0m9.733s
sys     0m5.246s
```

Model-Single is one of 4 teachers from the Model-Ensemble. Timing of then ensemble is roughly 4x the single. Enabling logging shows that the models are loaded as scorers.

src/translator/byte_array_util.h
```diff
@@ -5,7 +5,7 @@ namespace marian {
 namespace bergamot {
 
 AlignedMemory loadFileToMemory(const std::string& path, size_t alignment);
-std::vector<std::shared_ptr<AlignedMemory>> getModelMemoryFromConfig(marian::Ptr<marian::Options> options);
+std::vector<AlignedMemory> getModelMemoryFromConfig(marian::Ptr<marian::Options> options);
```
XapaJIaMnu, line 8: Why were those shared to begin with?

Is this the design envisioning models would persist in memory after being unloaded in case an async service call has not completed its promise?

XapaJIaMnu (commented):
Just a question about the shared ptr removal and then ready to merge.

src/translator/byte_array_util.h
graemenail: I discussed it here https://github.com/browsermt/bergamot-translator/pull/450#discussion_r1278586467 a bit.

Basically, I saw the structure for vocabulary and assumed someone knew something about the lifetime of these objects that I didn't. But looking closer, `shared_ptr` for vocabularies exists to avoid a second allocation of memory when you have joint vocabs (which we do more often than not). 

I'm not fully clear on that. The TranslationModel owns the memory for the models/vocab/shortlist, but exposes does expose raw pointers to the marian backend. But given the AsyncService creates the TranslationModel instance, I'm not sure in what instances it would be unloaded before the expected fulfilment of a promise.

PR #454, opened 2023-07-31
Fix CI

graemenail:
Fixes CI for cibuildwheel and docs
If we want to enable Python 3.11 then we'll have to update our cibuildwheel action beyond 2.6.1

MacOS based native builds should start working with https://github.com/browsermt/bergamot-translator-tests/pull/63 and a the corresponding submodule update.


XapaJIaMnu:
I merged it to main. I guess it doesn't matter cause the HEAD commit will be the same, so can merge.

PR #459, opened 2023-08-13
Remove wormhole references

### LINKED ISSUE #421: Remove wormhole references

XapaJIaMnu:
Should fix #421 
I am not sure if the import intgemm as a separate step is still a thing though.


jelmervdl:
There's the _try linking to mozilla's native intgemm_ bit, but I don't think there's any wormhole code in the module itself.

https://github.com/browsermt/bergamot-translator/blob/ca954670aa4327630a3aee427668728b12b02df7/wasm/module/worker/translator-worker.js#L248-L253

I don't think we ever had a wasm binary that didn't use SIMD instructions to support very old machines for which Firefox doesn't support those. (But I don't know if Firefox supports those machines anyway anymore)

kpu:
> There's the _try linking to mozilla's native intgemm_ bit, but I don't think there's any wormhole code in the module itself.
> 
> https://github.com/browsermt/bergamot-translator/blob/ca954670aa4327630a3aee427668728b12b02df7/wasm/module/worker/translator-worker.js#L248-L253

I think some form of that is intended for production.  At least, they sent me a rewrite of intgemm for multiple architectures.  

> I don't think we ever had a wasm binary that didn't use SIMD instructions to support very old machines for which Firefox doesn't support those. (But I don't know if Firefox supports those machines anyway anymore)

Agreed, we didn't.  And if we did, it would have been to support Safari, which we decided to wait on until they added SIMD...

PR #470, opened 2024-03-23
model name update in download url in example script

Kirandevraj:
The model name in the download url for de-en translation in example script seems to have changed from `ende.student.tiny11.tar.gz` to `ende.student.tiny11.v2.93821e13b3c511b5.tar.gz`. I have updated the download url in this pr.


XapaJIaMnu (approved):
Those are not consistent since there's a git release tag attached to them now. I'll merge this but it may cause issues in the future...

PR #472, opened 2024-04-19
fix downloading of models in the python binding

bjesus:
models come in files named like `csen.student.base.v1.cd5418ba6a412fc7.tar.gz`, but the directory they create when extracted are named like `csen.student.base`. we therefore need to remove not just the extension but everything following and including the 3rd period


XapaJIaMnu:
It's probably more robust to take the first 3 rather than remove bits from the back. Could you change your pr to do that? (and sorry this project has lost lots of maintainers)

XapaJIaMnu:
... I shouldn't reply to things when I'm tired..

PR #477, opened 2025-08-11
Fix workerUrl and make some IA-specific mods

cdrini:


cdrini:
Ooops, meant to create this on the fork! Apologies.

PR #480, opened 2026-04-09
bump pybind11 3rdparty to 3.0.3

kamiccolo:
Motivation - support for Python >= 3.11 bindings


XapaJIaMnu:
Sorry, it seems like we need to update the workflows to a modern version before we can run testing. Have you tested on your end?

kamiccolo:
> Sorry, it seems like we need to update the workflows to a modern version before we can run testing. Have you tested on your end?

Yup, I can see that. Got a bit scared for a moment. 

"On my end" I've just built the project including Python bindings and made a few calls through the bindings (ARMv8). Can doublecheck that tonight and get back to you.

Also, would you like me to get a stab at those action scripts? Some of the issues do seem trivial (like action/v* dependency updates).

XapaJIaMnu:
Hi!

If you have time, i can review a PR that fixes the workflows first, but I think it might be a lot of work...

